PR #35, opened 2018-10-16
fix panic from nil LeadingComments

noahdietz:
Checks that `loc.LeadingComments` is not `nil` before dereferencing it.

I ran into this after I pulled master of this repo and tried to generate a client for the [Kiosk API](https://github.com/googleapis/kiosk/blob/master/protos/kiosk.proto).

I'm not entirely sure what a LeadingComment is. Also not sure if something about this proto made it crash or if it was something else I did. Once I put the check in, the plugin progressed passed the afflicted `init` method.

### DIFF
--- a/internal/gengapic/gengapic.go
+++ b/internal/gengapic/gengapic.go
@@ -155,9 +155,9 @@ func (g *generator) init(files []*descriptor.FileDescriptorProto) {
 			// since the field tag of Method is 2.
 			p := loc.Path
 			switch {
-			case len(p) == 2 && p[0] == 6:
+			case len(p) == 2 && p[0] == 6 && loc.LeadingComments != nil:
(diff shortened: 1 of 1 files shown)


internal/gengapic/gengapic.go
```diff
@@ -155,9 +155,9 @@ func (g *generator) init(files []*descriptor.FileDescriptorProto) {
 			// since the field tag of Method is 2.
 			p := loc.Path
 			switch {
-			case len(p) == 2 && p[0] == 6:
+			case len(p) == 2 && p[0] == 6 && loc.LeadingComments != nil:
```
pongad: We'd have to do this at ever branch right? I think it's probably easier to 

```go
if loc.LeadingComments == nil {
  continue
}
```

before the `switch`.

noahdietz: Good call, I'll change it up.

PR #53, opened 2018-12-03
add gencli plugin

noahdietz:
Adds a new a protoc plugin that generates [cobra](https://github.com/spf13/cobra) CLI application source-code (`gcli`) for gRPC APIs.

Code relevant to `gcli` construction is isolated to the `internal/gencli`  pkg and `cmd/protoc-gen-go_gcli` application.

Some changes are made to the `internal/pbinfo` pkg to support storage of comments for relevant proto elements.

CLIs have been generated for [googleapis/kiosk](https://github.com/googleapis/kiosk), [googleapis/gapic-showcase](https://github.com/googleapis/gapic-showcase), and [googleapis/google/cloud/language/v1](https://github.com/googleapis/googleapis/tree/master/google/cloud/language/v1) APIs and maunally tested.


internal/gencli/completion_file.go
```diff
@@ -0,0 +1,63 @@
...
+package gencli
+
+import (
+	"strings"
+	"text/template"
+)
+
+const (
+	// CompletionTemplate is the template string for the bash completion generation command
+	CompletionTemplate = `// AUTO-GENERATED CODE. DO NOT EDIT.
```
pongad: generated code comment.

internal/gencli/completion_file_test.go
```diff
@@ -0,0 +1,47 @@
...
+
+import (
+	"path/filepath"
+	"testing"
+
+	"github.com/googleapis/gapic-generator-go/internal/txtdiff"
+)
+
+func TestCompletionFile(t *testing.T) {
+	defer func() {
```
pongad: `go test` should catch panics automatically right?

internal/gencli/flag.go
```diff
@@ -0,0 +1,145 @@
...
+
+// GenOtherVarName generates the Go variable to store repeated Message & Enum string values
+func (f *Flag) GenOtherVarName(in string) string {
+	return in + strings.Replace(f.InputFieldName(), ".", "", -1)
+}
+
+// GenOneOfVarName generates the variable name for a oneof entry type
+func (f *Flag) GenOneOfVarName(in string) string {
+	name := f.InputFieldName()
+	if !f.IsNested && strings.Count(f.Name, ".") > 1 {
```
pongad:
```
if p := strings.LastIndex(f.Name, "."); p >= 0 && !f.IsNested {
  name = f.Name[:p]
}
```

internal/gencli/flag.go
```diff
@@ -0,0 +1,145 @@
...
+		name = f.Name[:strings.LastIndex(f.Name, ".")]
```
pongad: I don't really understand why we need to strip out the last section?

internal/gencli/util.go
```diff
@@ -0,0 +1,134 @@
...
+	return false
+}
+
+func copyImports(from, to map[string]*pbinfo.ImportSpec) {
+	for _, val := range from {
+		putImport(to, val)
+	}
+}
+
+func putImport(imports map[string]*pbinfo.ImportSpec, pkg *pbinfo.ImportSpec) {
```
pongad: Does this serve to make sure that we import each path as one name? I don't think this is necessary. Importing the same package twice under different names is perfectly valid:

```
import (
  foo "some/url"
  bar "some/url"
)
```

internal/gencli/util.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+}
+
+func parseMessageName(field *descriptor.FieldDescriptorProto, msg *descriptor.DescriptorProto) (name string) {
+	t := field.GetTypeName()
+	last := strings.LastIndex(t, ".")
+	name = t[last+1:]
+
+	// check if it is a nested type
+	if strings.Contains(t, msg.GetName()) {
```
pongad: How does this check for nested type?

internal/gencli/util.go
```diff
@@ -0,0 +1,134 @@
...
+func title(name string) string {
+	split := strings.Split(name, "_")
+	for ndx, tkn := range split {
+		split[ndx] = strings.Title(tkn)
+	}
+
+	return strings.Join(split, "")
+}
+
+func isNestedOneOf(in, cur string) bool {
```
pongad: I'm confused by this function :P

internal/gencli/util.go
```diff
@@ -0,0 +1,134 @@
...
+		parent := pre[strings.LastIndex(pre, ".")+1:]
+		name = parent + "_" + name
+	}
+
+	return
+}
+
+func title(name string) string {
+	split := strings.Split(name, "_")
+	for ndx, tkn := range split {
```
pongad: `for i, s := range split`

internal/gencli/flag.go
```diff
@@ -0,0 +1,145 @@
...
+	Required      bool
+	Usage         string
+	MessageImport pbinfo.ImportSpec
+	OneOfs        map[string]*Flag
+	IsOneOfField  bool
+	IsNested      bool
+}
+
+// GenOtherVarName generates the Go variable to store repeated Message & Enum string values
+func (f *Flag) GenOtherVarName(in string) string {
```
pongad: Do these `Name` functions create variable names to be used internally by the implementation? If so, I don't want to be too picky about them. Perhaps because this PR is large, I'm having a hard time figuring out how they are used.

internal/pbinfo/comments.go
```diff
@@ -0,0 +1,130 @@
...
+	SourcePathExtensionTypeValue = 7
+	// SourcePathEnumTypeValue represents the value associated
+	// with the Enum type in a source code path
+	SourcePathEnumTypeValue = 5
+	// SourcePathEnumValueTypeValue represents the value associated
+	// with the Enum Value type in a source code path
+	SourcePathEnumValueTypeValue = 2
+)
+
+func addComments(comments map[string]string, f *descriptor.FileDescriptorProto, loc *descriptor.SourceCodeInfo_Location) {
```
pongad: If you don't mind, I think we should move this out of `pbinfo` for now. I think the question of how we should represent comments are still not well-understood.

I believe it might be better to use `map[proto.Message]string` like https://github.com/googleapis/gapic-generator-go/blob/eaeab394a8a4df04f7260ed36245c17de8041699/internal/gengapic/gengapic.go#L145, so that we can look up directly. Otherwise, we have to make sure keys for two types cannot conflict. I think this implementation will run into trouble if we have
```
message A {
  message X {}
}
message B {
  message X{}
}
```

The two X's, I think, will end up with the same key.

internal/pbinfo/comments.go
```diff
@@ -0,0 +1,130 @@
...
+	var key string
+	p := loc.Path
+
+	switch {
+	// Service comment
+	case len(p) == SourcePathTopItemLength &&
```
pongad: Perhaps I'm running counter to modern programming practice, but I think using plain numbers would be better here.

```
case len(p) == 2 && p[0] == 6:
  f.Service[p[1]]
```
obviously won't go out of bound, and I still know that 6 refers to service because you write `f.Service` below.

cmd/protoc-gen-go_gcli/README.md
```diff
@@ -0,0 +1,137 @@
+# protoc-gen-go_gcli
```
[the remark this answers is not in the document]
noahdietz: +1

go.mod
```diff
@@ -8,6 +8,11 @@ require (
 	github.com/golang-commonmark/puny v0.0.0-20180910110745-050be392d8b8 // indirect
 	github.com/golang/protobuf v1.2.0
 	github.com/google/go-cmp v0.2.0
+	github.com/opennota/wd v0.0.0-20180911144301-b446539ab1e7 // indirect
+	github.com/russross/blackfriday v2.0.0+incompatible // indirect
+	github.com/shurcooL/sanitized_anchor_name v0.0.0-20170918181015-86672fcb3f95 // indirect
+	golang.org/x/sync v0.0.0-20181108010431-42b317875d0f // indirect
+	golang.org/x/tools v0.0.0-20181201035826-d0ca3933b724
```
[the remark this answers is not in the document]
noahdietz: `gencli` runs `tools/import` on all Go files added to 1. format the code 2. tidy the imports & remove extraneous ones (to make the generation code simpler)

The imports for diff lines 12-14 I believe were added because of other code in the project that uses them. I ran `go mod tidy` to add my deps (`tools`) and the others were added too.

I can remove the others if they aren't necessary

internal/gencli/cmd_file.go
```diff
@@ -0,0 +1,403 @@
...
+
+package gencli
+
+import (
+	"text/template"
+)
+
+const (
+	// CmdTemplate is the template for a cobra subcommand
+	CmdTemplate = `{{$inputVar := (print .Method "Input")}}
```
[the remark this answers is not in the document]
noahdietz: +1

internal/gencli/cmd_file.go
```diff
@@ -0,0 +1,403 @@
...
+{{$methodCmdVar := (print .Method "Cmd")}}
+{{$pollingCmdVar := (print .Method "PollCmd")}}
+{{$pollingOperationVar := (print .Method "PollOperation")}}
+{{$fromFileVar := (print .Method "FromFile")}}
+{{$outFileVar := (print .Method "OutFile")}}
+{{$serviceCmdVar := (print .Service "ServiceCmd")}}
+{{$followVar := (print .Method "Follow")}}
+{{ $serviceClient := ( print .Service "Client" ) }}
+// AUTO-GENERATED CODE. DO NOT EDIT.
```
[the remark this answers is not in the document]
noahdietz: +1

internal/gencli/flag.go
noahdietz: Yes, they generate the names for variables used to store the `pflag` values in some cases and the final request payload message object always.

Example: `Kiosk.name` field might be the `--name` flag and have a containing variable `var KioskInputName string`.

Because these methods are used in the templates as helpers, I believe they have to be exported.

internal/pbinfo/comments.go
noahdietz: I'm cool with using plain numbers here because you provide a good explanation, but yes this implementation was to avoid magic numbers.

I would still like to leave the comment on each case, if there is no objection there.

internal/gencli/util.go
noahdietz: A `oneof` found in a nested Message (a "nested oneof" as I refer to it) ends up having a name like NestedMessage.field vs. a top-level `oneof` is just a regular field and is TopLevel.field

This is only used in 1 place and this logic could just be run there, in context. I will make this change, if you have an objection lmk.

Making the function was one of those things where I try to predict what I'll need later and end up not needing it anywhere else :man_shrugging:

internal/gencli/completion_file_test.go
noahdietz: Yes, but the resulting message is not useful in debugging the test. The panic I aim to catch here is from `tools/import` panicking do to a formatting error, typically a result of a broken generation.

internal/gencli/flag.go
noahdietz: I cannot do this because the fields of nested messages are named using dot-notation i.e. `--size.width` and `oneof` fields I prefix with the `oneof`'s name, followed by the field name. Thus, a `oneof` field that is a message might have multiple `.` but not be nested (a oneof found inside of a nested message rather than top level). That is what this case handles.

Because all `oneof` field flags are prefixed with the `oneo` name, there will always be a `.`.

internal/gencli/util.go
noahdietz: nested types will have a type name like `package.Parent.Child` but a name like `Child`. So an indication from just a MessageDescriptor and FieldDescriptor (without looking through all of the NestedTypes of another message to see if it its in there) is if the field's type name contains it's parent message name.

```
message Kiosk {
  message ScreenSize {
    ...
  }

  ScreenSize size = 1;
}

f = size field
msg = some kiosk

f.GetTypeName() --> Kiosk.ScreenSize
msg.GetName() --> Kiosk
```

The resulting naming as generated by gRPC is different, and needs to be handled.

internal/gencli/util.go
noahdietz: Good point. I will change the import map key to the `pkg.Name` field instead of the `pkg.Path`.

go.mod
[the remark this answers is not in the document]
pongad: Hmm. I see what you mean. How hard would it be to only add imports we actually need?

This came up when I worked on `gapic-generator-go` earlier. The problem was that eventually you create a bug where the generator prints syntactically invalid code. Then `tools/import` can't work on the code and things become frustrating to debug.

internal/gencli/flag.go
pongad: Oh right, because you're looking for more than one dot. OK.

internal/gencli/gencli.go
```diff
@@ -441,6 +452,49 @@ func (g *gcli) addImport(cmd *Command, t pbinfo.ProtoType) (*pbinfo.ImportSpec,
...
+	// Enum Value comment
+	case len(p) == 4 && p[0] == 5 && p[2] == 2:
+		key = f.EnumType[p[1]].Value[p[3]]
+
+	// TODO(ndietz): this is an incomplete mapping of comments
+	default:
+		return
+	}
+
+	if _, ok := g.comments[key]; !ok {
```
pongad: Is this check necessary? I think we're in some weird place if we can reach one element two different ways.

internal/gencli/completion_file.go
```diff
@@ -0,0 +1,62 @@
...
+		Long: ` + "`Enable bash completion like so:\n" +
+		"Linux:\n" +
+		"  source <({{ .MethodCmd }} completion)\n" +
+		"Mac:\n" +
+		"  brew install bash-completion\n" +
+		"  {{ .MethodCmd }} completion > $(brew --prefix)/etc/bash_completion.d/{{ .MethodCmd }}`,\n" +
+		`Run: func(cmd *cobra.Command, args []string) {
+			rootCmd.GenBashCompletion(os.Stdout)
+		},
+	}`
```
pongad: I think it's better if we can keep everything in one string literal. If you need backticks, one trick is to `{{.Backtick}}` and put it in the template argument :rofl:

internal/gencli/testdata/watch-todo.want
```diff
@@ -0,0 +1,82 @@
...
+			cmd.MarkFlagRequired("id")
+
+		}
+
+	},
+	RunE: func(cmd *cobra.Command, args []string) (err error) {
+
+		in := os.Stdin
+		if WatchTodoFromFile != "" {
+			in, err = os.Open(WatchTodoFromFile)
```
pongad: We need to close this file. Alternatively, use `iotuil.ReadFile`.

internal/gencli/testdata/manage-todos.want
```diff
@@ -0,0 +1,104 @@
...
+				}
+				fmt.Fprintln(out, str)
+			}
+		}()
+
+		if Verbose {
+			fmt.Println("Client stream open. Close with blank line.")
+		}
+
+		scanner := bufio.NewScanner(in)
```
pongad: I'd just do `bufio.NewScanner(os.Stdin)`. Declaring `in` seems to suggest you might have logic to change the input.

internal/gencli/testdata/manage-todos.want
```diff
@@ -0,0 +1,104 @@
...
+	"fmt"
+	"os"
+
+	"github.com/golang/protobuf/jsonpb"
+	"github.com/spf13/cobra"
+
+	todopb "github.com/googleapis/todo/generated"
+)
+
+var ManageTodosInput todopb.Todo
```
pongad: Let's put this in the function, I don't think it's needed in the package scope.

internal/gencli/testdata/manage-todos.want
```diff
@@ -0,0 +1,104 @@
...
+
+		if Verbose {
+			fmt.Println("Client stream open. Close with blank line.")
+		}
+
+		scanner := bufio.NewScanner(in)
+		for scanner.Scan() {
+			input := scanner.Text()
+			if input == "" {
+				break
```
pongad: Shouldn't this be `continue`? Presumably you sometimes have blank lines?

internal/gencli/testdata/root_file.want
```diff
@@ -0,0 +1,59 @@
...
+	"fmt"
+	"os"
+
+	"github.com/golang/protobuf/proto"
+	"github.com/spf13/cobra"
+	"golang.org/x/net/context"
+)
+
+var Verbose, OutputJSON bool
+var ctx context.Context
```
pongad: I think we should inline `context.Background()`. Passing `ctx` implies some context propagation.

internal/gencli/testdata/watch-todo.want
noahdietz: oops I totally forgot to close it.

internal/gencli/testdata/manage-todos.want
noahdietz: > you might have logic to change the input

A user is always able to supply input from a file, using `--from_file`, containing JSON representation of proto messages.

```
{"content": "this"}
{"content": "is"}
{"content": "a"}
{"content": "test"}
```

internal/gencli/testdata/manage-todos.want
noahdietz: These need to be globally scoped so that both the `pflag` configuration in the following `init()` function and the command execution function given to `RunE` can use them. 

I could make them unexported if that helps?

internal/gencli/testdata/manage-todos.want
pongad: Oh whoops sorry I missed the if-stmt

internal/gencli/testdata/manage-todos.want
noahdietz: I made the decision to say that a blank line ends the client stream, requiring an entire message to be provided on within a single line break. So this was on purpose. Usage similar to the example I gave for file input on an earlier comment.

Wdyt? Should it look for a close sequence instead?

internal/gencli/testdata/root_file.want
noahdietz: I'm cool with inlining it.

I just initialized it globally, at the root-command level so that it didn't have to be generated in every command. 

Wdyt? Inline or leave it?

noahdietz:
I agree with your sentiment and approach. I'm happy to rework it, but I think it would be better in another PR. wdyt?

internal/gencli/testdata/manage-todos.want
pongad: `ManageTodosInput` only appears in one function though. Is this the case where it's needed in `init()` some of the times?

internal/gencli/testdata/manage-todos.want
pongad: I opine that blank line ending the stream sounds a little funny. It seems like you should be able to do something like

gcli <<EOF
foo1
foo2
foo3

foo4
foo5
foo6
EOF
```
and use the line break as a visual hint for some kind of logical separation. I admit I don't have a concrete use case in mind. If you think blank line is the right approach I won't fight it very hard.

internal/gencli/testdata/root_file.want
pongad: I think we should either inline it or actually init it globally like `var ctx = context.Background()`.

internal/gencli/testdata/manage-todos.want
noahdietz: Ahh in this case it is a client streaming method, so it is only used in the command execution. Typically the variable is used elsewhere (in `init()`), but in this case it is not.

I can account for this case and only declare the global variable in non-client streaming cases.

internal/gencli/testdata/manage-todos.want
noahdietz: I understand where you are coming from. I will brain storm and address in a separate PR!

go.mod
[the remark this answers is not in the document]
noahdietz: There aren't a lot of imports, so I can take another look at this.

Personally, I like formatting the generated code prior to emission. I'd prefer to not generate anything in the case where something comes out broken rather than wait for another step to catch it (i.e. attempting `go build` on the generated code).

I do sympathize with the difficulty in debugging such errors. No code gets generated to debug! 

Wdyt about an option to disable formatting so that a failed run could be debugged with a subsequent `--gcli_opt 'fmt:false` run?

I'm not married to formatting, so I don't mind throwing it out altogether, but would like to see it stay in some capacity if possible because I find it useful.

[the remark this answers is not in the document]
pongad: Another part of my thinking is that if an import is used sometimes, it sounds cleaner for the program to know when to import it than to generate too many imports and strip out later. One more thing to watch out for is that the result of `go/format` which the `import` library uses can change with version, so you probably want to do baseline test before formatting.

I understand your argument though; I can be convince either way. If we end up doing the wrong thing, it isn't that hard to switch. I think the flag you suggested sounds reasonable.

internal/gencli/util.go
noahdietz: Changing this to look up the parent of the field's message type with `pbinfo` and use that in naming if present

PR #54, opened 2018-12-05
gen versionGo in a util file

noahdietz:
Fixes a bug I introduced with #52 : `versionGo()` defined in each client file when generating multiple Services in the same package

Solution: generate a singular `util.go` file that contains the `versionGo()` method for all generated clients in a package to use 

I chose to put it in a `util.go` file (instead of `version.go`) in case there were other helpers in the future that would be shared across clients in a package


pongad:
Code LGTM but let's put this in `doc.go` instead. It's a little strange but that's where we put it in the monolithic generator right now. https://github.com/GoogleCloudPlatform/google-cloud-go/blob/master/dlp/apiv2/doc.go

PR #58, opened 2018-12-11
add docker invocation wrapper

noahdietz:
* add `Dockerfile` for invocation wrapper image per micro-generator docker invocation spec
* add Makefile w/image building target
* add docker wrapper usage instructions to the README
* add copy of `gapic.sh` from [gapic-generator-python](https://github.com/googleapis/gapic-generator-python/blob/master/gapic.sh) w/added required flag

Note: Using `debian:stable-slim` as the base image because I couldn't get `alpine` to work (would always get `protoc: not found` even though it was in the right dir & marked executable). Image is only 65.6MB in size.

This was tested on linux & macOS machines.

Once this is merged, I will manually publish images tagged `0.1.0`  (only current release) & `latest`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,4 @@
+image:
+	GOOS=linux GOARCH=amd64 go build github.com/googleapis/gapic-generator-go/cmd/protoc-gen-go_gapic
+	docker build -t gcr.io/gapic-images/gapic-generator-go .
+	rm protoc-gen-go_gapic
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
 image:
-	GOOS=linux GOARCH=amd64 go build github.com/googleapis/gapic-generator-go/cmd/protoc-gen-go_gapic
+	CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build github.com/googleapis/gapic-generator-go/cmd/protoc-gen-go_gapic
 	docker build -t gcr.io/gapic-images/gapic-generator-go .
 	rm protoc-gen-go_gapic
--- a/README.md
+++ b/README.md
@@ -52,9 +52,9 @@ $ docker pull gcr.io/gapic-images/gapic-generator-go
 $ docker run \
   --rm \
   --user $UID \
-  --mount type=bind,source=/abs/path/to/protos,destination=/in/protos,readonly \
+  --mount type=bind,source=</abs/path/to/protos>,destination=/in,readonly \
   --mount type=bind,source=$GOPATH/src,destination=/out/ \
-  --env "GO_GAPIC_OPT=github.com/package/import/path;name" \
+  --env "GO_GAPIC_OPT=<github.com/package/import/path;name>" \
   gcr.io/gapic-images/gapic-generator-go
 ```
 
--- a/README.md
+++ b/README.md
@@ -42,6 +42,38 @@ Idiomatically the name is last element of the path but it need not be.
 For instance, the last element of the path might be the package's version, and the package would benefit
 from a more descriptive name.
 
+Docker Wrapper
+--------------
+The generator can also be executed via a Docker container. The image containes `protoc`, the microgenerator
+binary, and the standard API protos.
+
+```bash
+$ docker pull gcr.io/gapic-images/gapic-generator-go
+$ docker run \
+  --rm \
+  --user $UID \
+  --mount type=bind,source=/abs/path/to/protos,destination=/in/protos,readonly \
+  --mount type=bind,source=$GOPATH/src,destination=/out/ \
+  --env "GO_GAPIC_OPT=github.com/package/import/path;name" \
+  gcr.io/gapic-images/gapic-generator-go
+```
+
+Replace `/abs/path/to/protos` with the absolute path to the input protos and `github.com/package/import/path;name`
+with the desired import path & name for the `gapic`, as described in [Invocation](#Invocation).
+
+For convenience, the [gapic.sh](./gapic.sh) script wraps the above `docker` invocation.
+An equivalent invocation using `gapic.sh` is:
+
+```bash
+$ gapic.sh \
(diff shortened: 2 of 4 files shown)


gapic.sh
```diff
@@ -0,0 +1,105 @@
...
+    -o | --out ) OUT="$2"; shift 2 ;;
+    -p | --path ) PROTO_PATH=$2; shift 2 ;;
+    -- ) shift; break; ;;
+    * ) break ;;
+  esac
+done
+
+# Ensure that all required options are set.
+if [ -z "$GO_GAPIC_OPT" ] || [ -z "$IMAGE" ] || [ -z "$IN" ] || [ -z "$OUT" ]; then
+  >&2 echo "Required argument missing."
```
pongad: Prefer `cat` over repeated `echo`s.

gapic.sh
```diff
@@ -0,0 +1,105 @@
...
+# Ensure Docker is running and seems healthy.
+# This is mostly a check to bubble useful errors quickly.
+if ! docker ps > /dev/null; then
+  exit $?
+fi
+
+# If the output directory does not exist, create it.
+if ! mkdir -p $OUT ; then
+  exit $?
+fi
```
pongad: Maybe `set -e` up top so we automatically fail out if any command fails?

README.md
```diff
@@ -42,6 +42,38 @@ Idiomatically the name is last element of the path but it need not be.
...
+binary, and the standard API protos.
+
+```bash
+$ docker pull gcr.io/gapic-images/gapic-generator-go
+$ docker run \
+  --rm \
+  --user $UID \
+  --mount type=bind,source=/abs/path/to/protos,destination=/in/protos,readonly \
+  --mount type=bind,source=$GOPATH/src,destination=/out/ \
+  --env "GO_GAPIC_OPT=github.com/package/import/path;name" \
```
lukesneeringer: I do not think it should hold up this PR, but we should actually make a convention for passing options to protoc. Since there is only one option switch available per plugin, it might be wise to use URI query string syntax (`key=value&key2=value2`) for extensibility. This should be something that has easy tooling to consume in every language. Thoughts?

Summoning @andreamlin who might also be interested in this.

gapic.sh
```diff
@@ -0,0 +1,105 @@
...
+IN=
+OUT=
+PROTO_PATH=`pwd`
+
+# Print help and exit.
+function show_help {
+  echo "Usage: $CMD --go_gapic_opt GO_GAPIC_OPT --image IMAGE --in IN_DIR --out OUT_DIR [--path PATH_DIR]"
+  echo ""
+  echo "Required arguments:"
+  echo "      --go_gapic_opt  The import path and name for the generated package"
```
lukesneeringer: It still need not block this PR, but let's have a meeting to discuss this. Go requires this option, but other languages likely need to permit it.

gapic.sh
```diff
@@ -0,0 +1,105 @@
...
+  exit $?
+fi
+
+# If the output directory is not empty, warn (but continue).
+if [ "$(ls -A $OUT )" ]; then
+  >&2 echo "Warning: Output directory is not empty."
+fi
+
+# If the image is not yet on the machine, pull it.
+if ! docker images $IMAGE > /dev/null; then
```
pongad: I don't think this test works, if I run `docker images derpderpderp; echo $?`, I got exit status 0. Maybe `docker images | grep -q $IMAGE`?

noahdietz: we don't actually need this test because `docker run` pulls the image if it's not present. Removing it.

CC @lukesneeringer

Makefile
```diff
@@ -0,0 +1,4 @@
+image:
+	GOOS=linux GOARCH=amd64 go build github.com/googleapis/gapic-generator-go/cmd/protoc-gen-go_gapic
+	docker build -t gcr.io/gapic-images/gapic-generator-go .
+	rm protoc-gen-go_gapic
```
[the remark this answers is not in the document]
noahdietz: Call me old school, but I find Makefiles more useful & accessible as more tasks are automated. Considering that this project hosts two plugins now, I anticipate adding more targets and I'd prefer to refrain from adding more script files :smile:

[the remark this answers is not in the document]
pongad: I'm good with it :smile:

README.md
noahdietz: `only one option switch available per plugin` - but it can be present several times in a single invocation of `protoc`. Not sure we need the `&` part. imo it would be more user-friendly to supply multiple instances of the option with their `key=value` pairs.

```bash
protoc ... \
  --go_gapic_out=/out/ \
  --go_gapic_opt=key1=value1 \
  --go_gapic_opt=key2=value2 \
  my.proto
```

But as we wrap the plugin invocation more, this becomes more of an implementation detail and less of a UX concern. Thus, I'm not too concerned on how we do this

lukesneeringer: Oh, I see. What happens if `--go_gapic_out` is specified multiple times? I just checked, and `string parameter` in `plugin.proto` [is not a repeated field](https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/compiler/plugin.proto#L74), so what does the plugin get if it is multiply specified?

noahdietz: They are concatenated with commas. The same example input is presented to the plugin impl as
`key1=value1,key2=value2,key3=value3`.

The `gcli` plugin requires multiple options (two to be exact) and I went with the key:value approach similar to what I've shown above.

noahdietz: I couldn't find any docs detailing this behavior but here is the [source](https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/compiler/command_line_interface.cc#L1694)!

lukesneeringer: Okay, in that case, let's comma separate, and then it does not matter which approach is used; they will both work. (It is unfortunate that we lose URL escaping tools, but oh well.)

PR #63, opened 2018-12-12
formatting option & generation execution improvements

noahdietz:
* remove tools/import dep
  * using go/format instead
* add fmt plugin option per [comment](https://github.com/googleapis/gapic-generator-go/pull/53#discussion_r238858506)
  * `--gcli_opt=fmt=<boolean>` default `true`
* refactor cmd gen
  * generates the code immediately following processing of method, rather than after processing it all
* explicitly add necessary subcmd imports per [comment](https://github.com/googleapis/gapic-generator-go/pull/53#discussion_r238858506)
  * previously depended on tools/import to resolve deps on generated code -> this drastically increased plugin execution time
* change plugin option key=value delimiter from `:` to `=` to align with [discussion](https://github.com/googleapis/gapic-generator-go/pull/58#discussion_r240440718)


internal/gencli/gencli.go
```diff
@@ -208,15 +175,46 @@ func (g *gcli) buildCommands() {
...
 
-			// build input fields into flags
-			if cmd.InputMessageType != EmptyProtoType && !cmd.ClientStreaming {
-				cmd.Flags = append(cmd.Flags, g.buildFieldFlags(&cmd, msg, "", false)...)
+			if cmd.InputMessageType != EmptyProtoType {
+				// imports handling input
+				putImport(cmd.Imports, &pbinfo.ImportSpec{
+					Name: "os",
+					Path: "os",
+				})
```
pongad: If the `Name` is the same as the name of the package, I think we can just drop the name. Then if we `fmt.Printf("%s %q", imp.Name, imp.Path)` we get back `import "os"` as expected.

internal/gencli/gencli.go
```diff
@@ -208,15 +175,46 @@ func (g *gcli) buildCommands() {
...
+
+				putImport(cmd.Imports, &pbinfo.ImportSpec{
```
pongad: same as above.

PR #65, opened 2018-12-13
add testing make targets

noahdietz:
* refactor gcli sanity testing script to download/generate dependencies
* add testprotos to gcli plugin
  * these are copies of existing protos with adjustments for gapic & gcli compatibility
* add make targets for testing gapic, gcli, and everything


cmd/protoc-gen-gcli/testprotos/kiosk/kiosk.proto
```diff
@@ -0,0 +1,141 @@
+// Copyright (c) 2018, Google LLC.
```
pongad: I assume this was copied from somewhere?

noahdietz: This proto was taken from github.com/googleapis/kiosk, yes

PR #66, opened 2018-12-13
make gapic shortcut expose generic options flag

noahdietz:
* changes `gapic.sh` docker shortcut to use expose the generic `--options` flag
* changes plugin options environment variable for the image to generic `PLUGIN_OPTIONS`
* update README Docker Wrapper section

### DIFF
--- a/gapic.sh
+++ b/gapic.sh
@@ -18,29 +18,29 @@ CMD="$0"
 
 # Set variables used by this script.
 # All of these are set in options below, and all but $PATH are required.
-GO_GAPIC_OPT=
 IMAGE=
 IN=
 OUT=
+PLUGIN_OPTIONS=
 PROTO_PATH=`pwd`
 
 # Print help and exit.
 function show_help {
   cat << EOF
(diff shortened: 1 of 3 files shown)


gapic.sh
```diff
@@ -50,21 +50,21 @@ EOF
...
   esac
 done
 
 # Ensure that all required options are set.
-if [ -z "$GO_GAPIC_OPT" ] || [ -z "$IMAGE" ] || [ -z "$IN" ] || [ -z "$OUT" ]; then
+if [ -z "$IMAGE" ] || [ -z "$IN" ] || [ -z "$OUT" ]; then
   cat << EOF
 Required argument missing.
-The --go_gapic_opt, --image, --in, and --out arguments are all required.
+The --image, --in, and --out arguments are all required.
```
pongad: `--options` is required too right?

noahdietz: In the Go generator's case, yes, but the plugin will fail with the proper errors message. We are making this flag on the shortcut script generic and optional so that the script maybe used with any of the gapic images.

PR #68, opened 2018-12-14
change gapic gax-go to v2

noahdietz:
Mirroring work for [gapic-generator #2483](https://github.com/googleapis/gapic-generator/issues/2483)

Note: tested by successfully generating & installing a couple of gapics, and by running showcase tests, since the imports aren't covered in the diff testing

cc @jadekler


jadekler:
Thanks! LG!

Issue #70, opened 2018-12-20
support more markdown types for comments


noahdietz:
Currently, the [generator panics](https://github.com/googleapis/gapic-generator-go/blob/d4b1779bb04c1917a9a1cc39242a7fc3b74c33a1/internal/gengapic/markdown.go#L72-L74) when it encounters markdown it doesn't know how to handle. 

We need to support at least the subset of markdown features utilized by Google APIs in comments. For example, the [Cloud Vision protos](https://github.com/googleapis/googleapis/blob/master/google/cloud/vision/v1/product_search_service.proto) contain more complex usage of markdown comments.

A first step would be to simply ignore the markdown and generate the comments in plain text. We could print a warning, but this results in a lot of annoying output, so I'd prefer to silently ignore unknown markdown types as we add support.

noahdietz:
Closing until we get reports of more markdown types present in proto comments that we aren't supporting.

PR #71, opened 2018-12-20
change markdown panic to warning

noahdietz:
Takes initial step of changing the panic at unknown markdown types to a warning, described in #70 

While evaluating generated output for Go Cloud clients, missing `pkgName` in client example headers caused some automated vetting (a comparison to `goimport` output) to fail.

### DIFF
--- a/internal/gengapic/markdown.go
+++ b/internal/gengapic/markdown.go
@@ -20,6 +20,7 @@ package gengapic
 import (
 	"fmt"
 	"log"
+	"os"
 	"strings"
 
 	"github.com/golang-commonmark/markdown"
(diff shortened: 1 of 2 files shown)


internal/gengapic/markdown.go
```diff
@@ -71,6 +72,7 @@ func (m *mdRenderer) plain(t markdown.Token) {
 	default:
 		// TODO(pongad): When going into production, we should turn this to warn.
 		// In the meantime, it's nice to crash to make sure we see it.
-		log.Panicf("unhandled type: %#v", t)
+		log.SetOutput(os.Stderr)
+		log.Printf("unhandled type: %#v", t)
```
pongad: Consider using `%T`. It just prints the type, not the value.

PR #76, opened 2019-01-03
add more default retry codes

### LINKED ISSUE #74: gapic: clients should retry on more codes
Per offline discussion and updates to the generator implementation spec, client retries *should* be supported for the following codes:

* idempotent: `ABORTED`, `INTERNAL` (supported), `UNAVAILABLE` (supported), `UNKNOWN`
* non-idempotent: `UNAVAILABLE`

noahdietz:
Fix #74 

* group RPC retries based on idempotency
* add `ABORTED` and `UNKNOWN` to the existing idempotent retry config
* add retry config on `UNAVAILABLE` for non-idempotent methods


andreamlin:
So this is going into the updated API config spec?

PR #79, opened 2019-01-16
add minimum service config support

noahdietz:
* adds `gapic-service-config` plugin option
* uses service config as **primary** source for the documentation summary header comment
* uses service config as **secondary** source (defers to annotations when present) for API Name, oauth scopes, and default_host 

This is a temporary workaround for: 
1. severely delayed metadata annotations
2. annotation config lacking the doc summary content
3. TBD acceptance of the oauth scope & default_host annotations

This workaround can be removed entirely once the above gaps have been closed. Also, individual workarounds can be removed as each individual gap is closed.

### DIFF
--- a/internal/gengapic/testdata/doc_file.want
+++ b/internal/gengapic/testdata/doc_file.want
@@ -17,7 +17,6 @@
 // Package awesome is an auto-generated package for the
 // Awesome Foo API.
 
-//
 // The Awesome Foo API is really really awesome. It enables the use of Foo
 // with Buz and Baz to acclerate bar.
 package awesome // import "path/to/awesome"
--- a/internal/gengapic/testdata/doc_file.want
+++ b/internal/gengapic/testdata/doc_file.want
@@ -16,6 +16,10 @@
 
 // Package awesome is an auto-generated package for the
 // Awesome Foo API.
+
+//
+// The Awesome Foo API is really really awesome. It enables the use of Foo
+// with Buz and Baz to acclerate bar.
 package awesome // import "path/to/awesome"
 
 import (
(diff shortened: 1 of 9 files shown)


internal/gengapic/testdata/doc_file.want
```diff
@@ -16,6 +16,10 @@
 
 // Package awesome is an auto-generated package for the
 // Awesome Foo API.
+
```
pongad: I think this was a bug in the previous generator. We shouldn't have new newline.

noahdietz:
Ok, I'm cool with creating a new type. I will definitely add the comment on the type def so that we do not expand its definition to depend on it more.

internal/gengapic/testdata/doc_file.want
pongad: Looks like we still have a new line here though?

noahdietz: Oh i thought you meant the empty new line comment that was immediately below it.

Should it be:

```
// Package awesome is an auto-generated package for the
// Awesome Foo API.
//
// Documentation description goes here
package foo
```

or

```
// Package awesome is an auto-generated package for the
// Awesome Foo API.
// Documentation description goes here
package foo
```

or

```
// Package awesome is an auto-generated package for the
// Awesome Foo API.

// Documentation description goes here
package foo
```

pongad: I think it should be the first. The `//` line makes `godoc` renders "Package awesome..." and "Documentation..." on different paragraphs.

The second one will render them in the same paragraph, which works but I think not as good.
The third will render only "Documentation.." because "Package awesome..." will become "orphan" I think.

pongad:
still lgtm

PR #83, opened 2019-02-02
fix repeated bytes bug

### LINKED ISSUE #81: CLI Gen fails on repeated bytes field.
For gapic-showcase, the following line was produced. I think the main culprit is the `[]` before ByteSliceVar.
```go
VerifyTestCmd.Flags().[]ByteSliceVar(&VerifyTestInput.Answers, "answers", [][]byte{}, "")
```

noahdietz:
Fix #81 

* update gcli sanity testing to pull test protos
* fix oneof variable name bug


noahdietz:
Its a workaround for the time being to unblock gapic-showcase stuff. Per #81 , we are just going to ignore them right now, only supporting them from the `--from_file` input.

PR #90, opened 2019-02-12
Samplegen: change CLI flag names to snake_case

hzyi-google:


hzyi-google:
PTAL

cmd/gen-go-sample/main.go
```diff
@@ -356,11 +358,11 @@ func (g *generator) genSample(ifaceName string, methConf GAPICMethod, regTag str
 
 	p("func main() {")
 
-	for i, name := range argNames {
+	for i := 0; i < len(argNames); i++ {
```
pongad: You can probably do `i := range argNames` here. If there's only one variable on the left, only the index is assigned.

hzyi-google:
Thanks for reviewing it!

PR #91, opened 2019-02-14
SampleGen: initialize bytes field correctly

hzyi-google:


cmd/gen-go-sample/bytes.go
```diff
@@ -0,0 +1,21 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+func bytesFmt() func(*generator, string) (string, error) {
```
hzyi-google: We probably could inline this function, but we need to handle copying contents from local files to bytes, which could potentially be put here too.

hzyi-google:
PTAL

cmd/gen-go-sample/bytes.go
```diff
@@ -0,0 +1,21 @@
...
+	return func(g *generator, s string) (string, error) {
```
pongad: I think `g` isn't used right? It should probably be `_`.

cmd/gen-go-sample/gapic.go
```diff
@@ -49,6 +49,7 @@ type SampleParameter struct {
 type SampleAttribute struct {
 	Parameter          string
 	SampleArgumentName string `yaml:"sample_argument_name"`
+	ReadFile bool `yaml:"read_file"`
```
pongad: Run `go fmt`. I think the type should be aligned with the previous line. This probably means we should add a format check to CI as well.

cmd/gen-go-sample/main_test.go
```diff
@@ -48,9 +49,9 @@ func TestUnary(t *testing.T) {
 				`resource_field%bar="mybar"`,
 			},
 			Attributes: []SampleAttribute{
-				{"a.x", "the_x"},
-				{"b", "the_b"},
-				{"resource_field%foo", "the_foo"},
+				{"a.x", "the_x", false},
```
pongad: Since the keyword `false` is not descriptive by itself, I think we should add field name here: `{Parameter: "a.x", SampleArgumentName: "the_x"}` (false is zero value for bool, so it can be skipped).

hzyi-google:
PTAL

PR #93, opened 2019-02-20
Format code and enforce go format in CircleCI

hzyi-google:


hzyi-google:
@pongad This comes from the comments of a PR last week. I'm not sure if formatting should be a separate job though.

pongad:
I think separate job sounds good, so there's a good description about what's wrong if check fails.

PR #94, opened 2019-02-20
gapic: in-repo showcase integration

noahdietz:
* add a `showcase` module with (refactored from gapic-generator) Echo Service tests
* update utils/showcase.bash to use in-repo tests


showcase/go.mod
```diff
@@ -0,0 +1,16 @@
...
+	github.com/googleapis/gax-go v0.0.0-20181219185031-c8a15bac9b9f // indirect
+	github.com/googleapis/gax-go/v2 v2.0.3 // indirect
+	google.golang.org/api v0.1.0
+	google.golang.org/genproto v0.0.0-20190219182410-082222b4a5c5
+	google.golang.org/grpc v1.18.0
+)
+
+replace cloud.google.com/go/showcase => ./gen/cloud.google.com/go/showcase
+
+replace github.com/googleapis/gapic-showcase => ./gen/github.com/googleapis/gapic-showcase
```
pongad: Why do we need this? Shouldn't we just depend on a specific version of `github.com/googleapis/gapic-showcase`?

pongad: Oh, it's because we don't actually have that package do we :P

noahdietz: Well actually you have a point. This is the base module for `github.com/googleapis/gapic-showcase/server/genproto`. We could depend on the tagged release.

Idk why but i got it in my head that these files should be generated at testing time as well.

pongad: Oh right that would make sense.

noahdietz: Removed it & the stub generation as well. Depending on the tagged version now.

PR #96, opened 2019-02-22
Strip 'v' from release tags

jbolinger:
I noticed that most of the go images in GCR start with a 'v'. Unless there's a reason for doing it that way, it may be better to just stick to the convention of using a version number.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -73,7 +73,7 @@ jobs:
           name: Tag the Docker image and push it to Google Container Registry.
           command: |
             if [ -n "$CIRCLE_TAG" ]; then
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -73,7 +73,7 @@ jobs:
           name: Tag the Docker image and push it to Google Container Registry.
           command: |
             if [ -n "$CIRCLE_TAG" ]; then
-              export MAJOR=`echo $CIRCLE_TAG | awk -F '.' '{ print $1; }'`
+              export MAJOR=`echo ${CIRCLE_TAG:1} | awk -F '.' '{ print $1; }'`
```
noahdietz: Can you add this to the MINOR & PATCH tags below as well?

jbolinger: sure

PR #97, opened 2019-02-22
add comment markdown list support

noahdietz:
This PR adds support for markdown lists in proto comments for rendering in the gapic

### DIFF
--- a/internal/gengapic/gengapic.go
+++ b/internal/gengapic/gengapic.go
@@ -469,8 +469,7 @@ func (g *generator) comment(s string) {
 
 	lines := strings.Split(s, "\n")
 	for _, l := range lines {
-		l = strings.TrimSpace(l)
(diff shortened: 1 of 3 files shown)


internal/gengapic/gengapic.go
```diff
@@ -469,8 +469,7 @@ func (g *generator) comment(s string) {
 
 	lines := strings.Split(s, "\n")
 	for _, l := range lines {
-		l = strings.TrimSpace(l)
```
noahdietz: This was removing the leading white space/indentation used to signify list items. There are many more places that trim white space before it is ultimately written, so trailing white space shouldn't be a concern.

noahdietz:
Fixes #70 for known markdown features used by proto comments. We can reopen the issue if we run into new markdown features we don't support.

PR #98, opened 2019-02-26
respect default_host port if present

noahdietz:
Per offline discussions with @lukesneeringer and comparing to [gapic-generator-python](https://github.com/googleapis/gapic-generator-python/blob/master/gapic/templates/$namespace/$name_$version/$sub/services/$service/transports/base.py.j2#L42-L45), a port present in the `google.api.default_host` annotation value should be respected. A default of `443` should be used otherwise.

Note: This was not explicitly stated in the implementation or configuration specification documentation.

### DIFF
--- a/internal/gengapic/client_init.go
+++ b/internal/gengapic/client_init.go
@@ -15,8 +15,6 @@
 package gengapic
 
 import (
-	"fmt"
-	"regexp"
 	"strings"
 
 	"github.com/golang/protobuf/proto"
@@ -26,10 +24,6 @@ import (
 	"google.golang.org/genproto/googleapis/api/annotations"
 	"google.golang.org/genproto/googleapis/rpc/code"
 )
-
-const defaultPort = 443
-
-var portRegex = regexp.MustCompile(":[0-9]{1,5}$")
 
 func (g *generator) clientOptions(serv *descriptor.ServiceDescriptorProto, servName string) error {
 	p := g.printf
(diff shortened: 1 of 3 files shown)


pongad:
I read docs a little more, can we test if we can avoid this logic at all? According to [this gRPC doc](https://github.com/grpc/grpc/blob/master/doc/naming.md), gRPC itself defaults to 443 if the port isn't specified, so we might be able to delegate to gRPC completely.

noahdietz:
Sure I'll try it out. Thanks for doing that research

noahdietz:
Well, I just removed the port `443` from the host in an existing gapic, and attempted to make a request using the corresponding gcli and it just hangs. I replace the port `443` and it is able to make a request. Seems like that documentation hasn't been implemented?

noahdietz:
Different language client but same symptoms: [gax-nodejs#208](https://github.com/googleapis/gax-nodejs/issues/208). Possible issue in gax-go?

internal/gengapic/client_init.go
```diff
@@ -53,9 +59,13 @@ func (g *generator) clientOptions(serv *descriptor.ServiceDescriptorProto, servN
 			return errors.E(err, "cannot read default host")
 		}
 
+		if !portRegex.MatchString(host) {
```
pongad: Would
```go
if !strings.Contains(host, ":") {
  host += ":443"
}
```
work? I'm thinking it doesn't make much sense to define a constant for "443" if it's just going to be used once. Up to you on the regex though :)

noahdietz: Yep, I'll go with the simpler approach. Thanks

PR #102, opened 2019-03-02
SampleGen: generate file input for bytes fields

hzyi-google:
- When `parameters.attributes.read_file` is true, substitute the field with contents from a local file, whose name is what originally set to this field.
- Unlike languages for which we implemented SampleGen in gapic-generator, Go initializes the request object in a nested way, so we have to generate a local variable for the bytes read from the local file. If `SampleArgumentName` is also specified for this parameter, we use that with a `Content` suffix as the name of the local variable. If `SampleArgumentName` is not specified, we use the protobuf field name with a `Content` suffix instead.


hzyi-google:
@vchudnov-g @noahdietz @beccasaurus PTAL

@pongad FYI since we discussed about this. But don't feel obliged to review this PR :)

cmd/gen-go-sample/bytes.go
```diff
@@ -14,8 +14,80 @@
...
+	}
+}
+
+func file(info *fileInfo, buf *bytes.Buffer, g *generator) {
+	vname := info.varName
+	fileName := info.fileName
+	g.imports[pbinfo.ImportSpec{Path: "io"}] = true
+	g.imports[pbinfo.ImportSpec{Path: "io/ioutil"}] = true
+	
+	buf.WriteString("file, err := os.Open(")
```
noahdietz: When there are multiple files to be opened, the `file` & `err` variables will be redeclared in the same statement scope as "new" variables. Compilation would result in a `no new variables on left side of :=` error I believe. Also I'm not sure how the `defer file.Close()` will behave with multiple variables named `file` (maybe nothing because they are pointers to files?).

I think we might want to derive the `file` variable's name like we do the `Content` variable.

cmd/gen-go-sample/testdata/sample_unary.want
```diff
@@ -36,6 +38,25 @@ func sampleUnaryMethod(theX int64, theB string, theFoo string) error {
...
+	file, err := os.Open("path/to/local/file/default.txt")
+	if err != nil {
+		log.Fatalf("Failed to read file: %v", err)
+	}
+	defer file.Close()
+	fileDefaultContent, err := ioutil.ReadAll(file)
+	if err != nil {
+		log.Fatalf("Failed to read file: %v", err)
+	}
+	file, err := os.Open(fileArg)
```
noahdietz: Previous comment WRT file variable naming. This line will produce a `no new variables on left side of :=` error I believe.

vchudnov-g: Agreed. We should either have logic so that only the first time we use the `:=`, or use another structure like
```
var file io.Reader
var err error
file, err = os.Open(...)
...
file, err = os.Open(...)
...
```

cmd/gen-go-sample/bytes.go
```diff
@@ -14,8 +14,80 @@
...
+
 func bytesFmt() func(*generator, string) (string, error) {
 	return func(_ *generator, s string) (string, error) {
 		return "[]byte(" + s + ")", nil
 	}
 }
+
+// The information we need to generate reading all bytes of a file
+// and assign them to a local variable
+type fileInfo struct {
```
vchudnov-g: Here and below, please use the Go commenting style: complete sentences beginning with the name of the object being described.
https://blog.golang.org/godoc-documenting-go-code

cmd/gen-go-sample/bytes.go
noahdietz: After trying it out on the [Go playground](https://play.golang.org/p/DRsYwO0avSy), we should have *no issue* with `defer` (after I alluded that we might). The function call is evaluated immediately, so the variable changing points later wouldn't have an effect.

Initial comment still stands though.

cmd/gen-go-sample/testdata/sample_unary.want
hzyi-google: (Full disclaimer from Go newbee)

> We should either have logic so that only the first time we use the :=
i.e. something like this:

```
file, err := os.Open("the/first/file.txt")
...
file, err = os.Open("another/file.txt")
...
```

noahdietz: Yes that would address @vchudnov-g 's suggestion.

Personally i'd vote for the latter suggestion, explicit variable declaration 100% of the time. I think it would make your logic easier (don't have to save any state to indicate whether or not you've seen a bytes field already). And it's still solid Go.

Very minor bonus, but it's more self-documenting b.c the `io.Reader` is declared explicitly implying something other than an `os.File` could fulfill the example as well.

hzyi-google: Sounds good. I'll change the code to use this approach.

cmd/gen-go-sample/bytes.go
```diff
@@ -14,8 +14,80 @@
...
+	// The text of the file name. If it's a string literal, it's already quoted
+	fileName string
```
pongad: I

cmd/gen-go-sample/bytes.go
```diff
@@ -14,8 +14,80 @@
...
 func bytesFmt() func(*generator, string) (string, error) {
 	return func(_ *generator, s string) (string, error) {
 		return "[]byte(" + s + ")", nil
 	}
 }
+
+// The information we need to generate reading all bytes of a file
+// and assign them to a local variable
+type fileInfo struct {
+	// The text of the file name. If it's a string literal, it's already quoted
```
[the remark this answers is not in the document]
hzyi-google: It could be an `ident` if the `sample_argument_name` is also specified for this `Parameter`?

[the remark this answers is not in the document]
pongad: Oh I see. That makes sense.

hzyi-google:
PTAL

cmd/gen-go-sample/testdata/sample_unary.want
vchudnov-g: The one caveat with that second suggestion: if there is only one file that needs to be opened, then IIUC it's probably more idiomatic to skip the `var` and use the `:=`. But if we have more than one file, so that we need to reuse the vars, using `var` is probably preferred (eg for the reason's @noahdietz suggested)

cmd/gen-go-sample/main.go
```diff
@@ -353,7 +353,13 @@ func (g *generator) genSample(ifaceName string, methConf GAPICMethod, regTag str
 			}
 			buf.WriteByte('\n')
 		}
-		prependLines(&buf, "// ")
+		prependLines(&buf, "// ", false)
+
+		if len(files) > 0 {
```
vchudnov-g: I'd like to hear feedback from other Go experts on preferring `:=` when there's only one file, and using this `var` structure when there's more than one, just for idiomaticity.

(That said, this could be an incremental change in a subsequent PR. Or possibly not worth doing.)

cmd/gen-go-sample/main.go
```diff
@@ -505,7 +511,7 @@ func (g *generator) handleOut(meth *descriptor.MethodDescriptorProto, valSet Sam
 // prependLines adds prefix to every line in b. A line is defined as a possibly empty run
 // of non-newlines terminated by a newline character.
 // If b doesn't end with a newline, prependLines panics.
-func prependLines(b *bytes.Buffer, prefix string) {
+func prependLines(b *bytes.Buffer, prefix string, skipEmptyLine bool) {
```
vchudnov-g: Please comment on the semantics of `skipEmptyLine`. I take it that if true, it does not add the prefix to an empty line, and if false, it does?

cmd/gen-go-sample/bytes.go
```diff
@@ -14,8 +14,79 @@
...
+// fileInfo keeps track of information we need to generate reading all bytes of a file
+// and assign them to a local variable.
+type fileInfo struct {
+	// fileName is the text of the file name. If it's a string literal, it's already quoted.
+	fileName string
+	// varName is the name of the local variable to hold the bytes of the file.
+	varName string
+}
+
+// fileVarName generates a name for the local variable that hold bytes from a local file. It is Only
```
vchudnov-g: nit: s/Only/only/

cmd/gen-go-sample/bytes.go
```diff
@@ -14,8 +14,79 @@
...
+// used when SampleArgumentName is not provided.
+func fileVarName(param string) (string, error) {
+	// TODO: if there are multiple fields whose values are to be copied from local files,
+	// this method cannot generate a unique variable name for each of them when:
+	// 1) SampleArgumentName are not provided for some or all of thse fields
```
vchudnov-g: nit: s/thse/these/

cmd/gen-go-sample/testdata/sample_unary.want
pongad: See my suggestion on the code that generates this. I think we can use `ioutil.ReadFile` to avoid giving file handle a name altogether.

cmd/gen-go-sample/bytes.go
vchudnov-g: Yes! Let's use `ioutil.ReadFile` and not worry about `os.Open` and `defer`s

cmd/gen-go-sample/bytes.go
```diff
@@ -14,8 +14,80 @@
...
+// The information we need to generate reading all bytes of a file
+// and assign them to a local variable
+type fileInfo struct {
+	// The test of the file name. If it's a string literal, it's already quoted
+	fileName string
+	// The name of the local variable to hold the bytes of the file
+	varName string
+}
+
+// Generate a name for the local variable that hold bytes from a local file. Only
```
vchudnov-g: nit: s/Only/only/

cmd/gen-go-sample/testdata/sample_unary.want
vchudnov-g: Agreed. We should use `ioutil.ReadFile`

cmd/gen-go-sample/main.go
vchudnov-g: Let's use `ioutil.ReadFile` and not even have to think about this ;-)

cmd/gen-go-sample/testdata/sample_unary.want
hzyi-google: Ohhh, I thought the suggestion was just about to use `fmt.Fprintf`. I didn't realize using `ioutil.ReadFIle` will get rid of all the troubles!

cmd/gen-go-sample/bytes.go
```diff
@@ -14,8 +14,79 @@
...
+			varName = sc.TokenText()
+		case '[':
+			if r := sc.Scan(); r != scanner.Int {
+				return "", errors.E(nil, "expected int, found %q", sc.TokenText())
+			}
+			if r := sc.Scan(); r != ']' {
+				return "", errors.E(nil, "expected ']', found %q", sc.TokenText())
+			}
+		case scanner.EOF:
+			return snakeToCamel(varName) + "Content", nil
```
vchudnov-g: You use the `"Content"` suffix here and in main. I suggest for consistency and clarity, you define it as a `const` in this file (`fileContentSuffix`?) and use that symbol, rather than having a repeated literal.

cmd/gen-go-sample/bytes.go
hzyi-google: It seems the variable `err` may still be redeclared, would that be a problem?

cmd/gen-go-sample/main.go
```diff
@@ -271,6 +272,40 @@ func (g *generator) genSample(ifaceName string, methConf GAPICMethod, regTag str
 
 	// Some parts of request object are from arguments.
 	for _, attr := range valSet.Parameters.Attributes {
+		if attr.ReadFile {
+			varName, err := fileVarName(attr.Parameter)
```
vchudnov-g: You're performing this non-trivial calculation which may just get thrown away.

For clarity, I would 
```
var varName
if attr.SampleArgumentName != "" {
  varName = snakeToCamel(...)...
} else {
  varName = fileVarName(...)
}
```

cmd/gen-go-sample/main_test.go
```diff
@@ -160,6 +164,8 @@ func initTestGenerator() *generator {
 			{Name: proto.String("b"), Type: typep(descriptor.FieldDescriptorProto_TYPE_STRING)},
 			{Name: proto.String("e"), TypeName: proto.String(".foo.AType.EType")},
 			{Name: proto.String("f"), Type: typep(descriptor.FieldDescriptorProto_TYPE_STRING), OneofIndex: proto.Int32(0)},
+			{Name: proto.String("file_default"), Type: typep(descriptor.FieldDescriptorProto_TYPE_BYTES)},
+			{Name: proto.String("file_arg"), Type: typep(descriptor.FieldDescriptorProto_TYPE_BYTES)},
```
vchudnov-g: A request field whose contents we happen to read from a file will not necessarily have "file" in the name. I suggest renaming this to "data_bob" for clarity, and making the sample config have `SampleArgumentName: "bob_file"`. That will be a more realistic sample of sample generation usage, I think.

Similarly, make "file_default" into "data_alice". This would then suggest that the variable with contents of the file, when no filename arg is provided, should be named "data_alice" and not "data_alice_contents", whereas for "bob", when the filename arg _is_ provided, it _should_ be "bob_file_contents". WDYT?

If I'm not being clear here, just rename the fields in this proto and the sample config as I suggest and look at the .want file. Do the names make sense?

hzyi-google: Good point.
Taking one more step further, do you think it makes sense to always use the field name, regardless of whether `SampleArgumentName` is given?

cmd/gen-go-sample/testdata/sample_unary.want
```diff
@@ -36,6 +38,33 @@ func sampleUnaryMethod(theX int64, theB string, theFoo string) error {
...
+
+	var file io.Reader
+	var err error
+
+	file, err = os.Open("path/to/local/file/default.txt")
+	if err != nil {
+		log.Fatalf("Failed to read file: %v", err)
+	}
+	defer file.Close()
+	var fileDefaultContent []byte
```
beccasaurus: What does "file default content" mean?

cmd/gen-go-sample/testdata/sample_unary.want
```diff
@@ -36,6 +38,33 @@ func sampleUnaryMethod(theX int64, theB string, theFoo string) error {
 	// theX := 42
 	// theB := "foobar"
 	// theFoo := "myfoo"
+	// fileArg := "path/to/local/file/arg.txt"
+
+	var file io.Reader
+	var err error
+
+	file, err = os.Open("path/to/local/file/default.txt")
```
beccasaurus: This should presumably be `fileArg` instead of the literal string?

cmd/gen-go-sample/testdata/sample_unary.want
```diff
@@ -19,14 +19,16 @@ import(
...
+ "io/ioutil"
  "log"
 
 foo "path.to/client/foo"
 foopb "path.to/pb/foo"
 )
 // [START awesome_region]
 
-func sampleUnaryMethod(theX int64, theB string, theFoo string) error {
+func sampleUnaryMethod(theX int64, theB string, theFoo string, fileArg string) error {
```
beccasaurus: What is the YAML equivalent of the definition of this function?

I'm having a hard time reviewing it without clear YAML inputs

cmd/gen-go-sample/testdata/sample_unary.want
```diff
@@ -36,6 +38,33 @@ func sampleUnaryMethod(theX int64, theB string, theFoo string) error {
...
+		log.Fatalf("Failed to read file: %v", err)
+	}
+	defer file.Close()
+	var fileDefaultContent []byte
+	fileDefaultContent, err = ioutil.ReadAll(file)
+	if err != nil {
+		log.Fatalf("Failed to read file: %v", err)
+	}
+
+	file, err = os.Open(fileArg)
```
beccasaurus: I'd like to use this:

```go
fileBytes, err := ioutil.ReadFile(file)
if err != nil {
    return err
}
```

Examples:
 - https://github.com/GoogleCloudPlatform/golang-samples/blob/master/videointelligence/video_analyze/video_analyze.go#L37-L40
 - https://github.com/GoogleCloudPlatform/golang-samples/blob/master/speech/caption/caption.go#L107-L110 (uses data)

We use that simple `ReadFile` everywhere I can find

cmd/gen-go-sample/testdata/sample_unary.want
```diff
@@ -36,6 +38,33 @@ func sampleUnaryMethod(theX int64, theB string, theFoo string) error {
 	// theX := 42
 	// theB := "foobar"
 	// theFoo := "myfoo"
+	// fileArg := "path/to/local/file/arg.txt"
+
+	var file io.Reader
+	var err error
```
beccasaurus: `err` Doesn't need to be defined ahead of time.

```go
fileBytes, err := ioutil.ReadFile(file)
if err != nil {
    return err
}
```

cmd/gen-go-sample/testdata/sample_unary.want
beccasaurus: There doesn't need to be any additional code to read the "default"

The default is merely a string file path. The code sample should do nothing other than `ReadFile` the bytes from the provided string parameter to the file path

cmd/gen-go-sample/bytes.go
pongad: It shouldn't be. The `:=` should work as long as at least one variable is new. For full detail, see [spec](https://golang.org/ref/spec#Short_variable_declarations) 😄

cmd/gen-go-sample/testdata/sample_unary.want
hzyi-google: There are two file inputs here.

cmd/gen-go-sample/testdata/sample_unary.want
hzyi-google: Here's the yaml:
```
- parameters:
    defaults:
    - file_arg = "path/to/local/file/arg.txt"
    - file_default = "path/to/local/file/default.txt"
    attributes:
    - parameter: file_arg:
        sample_argument_name: file_arg
        read_file: true
    - parameter: file_default:
        read_file: true
```

cmd/gen-go-sample/testdata/sample_unary.want
hzyi-google: Sounds good!

cmd/gen-go-sample/main_test.go
vchudnov-g: As discussed offline: sounds good. We can use a simple algorithm to disambiguate in case of collisions (and we should try to have equivalent algorithms in all our languages). If the user doesn't like how we disambiguate, they'll have to provide a `SampleArgumentName`!

hzyi-google:
PTAL @vchudnov-g @beccasaurus

cmd/gen-go-sample/bytes.go
```diff
@@ -14,8 +14,74 @@
...
+			}
+		case scanner.EOF:
+			return snakeToCamel(varName), nil
+		default:
+			return "", errors.E(nil, "unhandled rune: %c", r)
+		}
+	}
+}
+
+func file(info *fileInfo, buf *bytes.Buffer, g *generator) {
```
vchudnov-g: Let's make this function name more descriptive, something like `readFile`. It might also be good to have a brief function comment clarifying this generates code that reads from the file configured in info.

cmd/gen-go-sample/main.go
```diff
@@ -271,6 +272,43 @@ func (g *generator) genSample(ifaceName string, methConf GAPICMethod, regTag str
...
 	// Some parts of request object are from arguments.
 	for _, attr := range valSet.Parameters.Attributes {
+		if attr.ReadFile {
+			var varName string
+			if attr.SampleArgumentName != "" {
+				varName = snakeToCamel(attr.SampleArgumentName) + fileContentSuffix
+			} else {
+				varName, err = fileVarName(attr.Parameter)
+			}
+			if err != nil {
```
vchudnov-g: consider having this check inside the else, which is where the error is generated.

hzyi-google:
Thank you all for reviewing this!

PR #108, opened 2019-03-11
gengapic: add import comment in example snippets

### LINKED ISSUE #107: gapic: comment type imports beside example code snippets
_Copied from [gapic-generator #2620](https://github.com/googleapis/gapic-generator/issues/2620)_

> There has been feedback indicating confusion around the difference between the genproto pkg and the gapic pkg. Because of naming & alias name misunderstandings, some see the former as the "old" client and the latter as the "new" one, which then gets folks confused as to why they need both imported.
> 
> The genproto pkg contains the generated types for the API protos, while the gapic pkg contains the generated client that consumes them.
> 
> To help clear this up, the generated example snippets should include a comment above the use of the genproto types indicating where it comes from.

cc @broady @jadekler

noahdietz:
Fixes #107 
Fixes [google-cloud-go#1356](https://github.com/googleapis/google-cloud-go/issues/1356)

In generated example snippets, add comment beside request payloads indicating import path of struct and for the iterator type.

@broady @jadekler can you review?

### DIFF
--- a/internal/gengapic/example.go
+++ b/internal/gengapic/example.go
@@ -71,19 +71,34 @@ func (g *generator) exampleMethod(pkgName, servName string, m *descriptor.Method
 	g.imports[inSpec] = true
 
 	p("func Example%sClient_%s() {", servName, m.GetName())
+
+	pf, err := g.pagingField(m)
+	if err != nil {
+		return err
+	}
+
+	if *m.OutputType != emptyType {
+		p("// import \"%s\" %s", inSpec.Path, inSpec.Name)
+		if pf == nil {
+			p("")
+		}
+	}
+
+	if pf != nil {
+		p("// import \"google.golang.org/api/iterator\"")
+		p("")
+	}
+
 	g.exampleInitClient(pkgName, servName)
 
 	if !m.GetClientStreaming() && !m.GetServerStreaming() {
 		p("")
-		p("// %s is import %s.", inSpec.Name, inSpec.Path)
 		p("req := &%s.%s{", inSpec.Name, inType.GetName())
 		p("  // TODO: Fill request struct fields.")
 		p("}")
 	}
 
-	if pf, err := g.pagingField(m); err != nil {
-		return err
-	} else if pf != nil {
+	if pf != nil {
 		g.examplePagingCall(m)
 	} else if *m.OutputType == lroType {
 		g.exampleLROCall(m)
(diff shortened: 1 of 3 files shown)


broady:
Sorry to be a pain, and I think this is harder to implement, is something like this easily doable?

```
func ExampleFooClient_BidiThings() {
	// import "mypackagepb" mypackage
	// import "google.golang.org/api/iterator"

	ctx := context.Background()
	c, err := Bar.NewFooClient(ctx)

// snip

}
```

noahdietz:
PTAL

internal/gengapic/testdata/foo_example.want
```diff
@@ -66,6 +70,9 @@ func ExampleFooClient_GetBigThing() {
 }
 
 func ExampleFooClient_GetManyThings() {
+	// import "mypackage" mypackagepb
```
broady: other way around - quotes around the package import, the name its imported as is unquoted

noahdietz: That is how this is implemented, the example is just a mediocre expression of it. `mypackagepb` is in fact the name while `"mypackage"` is the import path. :sweat_smile:

broady: in that case, they need to be swapped!

broady: (sorry, i see that i messed that up in my suggestion above. whoops!)

internal/gengapic/example.go
```diff
@@ -71,6 +71,24 @@ func (g *generator) exampleMethod(pkgName, servName string, m *descriptor.Method
...
 
 	p("func Example%sClient_%s() {", servName, m.GetName())
+
+	pf, err := g.pagingField(m)
+	if err != nil {
+		return err
+	}
+
+	if *m.OutputType != emptyType {
+		p("// import \"%s\" %s", inSpec.Path, inSpec.Name)
```
broady:
```suggestion
		p(`// import %s "%s"`, inSpec.Name, inSpec.Path)
```

internal/gengapic/testdata/foo_example.want
noahdietz: Done

internal/gengapic/example.go
noahdietz: Thanks! committed it with `.want` file changes

broady:
LGTM++

PR #115, opened 2019-03-26
gengapic: add http-based request header param support

### LINKED ISSUE #114: gapic: add request header params per HTTP annotation
When a method includes the `google.api.http` option and the HTTP path includes a request parameter, the request parameter must be duplicated as a request header param.

In the below example...
```
option (google.api.http) = {
      post: "/v1/{parent=projects/*/locations/*/keyRings/*/cryptoKeys/*}/cryptoKeyVersions"
```
...the `parent` field would be duplicated as a request header param.

noahdietz:
Fixes #114 

Adds insertion of request header params based on HTTP extension's pattern value

### DIFF
--- a/internal/gengapic/gengapic.go
+++ b/internal/gengapic/gengapic.go
@@ -444,15 +444,14 @@ func (g *generator) emptyUnaryCall(servName string, m *descriptor.MethodDescript
 
 func (g *generator) insertMetadata(m *descriptor.MethodDescriptorProto) {
 	if headers := parseRequestHeaders(m); len(headers) > 0 {
-		field := headers[0][1]
-		pairs := pair("x-goog-request-params", field)
-
-		for _, h := range headers[1:] {
-			field = h[1]
-			pairs += fmt.Sprintf(", %s", pair("x-goog-request-params", field))
+		var pairs strings.Builder
+		for _, h := range headers {
+			field := h[1]
+			fmt.Fprintf(&pairs, ", %s", pair("x-goog-request-params", field))
 		}
+		p := pairs.String()[2:]
 
-		g.printf("md := metadata.Pairs(%s)", pairs)
+		g.printf("md := metadata.Pairs(%s)", p)
 		g.printf("ctx = insertMetadata(ctx, c.xGoogMetadata, md)")
 
 		return
@@ -461,9 +460,9 @@ func (g *generator) insertMetadata(m *descriptor.MethodDescriptorProto) {
 	g.printf("ctx = insertMetadata(ctx, c.xGoogMetadata)")
 }
 
-func pair(key, val string) string {
+func pair(key, field string) string {
 	// TODO(ndietz) the formatted field accessor here only works for top-level fields
-	return fmt.Sprintf("%q, fmt.Sprintf(%q, %q, req.Get%s())", key, "%s=%v", val, snakeToCamel(val))
+	return fmt.Sprintf("%q, fmt.Sprintf(%q, req.Get%s())", key, field+"=%v", snakeToCamel(field))
 }
 
 func (g *generator) appendCallOpts(m *descriptor.MethodDescriptorProto) {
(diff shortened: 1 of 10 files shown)


internal/gengapic/gengapic.go
```diff
@@ -439,10 +442,30 @@ func (g *generator) emptyUnaryCall(servName string, m *descriptor.MethodDescript
...
+
+		return
+	}
+
 	g.printf("ctx = insertMetadata(ctx, c.xGoogMetadata)")
 }
 
+func pair(key, val string) string {
+	// TODO(ndietz) the formatted field accessor here only works for top-level fields
+	return fmt.Sprintf("%q, fmt.Sprintf(%q, %q, req.Get%s())", key, "%s=%v", val, snakeToCamel(val))
```
pongad: Nice :P

internal/gengapic/gengapic.go
```diff
@@ -439,10 +442,30 @@ func (g *generator) emptyUnaryCall(servName string, m *descriptor.MethodDescript
...
 }
 
-func (g *generator) insertMetadata() {
+func (g *generator) insertMetadata(m *descriptor.MethodDescriptorProto) {
+	if headers := parseRequestHeaders(m); len(headers) > 0 {
+		field := headers[0][1]
+		pairs := pair("x-goog-request-params", field)
+
+		for _, h := range headers[1:] {
+			field = h[1]
```
pongad: I'd change this slightly:
```
var pairs strings.Builder
for _, h := range headers {
   fmt.Fprintf(&pairs, ", %q, fmt.Sprintf ...", ...)
}
pairs.String()[2:]
```

internal/gengapic/testdata/method_GetBigThing.want
```diff
@@ -1,5 +1,6 @@
 func (c *FooClient) GetBigThing(ctx context.Context, req *mypackagepb.InputType, opts ...gax.CallOption) (*GetBigThingOperation, error) {
-	ctx = insertMetadata(ctx, c.xGoogMetadata)
+	md := metadata.Pairs("x-goog-request-params", fmt.Sprintf("%s=%v", "field_name", req.GetFieldName()), "x-goog-request-params", fmt.Sprintf("%s=%v", "other", req.GetOther()))
```
pongad: I suppose it's possible to optimize this a little to `fmt.Sprintf("field_name=%v", req.GetFieldName())`. Is the current way how it's currently implemented?

internal/gengapic/gengapic.go
```diff
@@ -532,3 +555,26 @@ func snakeToCamel(s string) string {
 	}
 	return sb.String()
 }
+
+func parseRequestHeaders(m *descriptor.MethodDescriptorProto) [][]string {
+	eHTTP, err := proto.GetExtension(m.GetOptions(), annotations.E_Http)
+	if err != nil {
```
pongad: Now that I think of it, I think we should explicitly check for `ErrMissingExtension`.

internal/gengapic/testdata/method_GetBigThing.want
noahdietz: Yeah, this is how it’s currently generated. But I’m happy to change it. It won’t have a big effect on the diffs

internal/gengapic/gengapic.go
noahdietz: Just curious, why do you think it’s important to make the distinction? If it’s not there, we just move on. I guess there could be some errors from parsing the bytes but I’m not sure how to reproduce that.

internal/gengapic/gengapic.go
noahdietz: 🙃
I haven’t seen an http transcoding option contain a nested field request param...figure i’d deal with it if it ever happens

internal/gengapic/testdata/method_GetBigThing.want
pongad: I'll leave that up to you :)

noahdietz: I took your suggestion :)

internal/gengapic/gengapic.go
pongad: That's a good question.I agree it'd be hard to repro this. Errors from parsing the bytes seem the most plausible. I'm can think of something like this:
- We accidentally change the annotation in a backward incompatible way
- Now if we recompile the generator, the descriptor using the old annotation is invalid.
- With the current code, we assume the annotation is missing and move on.

Situations like this is hopefully rare. I find checking for specific errors a good practice so that we can at least complain about it. Up to you though :)

noahdietz: That is reasonable to me! I'll add it. Thanks

noahdietz: Done

PR #124, opened 2019-04-10
gapic: add Response to gen'd iterator type

### LINKED ISSUE #123: gapic: gen'd iterator type should expose original response fields
Today, if a method is inferred to be paginated (based on it's output message schema), the response type of the gen'd method will be a new `{Message}Iterator` type. This new gen'd type does **not** include all of the original response message fields. It only  includes a slice of the `repeated` field  type and `iterator.PageInfo`.

After  offline discussions, pagination  should  not  "get  in the way"  of using the original response object. Thus, the gen'd iterator type should expose the original response fields either in their entirety or as a subset of the fields that don't overlap with pagination logic.

noahdietz:
Fix #123 

* add `Response *foopb.Foo` field to gen'd `FooIterator` type to preserve original response
* set `it.Response` to `resp` of fetch
* add `make golden` target for updating `want` files

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,6 +15,9 @@ test-go-cli:
 test-gapic:
 	go test github.com/googleapis/gapic-generator-go/internal/gengapic
 
+golden:
+	go test github.com/googleapis/gapic-generator-go/internal/gengapic -update_golden
+
 test:
 	go test ./...
 	./utils/showcase.bash
--- a/internal/gengapic/gengapic_test.go
+++ b/internal/gengapic/gengapic_test.go
@@ -149,6 +149,11 @@ func TestGenMethod(t *testing.T) {
 			},
 		},
 	}
+	paginatedField := &descriptor.FieldDescriptorProto{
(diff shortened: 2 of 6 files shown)


noahdietz:
> the user can get to it by iterator.Response.TotalResultsCount right?
@pongad yes that's right.

And this example happens to be for an extra field related to the repeated field in the response, but really if there were **any** other field(s) on the response message, we'd want to to have access. The key here is that _pagination helpers shouldn't get in the way_.

WDYT @jba ? Open to other ideas! This seemed like the most straight forward.

jba:
AFK so can't see the change, but the iterator spec should make it clear
that additional fields on the iterator are fine.

noahdietz:
Ok so the [iterator spec](https://github.com/googleapis/google-cloud-go/wiki/Iterator-Guidelines) states:
> Iterators may have fields and methods other than those described here.

I'd say we are safe WRT the spec!

Thanks y'all, I will submit this change.

PR #127, opened 2019-04-16
gcli: fix paged cmd iteration

### LINKED ISSUE #126: gencli: paged RPC commands only return 1 item regardless of page_size
The command generated for paged RPCs makes a duplicate call that doesn't make sense. The duplicate call needs to be removed.

noahdietz:
Fixes #126 

* fix paged cmd iteration
* remove broken integration test from gcli testing make target


internal/gencli/cmd_file.go
```diff
@@ -277,24 +277,21 @@ var {{$methodCmdVar}} = &cobra.Command{
 		{{ end }}
 		{{ else if and .Paged (not .IsLRO ) }}
 		// get requested page
-		page, err := iter.Next()
-		if err != nil {
+		data := make(map[string]interface{})
+		items := []interface{}{}
```
pongad: var items []interface{}

internal/gencli/cmd_file.go
```diff
@@ -277,24 +277,21 @@ var {{$methodCmdVar}} = &cobra.Command{
...
+		for i := {{ $inputVar }}.PageSize; i > 0; i-- {
```
pongad: Do we want this to just print an item from one page? I think `i := 0; i < input.PageSize; i++` is better. Note that we have no hard guarantee that we only make one RPC here, since a server might return fewer than PageSize elements. Is that OK for our use case?

noahdietz: I agree its better, but some `PageSize` field definitions are `int32` (i.e. all of them gapic-showcase). Initializing a new variable `i := 0` creates an `int`, resulting in a type mismatch between the `int32` `PageSize` and `i`. Flipping the iteration logic like so means the `0` in `i > 0` is created as an `int32`.

pongad: Yeah you're right. Let's put that in a comment then.

noahdietz: > Note that we have no hard guarantee that we only make one RPC here, since a server might return fewer than PageSize elements. Is that OK for our use case?

I am OK with this. There is no way for the consumer to know if the server returned the `PageSize` number of elements in 1 RPC anyways, so we can't do anything about it.

internal/gencli/cmd_file.go
noahdietz: Done!

internal/gencli/cmd_file.go
noahdietz: Comment added regarding PageSize decrement semantics.

noahdietz:
Updated, PTAL!

PR #130, opened 2019-04-30
gcli: refactor with protoreflect

noahdietz:
* use protoreflect/desc as descriptor API
* fix `go_cli` integration testing script + readd it to `make test-go-cli`
* fix bug generating comments containing double quotes
* refactor some `oneof` variable name generation code (still pretty wonky, need to seriously consider redoing some stuff)


internal/gencli/gencli.go
```diff
@@ -90,44 +93,44 @@ func Gen(genReq *plugin.CodeGeneratorRequest) (*plugin.CodeGeneratorResponse, er
...
-					continue
-				}
-
-				g.addComments(f, loc)
-			}
-
-			g.services = append(g.services, f.Service...)
+	for _, f := range req.FileToGenerate {
+		file, ok := g.protos[f]
+		if !ok {
```
pongad: Should this error?

noahdietz: yes, thanks for catching this. Done

PR #133, opened 2019-05-07
chore: add basic gen process explanation

noahdietz:
Adds a high level explanation of the code generation process for use by those interested in using `gapic-generator-go` as a reference. Per discucssion with lukesneeringer


noahdietz:
Got an offline LGTM from lukesneeringer. Merging.

PR #138, opened 2019-05-30
gapic: make iterator.Response an interface

noahdietz:
I realized a little late that generated Iterator structs are "singletons" based on the **repeated field type**. There is only one iterator class generated for a `repeated Foo foo = 1;` field, `FooIterator`. Any RPC that implements the paging interface that pages over such a field will use the same `FooIterator` struct. However, the RPC responses might (most likely) differ. ListFoo and ListFooAgain might have ListFooResponse and ListFooAgainResponse response types, resulting in the recently added `Response {RPC response type pb}` field breaking.

An example exists today in the Cloud Vision API between the [ListProductsResponse](https://github.com/googleapis/googleapis/blob/152dabdfea620675c2db2f2a74878572324e8fd2/google/cloud/vision/v1/product_search_service.proto#L456-L463) and the [ListProductsInProductSetResponse](https://github.com/googleapis/googleapis/blob/152dabdfea620675c2db2f2a74878572324e8fd2/google/cloud/vision/v1/product_search_service.proto#L683-L690).

Changing the returned iterator type today would break the generated API surface. But having the original response is also important. So i made the generated Iterator's `Response` field an `interface{}` requiring the caller to cast it to the proper type to access it. Not the best experience but maintains flexibility. Generics would be nice here.

cc @jadekler for opinion on this solution.

### DIFF
--- a/internal/gengapic/paging.go
+++ b/internal/gengapic/paging.go
@@ -27,9 +27,6 @@ import (
 type iterType struct {
 	iterTypeName, elemTypeName string
 
-	resType string
-	resSpec pbinfo.ImportSpec
-
 	// If the elem type is a message, elemImports contains pbinfo.ImportSpec for the type.
 	// Otherwise, len(elemImports)==0.
 	elemImports []pbinfo.ImportSpec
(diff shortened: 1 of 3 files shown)


jadekler:
Honestly don't full grok, or have cycles to page this in. Deferring to @pongad.

PR #140, opened 2019-06-05
SampleGen: iterate over a map field in response handling

hzyi-google:
Porting https://github.com/googleapis/gapic-generator/pull/2593 over to Go.


hzyi-google:
PTAL

cmd/gen-go-sample/inittree.go
```diff
@@ -106,6 +109,29 @@ func (t *initTree) get(k string, info pbinfo.Info) (*initTree, error) {
...
+						if f2.GetName() == "key" {
+							// key is always a primitive type
+							v.typ.keyType = &initType{prim: f2.GetType()}
+						}
+						if f2.GetName() == "value" {
+							if tn2 := f.GetTypeName(); tn2 == "" {
+								// value is a primitive type
+								v.typ.valueType = &initType{prim: f2.GetType()}
+							} else if typ := info.Type[tn2]; typ == nil {
+								return nil, errors.E(nil, "cannot find descriptor of %q", f.GetTypeName())
```
noahdietz: Do you mean `f2.GetTypeName()` here instead of `f`?

cmd/gen-go-sample/inittree.go
```diff
@@ -106,6 +109,29 @@ func (t *initTree) get(k string, info pbinfo.Info) (*initTree, error) {
...
+			if d, ok := typ.(*descriptor.DescriptorProto); ok {
+				if d.GetOptions().GetMapEntry() == true {
+					// type is a map entry
+					for _, f2 := range d.Field {
+						if f2.GetName() == "key" {
+							// key is always a primitive type
+							v.typ.keyType = &initType{prim: f2.GetType()}
+						}
+						if f2.GetName() == "value" {
+							if tn2 := f.GetTypeName(); tn2 == "" {
```
noahdietz: Do you mean `f2.GetTypeName()` here? `f` is the outer Field variable.

Nit: Have you considered using [GetType](https://godoc.org/github.com/golang/protobuf/protoc-gen-go/descriptor#FieldDescriptorProto.GetType) & comparing to the type const? Personally, I think it would be easier to grok this code. (I know that TypeName is only set for Messages & Enums, but that requires looking at the FieldDescriptorProto.type_name comments)

```go
v.typ.valueType = &initType{prim: f2.GetType()}

if f2.GetType() == descriptor.FieldDescriptorProto_TYPE_MESSAGE {
  tn2 := f2.GetTypeName()
  typ := info.Type[tn2]

  if typ == nil {
    return nil, errors.E(nil, "cannot find descriptor of %q", tn2)
  }

  v.typ.valueType = &initType{desc: typ}
}
```

cmd/gen-go-sample/out.go
```diff
@@ -86,7 +86,13 @@ func writeOutputSpec(out OutputSpec, st *symTab, gen *generator) error {
 	}
 	if l := out.Loop; l != nil {
 		used++
-		err = writeLoop(l, st, gen)
+		if l.Collection != "" {
```
noahdietz: Suggestion, refactor to

```go
err = errors.E(nil, "")
if l.Collection != "" {
  err = writeLoop(l, st, gen)
} else if l.Map != "" {
  err = writeMap(l, st, gen)
}
```

cmd/gen-go-sample/inittree.go
```diff
@@ -106,6 +109,29 @@ func (t *initTree) get(k string, info pbinfo.Info) (*initTree, error) {
 		} else {
 			// type is a message
 			v.typ.desc = typ
+
+			if d, ok := typ.(*descriptor.DescriptorProto); ok {
+				if d.GetOptions().GetMapEntry() == true {
+					// type is a map entry
+					for _, f2 := range d.Field {
+						if f2.GetName() == "key" {
```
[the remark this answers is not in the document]
hzyi-google: Nice! I never thought about `switch` here. Good call.

cmd/gen-go-sample/inittree.go
hzyi-google: Yes!

cmd/gen-go-sample/inittree.go

hzyi-google:
PTAL

cmd/gen-go-sample/inittree.go
```diff
@@ -114,19 +114,23 @@ func (t *initTree) get(k string, info pbinfo.Info) (*initTree, error) {
...
+							if t2 := f2.GetType(); t2 == descriptor.FieldDescriptorProto_TYPE_MESSAGE || t2 == descriptor.FieldDescriptorProto_TYPE_ENUM {
+								// value is a protobuf message or enum
+								tn2 := f2.GetTypeName()
+								typ2 := info.Type[tn2]
+								if typ2 == nil {
+									return nil, errors.E(nil, "cannot find descriptor of %q", tn2)
+								}
+								
+								v.typ.valueType = &initType{desc: typ2}
 							} else {
```
noahdietz: So WRT to using an `else` statement here, I tend to take the conventions described in [Effective Go](https://golang.org/doc/effective_go.html#if) beyond `error` cases. What I mean by that is that the **basic** flow of the code should *not* be in an `else` and special cases should be handled by `if`/`else if`. This is why in my suggested snippet for this section I had something like

```go
v.typ.valueType = &initType{prim: t2}

if t2 := f2.GetType(); t2 == Message ...
  ...
  v.typ.valueType = &initType{desc: typ2}
}
```

The primitive type case becomes the basic case (because there is no extra work necessary other than assignment) and the special case of a Message or Enum type becomes handled specifically (it has more involved logic). I think this is easier to read because I don't need to go back up to the original `if` to understand what was being evaluated to get me in the `else`, it reads straight down. Combined with the comment you added `// value is a primitive type` I think this is much easier to read.

Sorry for being so nit-picky and if you feel this is unnecessary feel free to say "thanks, but no thanks".

hzyi-google: Thanks and thanks :)

I didn't know this is the convention adopted by Go, and therefore I completely missed this difference in your code snippet.

Updated :)

cmd/gen-go-sample/inittree.go
```diff
@@ -119,6 +119,9 @@ func (t *initTree) get(k string, info pbinfo.Info) (*initTree, error) {
 							// key is always a primitive type
 							v.typ.keyType = &initType{prim: f2.GetType()}
 						case "value":
+							// value is a primitive type
+							v.typ.valueType = &initType{prim: t2}
```
noahdietz: t2 := f2.GetType(), remove the variable init in the `if` and use it there too

hzyi-google: (Tried to edit the file using directory in github, turns out that's a bad decision.)

cmd/gen-go-sample/main_test.go
```diff
@@ -299,6 +347,61 @@ func initTestGenerator() *generator {
 	}
 }
 
+func labelp(l descriptor.FieldDescriptorProto_Label) *descriptor.FieldDescriptorProto_Label {
+	return &l
+}
+
+// mapTypeAndField creates the generated MapEntry protobuf message and the actually map field.
```
vchudnov-g: s/actually/actual/

cmd/gen-go-sample/inittree.go
```diff
@@ -43,6 +43,9 @@ type initType struct {
 
 	repeated bool
 
+	keyType   *initType
```
vchudnov-g: Maybe add a brief comment that these fields are used for maps?

cmd/gen-go-sample/main_test.go
```diff
@@ -186,16 +187,76 @@ func TestEmpty(t *testing.T) {
...
+	aType := &descriptor.DescriptorProto{
+		Name: proto.String("AType"),
+		Field: []*descriptor.FieldDescriptorProto{
+			{Name: proto.String("x"), Type: typep(descriptor.FieldDescriptorProto_TYPE_INT64)},
+			{Name: proto.String("y"), Type: typep(descriptor.FieldDescriptorProto_TYPE_FLOAT)},
+		},
+		EnumType: []*descriptor.EnumDescriptorProto{eType},
+	}
+
+	mapType, mapField := mapTypeAndField("mappy_map", ".foo.InputType", typep(descriptor.FieldDescriptorProto_TYPE_STRING), typep(descriptor.FieldDescriptorProto_TYPE_MESSAGE), ".foo.AType")
```
vchudnov-g: nit: maybe break the line?

cmd/gen-go-sample/main_test.go
```diff
@@ -299,6 +347,61 @@ func initTestGenerator() *generator {
 	}
 }
 
+func labelp(l descriptor.FieldDescriptorProto_Label) *descriptor.FieldDescriptorProto_Label {
+	return &l
+}
+
+// mapTypeAndField creates the generated MapEntry protobuf message and the actually map field.
+// If valueType is enum or message, vTypeName must not be empty.
+func mapTypeAndField(
```
vchudnov-g: nit: this will read better if you  precede it with a verb, eg `createMapTypeAndField`

cmd/gen-go-sample/main_test.go
```diff
@@ -299,6 +347,61 @@ func initTestGenerator() *generator {
...
+		Field: []*descriptor.FieldDescriptorProto{
+			{Name: proto.String("key"), Type: keyType},
+			vf,
+		},
+		Options: &descriptor.MessageOptions{
+			MapEntry: proto.Bool(true),
+		},
+	}
+
+	mf := &descriptor.FieldDescriptorProto{
```
vchudnov-g: nit: suggest s/`mf`/`mapField`/ for symmetry with `mapEntry`and because I have a bias against too-short names

cmd/gen-go-sample/out.go
```diff
@@ -86,7 +86,12 @@ func writeOutputSpec(out OutputSpec, st *symTab, gen *generator) error {
 	}
 	if l := out.Loop; l != nil {
 		used++
-		err = writeLoop(l, st, gen)
+		err = errors.E(nil, "")
+		if l.Collection != "" {
```
vchudnov-g: This might be the clearest, most central place to check that not both `.Collection` and `.Map` are specified, instead of the distributed checks you have elsewhere.

cmd/gen-go-sample/out.go
```diff
@@ -145,6 +150,14 @@ func writePrint(pFmt string, pArgs []string, st *symTab, gen *generator) error {
 }
 
 func writeLoop(l *LoopSpec, st *symTab, gen *generator) error {
+	if l.Variable == "" {
+		return errors.E(nil, "variable not specified for looping over arrays")
+	}
+
+	if l.Map != "" || l.Key != "" || l.Value != "" {
```
vchudnov-g: Checks like this break encapsulation: why should the loop-specific function check any entries for maps? This is why I suggest the more centralized check above.

PR #143, opened 2019-06-21
gapic: fix header request params

noahdietz:
#114 did this incorrectly based on [aip/4222](https://aip.dev/client-libraries/4222). This PR fixes that.

* Multiple key-value pairs are concatenated with `&`
* The entire "query" is URL encoded (prevents stuff like https://github.com/googleapis/gax-nodejs/issues/520)
* Implements nested request parameter access

### DIFF
--- a/internal/gengapic/gengapic.go
+++ b/internal/gengapic/gengapic.go
@@ -444,14 +444,17 @@ func (g *generator) insertMetadata(m *descriptor.MethodDescriptorProto) error {
 	}
 
 	if len(headers) > 0 {
-		var pairs strings.Builder
+		var formats, values strings.Builder
 		for _, h := range headers {
 			field := h[1]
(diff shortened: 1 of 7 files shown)


internal/gengapic/gengapic.go
```diff
@@ -464,9 +468,13 @@ func (g *generator) insertMetadata(m *descriptor.MethodDescriptorProto) error {
 	return nil
 }
 
-func pair(key, field string) string {
-	// TODO(ndietz) the formatted field accessor here only works for top-level fields
-	return fmt.Sprintf("%q, fmt.Sprintf(%q, req.Get%s())", key, field+"=%v", snakeToCamel(field))
+func buildAccessor(field string) string {
+	var ax string
```
pongad: Let's use `strings.Builder`?

noahdietz: Done!

PR #145, opened 2019-06-27
gapic: encode metadata pairs separately

### LINKED ISSUE #144: x-goog-request-params key and value need to be escaped separately
This was probably already on the mind of @noahdietz, but I wanted to make sure it didn't get lost.

#143 escapes the whole key=value pair, meaning that the = is escaped. Instead, the key and value should be escaped separately. In reality, we're *probably* fine not to escape the key, as that's fixed at configuration time and highly unlikely to ever contain character that need escaping. It could certainly be escaped at generation time rather than on every request.

noahdietz:
Encodes the keys & values of `x-goog-request-params` pairs separately.

Fixes #144

### DIFF
--- a/internal/gengapic/gengapic.go
+++ b/internal/gengapic/gengapic.go
@@ -449,12 +449,12 @@ func (g *generator) insertMetadata(m *descriptor.MethodDescriptorProto) error {
 			field := h[1]
 
 			formats.WriteString("%s=%v&")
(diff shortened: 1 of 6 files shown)


noahdietz:
> do the key escaping at generation time

Yeah I like this better. Cluttering the generated code for something that will likely never happen seems wrong.

PR #146, opened 2019-07-15
gapic: introduce gRPC ServiceConfig as source of retry

noahdietz:
* adds `grpc_service_config` plugin option
* consumes gRPC ServiceConfig for retry configuration
* add copy of `service_config.proto` and it's generated type


noahdietz:
@jadekler can you take a look at the generated product? The `.want` files in this PR will convey what it will look like.

internal/gengapic/client_init.go
```diff
@@ -73,80 +74,63 @@ func (g *generator) clientOptions(serv *descriptor.ServiceDescriptorProto, servN
...
-				return errors.E(err, "cannot read HTTP annotation")
-			}
-			// Generator spec mandates we bucket methods into idempotent and non-idempotent for retries, unless there is an override.
-			// TODO(pongad): implement the override.
-			if _, ok := eHttp.(*annotations.HttpRule).Pattern.(*annotations.HttpRule_Get); ok {
-				idempotent = append(idempotent, m.GetName())
+		policies := map[string]*conf.MethodConfig_RetryPolicy{}
+
+		// gather retry policies from gRPC ServiceConfig
+		if g.grpcConf != nil {
```
pongad: I think the block below is a little uncomfortably far to the right. Maybe something like this might help?
```
var methCfgs []MethodConfig
if g.grpcConf != nil {
  methCfgs = g.grpcConf.GetMethodConfig()
}
for ...
```

internal/gengapic/client_init.go
```diff
@@ -20,14 +20,15 @@ import (
...
+	"github.com/golang/protobuf/ptypes/duration"
+	conf "github.com/googleapis/gapic-generator-go/internal/grpc_service_config"
 	"github.com/googleapis/gapic-generator-go/internal/pbinfo"
 	"google.golang.org/genproto/googleapis/api/annotations"
-	"google.golang.org/genproto/googleapis/rpc/code"
 )
 
 func (g *generator) clientOptions(serv *descriptor.ServiceDescriptorProto, servName string) error {
 	p := g.printf
+	sFQN := fmt.Sprintf("%s.%s", g.descInfo.ParentFile[serv].GetPackage(), serv.GetName())
```
pongad: Do we need `sFQN` all the way here? If not, I think it might be better to move this closer to defaultCallOptions block?

internal/gengapic/client_init.go
noahdietz: Good call. Done!

internal/gengapic/client_init.go
noahdietz: Good call. Done!

jadekler:
> @jadekler can you take a look at the generated product? The `.want` files in this PR will convey what it will look like.

.want files look normal to me. I assume most of the change is in interpreting a new type of file, but mostly generating the same stuff?

noahdietz:
> but mostly generating the same stuff?

Almost. The GAPIC retry config actually names the retry config e.g. "idempotent" "non_idempotent". The generated rery config would be generated once with that name and then the named options assigned to each method. 

The gRPC ServiceConfig doesnt have a "naming" idea for the configuration. Thus, I just assign a retry config directly to the method it corresponds to, resulting in potentially duplicated config objects. This is more verbose.

 I could generate "retry_params_N" "no_retry_params" to do essentially the same thing as before. But I feel like the naming doesn't really help except for consolidation of code...

jadekler:
> > but mostly generating the same stuff?
> 
> Almost. The GAPIC retry config actually names the retry config e.g. "idempotent" "non_idempotent". The generated rery config would be generated once with that name and then the named options assigned to each method.
> 
> The gRPC ServiceConfig doesnt have a "naming" idea for the configuration. Thus, I just assign a retry config directly to the method it corresponds to, resulting in potentially duplicated config objects. This is more verbose.
> 
> I could generate "retry_params_N" "no_retry_params" to do essentially the same thing as before. But I feel like the naming doesn't really help except for consolidation of code...

Yes, that SGTM. Naming is an extra layer of confusion that is good of you to remove.

PR #152, opened 2019-07-18
SampleGen: generate comments in output handling

hzyi-google:


hzyi-google:
@googleapis/samplegen

cmd/gen-go-sample/out.go
```diff
@@ -252,6 +257,28 @@ func writeMap(l *LoopSpec, st *symTab, gen *generator) error {
 	return nil
 }
 
+func writeComment(cmtFmt string, cmtArgs []string, gen *generator) error {
+	var buf bytes.Buffer
+	args := make([]interface{}, len(cmtArgs))
+	for i := range cmtArgs {
```
noahdietz: Curious, why must we essentially cast this slice to that of `interface{}` type? It ultimately gets written via `fmt`, why can't we just use `cmtArgs...` there?

hzyi-google: Go doesn't let me do that... I got a `cannot use strs (type []string) as type []interface {} in argument to fmt.Fprintf`

noahdietz: Yep, you are definitely correct! My bad.

I still think this is kind of wonky to do just to use the expansion operator later. I think most folks' thoughts coming across this code right now would be "Why is the slice of strings being copied into a slice of interfaces?". Putting the buffer print in the loop rather than copying there feels "more simple" to me. 

WDYT?

noahdietz: This was discussed offline and we want to use the `cmtFmt` as it applies to all of the arguments given at once, rather than individually.

PR #153, opened 2019-07-19
SampleGen: allow using `{}` to specify an empty object in request

hzyi-google:
Similar to https://github.com/googleapis/gapic-generator/pull/2844 in gapic-generator.

### DIFF
--- a/cmd/gen-go-sample/inittree.go
+++ b/cmd/gen-go-sample/inittree.go
@@ -223,6 +223,11 @@ func (t *initTree) parseInit(txt string, info pbinfo.Info) error {
 		}
 		t.leafVal = tok
 
+	case '{':
+		if r := sc.Scan(); r != '}' {
+			return report(errors.E(nil, "bad format rhs: expecting '}', found %q", r))
+		}
+
 	default:
 		return report(errors.E(nil, "expected value, found %q", sc.TokenText()))
 	}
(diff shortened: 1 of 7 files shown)


hzyi-google:
cc/ @googleapis/samplegen

cmd/gen-go-sample/inittree.go
```diff
@@ -223,6 +223,11 @@ func (t *initTree) parseInit(txt string, info pbinfo.Info) error {
 		}
 		t.leafVal = tok
 
+	case '{':
+		if r := sc.Scan(); r != '}' {
+			return report(errors.E(nil, "bad format rhs: expecting '}', found %q", r))
```
noahdietz: Does `rhs` mean `right-hand side`? Please don't use an acronym here.

cmd/gen-go-sample/testdata/sample_empty_object.want
```diff
@@ -0,0 +1,59 @@
+// Copyright 2018 Google LLC
```
vchudnov-g: 2019

hzyi-google: We set that to a frozen year in the tests to avoid flakiness. We can do a global replacement to `2019` if you think that's appropriate?

vchudnov-g: Is the license part of the generated code? If so, how is it provided: hard coded or user-specified?

I know in the fullness of time we'll want users to provide the license that we include verbatim in the generated code. It's fine if we're not there yet and instead for now have the generators always whip up a Google-specific license, but in that case the license should include the current year.

Once those considerations are met (with tests), I'm fine with having tests like this one have a frozen year, even in the past: I agree it's less maintenance.

hzyi-google: The license part is generated. The `year` is a parameter passed to the generator's `commit` method. For production we pass in `time.Now().Year()` ([ref](https://github.com/googleapis/gapic-generator-go/blob/b489dd072a026f822353630f252a2433b9c771fb/cmd/gen-go-sample/main.go#L118)), and for testing we pass in a hard-coded `2018` ([ref](https://github.com/googleapis/gapic-generator-go/blob/b489dd072a026f822353630f252a2433b9c771fb/cmd/gen-go-sample/main_test.go#L477))

PR #155, opened 2019-07-22
gapic: avoid generating duplicate iterators

noahdietz:
* adds a `generated` field to the `iterType` 
* makes the auxiliary types a generator field for tracking across services
* only generate an iterator once per **package**

Attempting to generate the Ads v2 GAPICs results in a iterator type naming collision.

### DIFF
--- a/internal/gengapic/paging.go
+++ b/internal/gengapic/paging.go
@@ -72,6 +72,7 @@ func (g *generator) iterTypeOf(elemField *descriptor.FieldDescriptorProto) (*ite
 	if iter, ok := g.aux.iters[pt.iterTypeName]; ok {
 		return iter, nil
 	}
+	g.aux.iters[pt.iterTypeName] = &pt
 
 	return &pt, nil
 }
--- a/internal/gengapic/paging.go
+++ b/internal/gengapic/paging.go
@@ -30,11 +30,13 @@ type iterType struct {
 	// If the elem type is a message, elemImports contains pbinfo.ImportSpec for the type.
 	// Otherwise, len(elemImports)==0.
 	elemImports []pbinfo.ImportSpec
+
+	generated bool
 }
 
 // iterTypeOf deduces iterType from a field to be iterated over.
 // elemField should be the "resource" of a paginating RPC.
-func (g *generator) iterTypeOf(elemField *descriptor.FieldDescriptorProto) (iterType, error) {
+func (g *generator) iterTypeOf(elemField *descriptor.FieldDescriptorProto) (*iterType, error) {
 	var pt iterType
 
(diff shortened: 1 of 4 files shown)


noahdietz:
Worth noting: this *might* still be a problem for LROs if two services generated in the same package share a method name. This is because the auxiliary `*Operation` type name is based on just the method name. However this is much less likely to happen (hasn't happened yet).

internal/gengapic/paging.go
```diff
@@ -66,7 +68,12 @@ func (g *generator) iterTypeOf(elemField *descriptor.FieldDescriptorProto) (iter
 		pt.elemTypeName = pType
 		pt.iterTypeName = upperFirst(pt.elemTypeName) + "Iterator"
 	}
-	return pt, nil
+
+	if iter, ok := g.aux.iters[pt.iterTypeName]; ok {
```
pongad: Looks like you're using `g.aux.iters` as a "cache" right? Two comments:
- Consider moving the `g.aux.iters[name] = &pt` logic here, so that the dedupe behavior stays in one place.
- Maybe use `*descriptor.FieldDescriptorProto` as the key of the map? This lets you move the dedupe logic up at the top (I think this helps with clarity)

noahdietz: > Consider moving the g.aux.iters[name] = &pt logic here

er I'm not following what you mean. Cache accesses are only done here and in [gen](https://github.com/googleapis/gapic-generator-go/blob/8a10cc2d4f406efd0063576149d0920261156d7e/internal/gengapic/gengapic.go#L322) (where we know if we are actively generating a new type or not).

> Maybe use *descriptor.FieldDescriptorProto as the key of the map

I'm not sure how this would work. The problem is that the generated type has a naming collision e.g. two services in the same pkg have a paged method for `LogEntry`, which gets generated in each service client file as a `LogEntryIterator`. Using the name as the key in the cache is "good" because that is the conflicting part of the generated output. If we used the descriptor pointer, those two conflicting methods belong to diff descriptors and we don't catch the duplicate. Or am I missing something?

pongad: About `*descriptor.FieldDescriptorProto`, yeah you're right, that wouldn't work at all hahahahaha.

For the first one, to be more concrete, I think we should do all of this:
```
if iter, ok := g.aux.iters[pt.iterTypeName]; ok {
  return iter
}
g.aux.iters[pt.iterTypeName] = &pt
return &pt
```
in this function so that we don't look up the cache in one place and update the cache in another. Does that make sense?

noahdietz: > Does that make sense?

Totally! I missed the spot where it was set lol

Done!

PR #157, opened 2019-07-25
SampleGen: allow enums to be passed in as CLI flags

hzyi-google:
Allow samples to be invoked through `go sample.go --enum_value=FOO`


hzyi-google:
cc/ @googleapis/samplegen

cmd/gen-go-sample/enum.go
```diff
@@ -21,37 +21,55 @@ import (
...
-			return "", err
-		}
-		g.imports[impSpec] = true
+// enumType returns the Go Type name of a protobuf enum type.
+func enumType(info pbinfo.Info, enum *descriptor.EnumDescriptorProto, g *generator) (string, error) {
+	parts, impSpec, err := ancestors(info, enum, g)
+	if err != nil {
+		return "", err
+	}
+	return impSpec.Name + "." + strings.Join(parts, "_"), nil
```
noahdietz: Same comment regarding use of `fmt.Sprintf` or `strings.Builder`

cmd/gen-go-sample/enum.go
```diff
@@ -21,37 +21,55 @@ import (
...
+	var parts []string
+
+	for ; element != nil; element = info.ParentElement[element] {
+		parts = append(parts, element.GetName())
+		topElement = element
+	}
+
+	// We made array in [child, parent, grandparent] order, we want grandparent fist.
+	for i := 0; i < len(parts)/2; i++ {
+		parts[i], parts[len(parts)-1-i] = parts[len(parts)-1-i], parts[i]
```
noahdietz: can you put `len(parts)-1-i` in a variable since it is done twice (like in [SliceTricks](https://github.com/golang/go/wiki/SliceTricks#reversing) :wink:)

cmd/gen-go-sample/main.go
```diff
@@ -332,7 +333,15 @@ func (g *generator) genSample(ifaceName string, methConf GAPICMethod, regTag str
...
 		for i, n := range argNames {
-			fmt.Fprintf(&sb, ", %s %s", n, pbinfo.GoTypeForPrim[argTrees[i].typ.prim])
+			if e, ok := argTrees[i].typ.desc.(*descriptor.EnumDescriptorProto); ok {
+				en, err := enumType(g.descInfo, e, g)
+				if err != nil {
+					return err
+				}
+				fmt.Fprintf(&sb, ", %s %s", n, en)
+			} else {
+				fmt.Fprintf(&sb, ", %s %s", n, pbinfo.GoTypeForPrim[argTrees[i].typ.prim])
```
noahdietz: Can we make this the default?
```go
en := pbinfo.GoTypeForPrim[argTrees[i].typ.prim]
if e, ok := argTrees[i].typ.desc.(*descriptor.EnumDescriptorProto); ok {
  en, err = enumType(g.descInfo, e, g)
  if err != nil {
    return err
  }
}
fmt.Fprintf(&sb, ", %s %s", n, en)
```

Might have to declare an `err` var but this code structure flows better.

noahdietz: Also, separate comment, I'm not super familiar with the `argTrees` type and the guarantees around `.typ.prim` but what if there is typo and the value doesn't exist? Will it print the `nil`? Or should we be handling that here (or maybe in arg parsing?)

cmd/gen-go-sample/out.go
```diff
@@ -75,6 +75,21 @@ func (st *symTab) put(ident string, typ initType) error {
 	return nil
 }
 
+func (st *symTab) disambiguate(ident string, typ initType) string {
```
noahdietz: can you add a comment to this function or in the code? I'm having trouble understanding what this code does (might be my lack of context).

cmd/gen-go-sample/out.go
```diff
@@ -270,8 +285,8 @@ func writeComment(cmtFmt string, cmtArgs []string, gen *generator) error {
 	buf.WriteString("\n")
 	prependLines(&buf, "// ", false)
 	cmts := strings.Split(buf.String(), "\n")
-	for i, c := range(cmts) {
-		if i == len(cmts) - 1 {
+	for i, c := range cmts {
+		if i == len(cmts)-1 {
```
noahdietz: So this skips the last `cmt` item? maybe do `range cmts[:len(cmts)-1]` instead, or 
```go
end := len(cmts)-1
for _, c := range cmts[:end] {
  gen.pt.Printf(c)
}
```
Or some variation of that. Then you don't need `i` and it's a little more clear.

cmd/gen-go-sample/sampmain.go
```diff
@@ -0,0 +1,96 @@
...
+		if err := st.put(arg, dummyType); err != nil {
+			return err
+		}
+	}
+
+	p("func main() {")
+
+	for i := range argNames {
+		if _, ok := argTrees[i].typ.desc.(*descriptor.EnumDescriptorProto); ok {
+			p(`%s := flag.String(%s, "%s", "")`, argNames[i], flagNames[i], argTrees[i].leafVal)
```
noahdietz: use `%q` instead of `"%s"`

noahdietz: Does the first `%s` argument (`String(%s`) need to be wrapped in quotes? Or is it variable name? Just want to make sure because the following `p()` statement uses `%q` in the same position.

cmd/gen-go-sample/main_test.go
```diff
@@ -289,6 +289,35 @@ func TestWriteFile(t *testing.T) {
 
 }
 
+func TestEnum(t *testing.T) {
```
vchudnov-g: I suggest changing `"EType"` and `"AType"` in `initTestGenerator()` to be more descriptive. I assume "E" is for enum. What is "A" for?

cmd/gen-go-sample/main_test.go
```diff
@@ -289,6 +289,35 @@ func TestWriteFile(t *testing.T) {
...
+	t.Parallel()
+
+	g := initTestGenerator()
+	vs := SampleValueSet{
+		ID: "my_value_set",
+		Parameters: SampleParameter{
+			Defaults: []string{
+				`e = FOO`,
```
vchudnov-g: Could you use a unique name other than `FOO`? We already have a bunch of foos floating around, so this will make it easier to read and reason about. Thanks!

cmd/gen-go-sample/out.go
```diff
@@ -75,6 +75,21 @@ func (st *symTab) put(ident string, typ initType) error {
...
+	base := ident
+	sf := 1
+	_, ok := st.scope[ident]
+	for ok {
+		sf++
+		ident = fmt.Sprintf("%s%d", base, sf)
+		_, ok = st.scope[ident]
+	}
+	if err := st.put(ident, typ); err != nil {
+		panic("bad state: ident shouldn't have existed")
```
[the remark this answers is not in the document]
noahdietz: agreed, returning `err` is preferred to panicking (even if the end result is the tool panicking at the top level)

cmd/gen-go-sample/enum.go
```diff
@@ -21,37 +21,55 @@ import (
...
 			// EnumType.
-			parts = parts[1:]
+			parts = parts[:len(parts)-1]
 		}
 
-		// We made array in [child, parent, grandparent] order, we want grandparent fist.
-		for i := 0; i < len(parts)/2; i++ {
-			parts[i], parts[len(parts)-1-i] = parts[len(parts)-1-i], parts[i]
-		}
+		return impSpec.Name + "." + strings.Join(parts, "_") + "_" + s, nil
```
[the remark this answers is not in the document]
hzyi-google: Done

cmd/gen-go-sample/enum.go
hzyi-google: Done

cmd/gen-go-sample/enum.go
hzyi-google: Done

cmd/gen-go-sample/main.go
hzyi-google: Done

hzyi-google: Added an error handling statement at the end. Nice catch!

cmd/gen-go-sample/out.go
hzyi-google: Done.

cmd/gen-go-sample/out.go
hzyi-google: Done.

BTW, is there a more elegant way to skip the last empty string when `strings.Split("a\nb\nc\n", "\n")?

cmd/gen-go-sample/sampmain.go
hzyi-google: Done

hzyi-google: Totally! Fixed.

cmd/gen-go-sample/enum.go
```diff
@@ -21,37 +21,55 @@ import (
 	"github.com/googleapis/gapic-generator-go/internal/pbinfo"
 )
 
+// enumFmt returns a function that transforms a protobuf enum value literal into it's corresponding Go const.
```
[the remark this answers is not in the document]
hzyi-google: Done

cmd/gen-go-sample/enum.go
```diff
@@ -21,37 +21,55 @@ import (
...
+func enumType(info pbinfo.Info, enum *descriptor.EnumDescriptorProto, g *generator) (string, error) {
+	parts, impSpec, err := ancestors(info, enum, g)
+	if err != nil {
+		return "", err
+	}
+	return impSpec.Name + "." + strings.Join(parts, "_"), nil
+}
 
-		return impSpec.Name + "." + strings.Join(parts, "_") + "_" + s, nil
+// ancestors returns the names of all the types down to `enum`, included and the ImportSpec.
```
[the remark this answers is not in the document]
hzyi-google: Updated the comments. How about now?

cmd/gen-go-sample/enum.go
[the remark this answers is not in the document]
hzyi-google: Changed to `value`.

cmd/gen-go-sample/out.go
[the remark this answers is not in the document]
hzyi-google: Sooooo, I was hesitating on this, because this error should never happen: it keeps looking up the name in the table (the `for` loop), and when the name does not exist, it puts the name back. I didn't like to return an `err`, because there shouldn't be an error from the user's perspective.

Is it OK to swallow the error here?

cmd/gen-go-sample/main_test.go
hzyi-google: I think `AType` is just "a type". It is a very ordinary message.

cmd/gen-go-sample/main_test.go
hzyi-google: Change `EType` to `FruitEnum`, and the value to `BANANA`.

hzyi-google:
PTAL @noahdietz @vchudnov-g

cmd/gen-go-sample/out.go
pongad: I'm not sure if this is objectively better but I like it a little better: https://play.golang.org/p/MtudlnG2OE8

noahdietz: @pongad 's suggestion is certainly more readable. It's very clear that the code drops the last newline and goes about it's business, rather than dropping the last element  as I had suggested.

cmd/gen-go-sample/out.go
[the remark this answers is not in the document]
noahdietz: I'm OK with swallowing the error as long as we comment as to why it's being swallowed.

cmd/gen-go-sample/main.go
```diff
@@ -332,16 +332,21 @@ func (g *generator) genSample(ifaceName string, methConf GAPICMethod, regTag str
 	var argStr string
 	if len(argNames) > 0 {
 		var sb strings.Builder
-		for i, n := range argNames {
-			if e, ok := argTrees[i].typ.desc.(*descriptor.EnumDescriptorProto); ok {
-				en, err := enumType(g.descInfo, e, g)
+		for i, name := range argNames {
+			var typ string
```
noahdietz: We don't need to declare `typ` like this if it's being redeclared immediately via `typ, ok :=`.

hzyi-google: Ah my bad!

cmd/gen-go-sample/out.go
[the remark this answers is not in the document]
vchudnov-g: OK, sounds reasonable. +1 to comment

cmd/gen-go-sample/sampmain.go
```diff
@@ -0,0 +1,96 @@
...
+			if err != nil {
+				return err
+			}
+			vMap := tn + "_name"
+			st.put(v, dummyType)
+			fArgs = append(fArgs, fmt.Sprintf("%s(%s)", tn, v))
+
+			p(`%s, ok := %s[*%s]`, v, vMap, argNames[i])
+			p("if !ok {")
+			p(`log.fatal("enum type %s does not have value %%s", *%s)`, tn, v)
```
vchudnov-g: s/`v`/`argNames[i]`/ ? See  sample_enum.want

cmd/gen-go-sample/testdata/sample_enum.want
```diff
@@ -0,0 +1,62 @@
...
+}
+
+// [END awesome_region]
+
+func main() {
+	myFruit := flag.String("my_fruit", "APPLE", "")
+	flag.Parse()
+	myFruit2, ok := foopb.AType_FruitEnum_name[*myFruit]
+	if !ok {
+		log.fatal("enum type foopb.AType_FruitEnum does not have value %s", *myFruit2)
```
vchudnov-g: It seems to me that if `!ok!`, then the lookup of `*myFruit` failed, right?, so _that's_ what we should be reporting here (rather than `*myFruit2`---is that even defined in case of error?).

cmd/gen-go-sample/testdata/sample_enum.want
```diff
@@ -0,0 +1,62 @@
...
+	fmt.Printf("call finished.\n")
+	return nil
+}
+
+// [END awesome_region]
+
+func main() {
+	myFruit := flag.String("my_fruit", "APPLE", "")
+	flag.Parse()
+	myFruit2, ok := foopb.AType_FruitEnum_name[*myFruit]
```
vchudnov-g: IIUC, the name `myFruit2` is autogenerated from `myFruit`? Having the suffix 2 does not read very friendly to me.
I would suggest either:
 
(a) using `myFruitArg`
```
myFruitArg := flag.String("my_fruit", ...)
...
myFruit, ok := foopb.AType_FruitEnum_name[*myFruitArg]
```
OR 

(b) using `myFruitEnum` (or similar)
```
myFruit := flag.String("my_fruit", ...)
...
myFruitEnum, ok := foopb.AType_FruitEnum_name[*myFruit]
```

WDYT? I personally prefer (a). It's fine to do this in a follow-up PR, but I think we should.

cmd/gen-go-sample/main_test.go
vchudnov-g: This helps a lot. Thank you!

cmd/gen-go-sample/testdata/sample_enum.want
hzyi-google: Sounds good. I'll make `myFruitArg` the base name, and if that's already taken we use a numeric suffix to disambiguate.

cmd/gen-go-sample/out.go
```diff
@@ -75,6 +75,27 @@ func (st *symTab) put(ident string, typ initType) error {
...
+func (st *symTab) disambiguate(ident string, typ initType) string {
+	base := ident
+	sf := 1
+	_, ok := st.scope[ident]
+	for ok {
+		sf++
+		ident = fmt.Sprintf("%s%d", base, sf)
+		_, ok = st.scope[ident]
+	}
+	// We just checked that `ident` is not in the table, thus swallowing the error
```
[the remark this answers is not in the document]
hzyi-google: Done

cmd/gen-go-sample/sampmain.go
hzyi-google: Done

cmd/gen-go-sample/testdata/sample_enum.want
hzyi-google: Done

cmd/gen-go-sample/testdata/sample_enum.want
hzyi-google: Done

hzyi-google:
PTAL @vchudnov-g

PR #165, opened 2019-07-31
gapic: support max req & res size limits

noahdietz:
Addresses https://github.com/googleapis/gapic-generator/issues/2895 and makes things configurable via gRPC ServiceConfig.

FYI @jadekler


internal/gengapic/client_init.go
```diff
@@ -61,6 +61,9 @@ func (g *generator) clientOptions(serv *descriptor.ServiceDescriptorProto, servN
 		p("    option.WithEndpoint(%q),", host)
 		p("    option.WithGRPCDialOption(grpc.WithDisableServiceConfig()),")
 		p("    option.WithScopes(DefaultAuthScopes()...),")
+		p("		 option.WithGRPCDialOption(grpc.WithDefaultCallOptions(")
```
pongad: Too many leading white spaces here?

internal/gengapic/testdata/empty_opt.want
```diff
@@ -10,12 +10,17 @@ func defaultClientOptions() []option.ClientOption {
 		option.WithEndpoint("foo.bar.com:443"),
 		option.WithGRPCDialOption(grpc.WithDisableServiceConfig()),
 		option.WithScopes(DefaultAuthScopes()...),
+		option.WithGRPCDialOption(grpc.WithDefaultCallOptions(
+		grpc.MaxCallRecvMsgSize(-1),
```
pongad: I don't see the documentation saying that -1 refers to infinity. Did I miss something? If not, maybe we should send a small documentation PR to the gRPC repo?

internal/gengapic/client_init.go
noahdietz: There was a tab character there. Oops! Done.

internal/gengapic/testdata/empty_opt.want
noahdietz: I've sent y'all an email exchange with one of the gRPC TLs where he says this is supported. Furthermore, this is done already in some of the Python GAPICs

noahdietz: After looking at the source code, I might be second guessing this behavior for grpc-go specifically. The defaultClientMaxSendMsgSize is `math.maxInt32` ([here](https://github.com/grpc/grpc-go/blob/977142214c45640483838b8672a43c46f89f90cb/clientconn.go#L96)), rather than -1. This also precludes the need for us to set the maxSendMsgSize call option, its already maxed out by default.

noahdietz: Just double checked with a grpc-go team member, and they said they don't support -1. I will use the math lib max int value instead. Thanks @pongad for catching this!

noahdietz: Done! PTAL

PR #169, opened 2019-08-07
SampleGen: Preparation work for sample config migration

hzyi-google:
- Split up `genSample` method into smaller methods to make transitioning smoother
- Add sample config schema


cmd/gen-go-sample/schema_v1p2/schema.go
```diff
@@ -0,0 +1,63 @@
...
+
+type SampleConfig struct {
+	Type    string
+	Version string `yaml: config_schema_version`
+	Samples []Sample
+}
+
+type Sample struct {
+	ID              string `yaml: id`
+	Title           string
```
software-dov: Not an issue, just curious: where does the title come from? I don't see anything about a title in the samplegen user guide.

cmd/gen-go-sample/schema_v1p2/schema.go
```diff
@@ -0,0 +1,63 @@
...
+type LoopSpec struct {
+	Collection string
+	Map        string
+	Variable   string
+	Key        string
+	Value      string
+	Body       []ResponseConfig
+}
+
+type WriteFileSpec struct {
```
software-dov: It looks like this type is duplicated in `gapic.go`. Is that deliberate?

cmd/gen-go-sample/schema_v1p2/schema.go
hzyi-google: Hmmmm I think we found that we needed this after sharing out the user guide. I'll add it there. (It's used the generate this [`sample-metadata`](https://github.com/googleapis/gapic-generator/blob/a591ce660d04b0e9c6876b9836d8aac0e9f65b6d/src/test/java/com/google/api/codegen/gapic/testdata/java/java_library.baseline#L3302-L3305) part, but I don't think anyone needs the `sample-metadata` yet)

cmd/gen-go-sample/schema_v1p2/schema.go
hzyi-google: `gapic.go` consists of gapic configs needed for samplegen. Since we are moving sample config outside gapic config, I decided to move the struct to `schema_v1p2` too. A coming PR will have samplegen depend on the structs in the `schema_v1p2` package, and delete them here.

cmd/gen-go-sample/main.go
```diff
@@ -224,6 +225,77 @@ func (g *generator) genSample(ifaceName string, methConf GAPICMethod, regTag str
...
+	files := initInfo.files
+	itree := initInfo.reqTree
+
+	if err := g.sampFuncAndClient(serv.GetName(), meth.GetName(), initInfo, regTag); err != nil {
+		return err
+	}
+
+	// Set up the request
+	{
+		var buf bytes.Buffer
```
noahdietz: Why are we using a new buffer instead of just using the `g.pt.Writer()`? The buffer gets written to the `g.pt` later anyways.

cmd/gen-go-sample/typing.go
```diff
@@ -0,0 +1,41 @@
...
+package main
+
+import (
+	"github.com/golang/protobuf/protoc-gen-go/descriptor"
+	"github.com/googleapis/gapic-generator-go/internal/errors"
+	"github.com/googleapis/gapic-generator-go/internal/pbinfo"
+)
+
+func (g *generator) getGoTypeName(typ initType) (string, error) {
+	if typ.prim != 0 {
```
noahdietz: Please add a comment to this check against a magic number. I know that the type enum starts at 1 and 0 is for errors, but many might not know that.

cmd/gen-go-sample/typing.go
```diff
@@ -0,0 +1,41 @@
...
+		goType, ok := pbinfo.GoTypeForPrim[typ.prim]
+		if !ok {
+			return "", errors.E(nil, "unrecognized primitive type: %s", typ.prim)
+		}
+		return goType, nil
+	}
+
+	if enum, ok2 := typ.desc.(*descriptor.EnumDescriptorProto); ok2 {
```
noahdietz: you can (and should) just name this `ok`. The previous `ok` isn't in scope, so `ok2` is kind of confusing

cmd/gen-go-sample/schema_v1p2/schema.go
vchudnov-g: And fyi, there has been some discussion of moving the metadata to the manifest, but nothing's been decided in that regard.

cmd/gen-go-sample/main.go
pongad: It's the prependLines on L261 I think. It modifies the buffer afterwards.

hzyi-google: Yes! We need to insert those "//" to some of the lines here.

cmd/gen-go-sample/inittree.go
```diff
@@ -69,6 +69,21 @@ type initTree struct {
 	leafVal string
 }
 
+// request represents all the information needed to construct the request object.
```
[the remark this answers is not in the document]

cmd/gen-go-sample/typing.go
hzyi-google: Done.

cmd/gen-go-sample/typing.go
hzyi-google: Done.

hzyi-google:
PTAL

cmd/gen-go-sample/main.go
```diff
@@ -224,6 +225,77 @@ func (g *generator) genSample(ifaceName string, methConf GAPICMethod, regTag str
...
 	g.imports[inSpec] = true
 
+	initInfo, err := g.getInitInfo(inType, methConf, valSet)
+	argNames := initInfo.argNames
+	argTrees := initInfo.argTrees
+	flagNames := initInfo.flagNames
+	files := initInfo.files
+	itree := initInfo.reqTree
+
+	if err := g.sampFuncAndClient(serv.GetName(), meth.GetName(), initInfo, regTag); err != nil {
```
vchudnov-g: A suggestion: How about refactoring this function slightly?:
```
args, err := getFunctionArgList(...)
p.(...) // function signature, region tag, and client
 ...
err := setUpRequest(...)
err := processResponse(...)
p.(...) // end function, end region tag, etc.
  ...
return writeMain(...)
```
(not showing the error handling above, obvs)

This makes it easier to grok the high-level structure, while delegating groups of details to functions, and we get the symmetry of opening and closing the sample function and region tags in the same function. WDYT?

cmd/gen-go-sample/main.go
```diff
@@ -316,101 +390,52 @@ func (g *generator) genSample(ifaceName string, methConf GAPICMethod, regTag str
...
+		files:     files,
+		reqTree:   itree,
+	}
 
-	p("// [START %s]", regTag)
-	p("")
+	return initInfo, nil
+}
 
+func (g *generator) sampFuncAndClient(servName string, methName string, init initInfo, regTag string) (error) {
```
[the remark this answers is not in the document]
hzyi-google: The name seems to long and a bit anti-pattern in Go.
I'll split this function up into two and have each of them handle writing sample function signature and client initialization separately in a coming PR.

cmd/gen-go-sample/main.go
hzyi-google: SG.

hzyi-google:
@vchudnov-g I'll fix them in a coming PR.

cmd/gen-go-sample/main.go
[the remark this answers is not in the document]
vchudnov-g: You've heard my rant  by now, which I can summarize by saying "identifier names should be as short as meaningfully possible, but no shorter". I like having short names, but I stand defiant in saying that's second to making the names informative ;-)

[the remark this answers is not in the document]
hzyi-google: Yes I agree with informative names (man, I love Java), but the fact that this function needs a long name makes me think I have put too much logic into this function, and thus I think a better way is to split it up :)

PR #174, opened 2019-08-12
SampleGen: Switch to sample config in tests

hzyi-google:
Done:
- Generate samples from sample configs in unit tests
- Remove samplegen related fields from `gapic.go`
- Style issues in previous PR

Excluded (next PR):
- The end-to-end generation pipeline including:
  - Update CLI to take a sample config path
  - Read sample configs from yaml files and pass them to the generation pipeline
  - Sample ID collision resolution


cmd/gen-go-sample/testdata/sample_unary.want
```diff
@@ -72,6 +71,7 @@ func sampleUnaryMethod(theX int64, theB string, theFoo string, bobFile string) e
 				Y: 3,
 			},
 		},
+		DataBob: bobFileBytes,
```
hzyi-google: We used to have `defaults` and `attributes` in the request configuration items in these two directives do not have to be in the same order. But now default value and sample argument name are configured in the sample directive (`request.value` and `request.input_parameter`), thus causing poor `DataBob` to be moved to the end.

hzyi-google:
cc/ @googleapis/samplegen

cmd/gen-go-sample/inittree.go
```diff
@@ -185,71 +185,57 @@ func (t *initTree) index(k string) *initTree {
...
+		return report(errors.E(nil, "value already set to %q", lv))
 	}
 
-	// TODO(pongad): handle resource names
-	switch r := sc.Scan(); r {
-	case scanner.Int, scanner.Float, scanner.String, scanner.Ident:
-		if lv := t.leafVal; lv != "" {
-			return report(errors.E(nil, "value already set to %q", lv))
+	if _, ok := t.typ.desc.(*descriptor.DescriptorProto); ok {
+		if value != "{}" {
```
noahdietz: Personally, I don't understand why we want the value to be `"{}"`, so this code confuses me (my own lack of understanding of samplegen, slowly learning more though). A comment or putting this value in a `const` (possibly with a comment at its declaration too) would be appreciated.

Furthermore, if the value is in a `const`, it can be printed in the error message with the same quotations (using `%q`) as the "found" value. It's a super nit, but rn it would be printed as `expecting '{}', found "foo"` (or something of the like).

cmd/gen-go-sample/inittree.go
```diff
@@ -185,71 +185,57 @@ func (t *initTree) index(k string) *initTree {
...
+	} else {
+		pType := t.typ.prim
+		validPrim := validPrims[pType]
+		if validPrim == nil {
+			return report(errors.E(nil, "not a primitive type? %q", pType))
+		}
+		if !validPrim(value) {
+			return report(errors.E(nil, "invalid value for type %q: %q", pType, value))
+		}
+		if t.typ.prim == descriptor.FieldDescriptorProto_TYPE_BYTES {
```
noahdietz: `if pType ==`, we put `t.typ.prim` in a local variable already

cmd/gen-go-sample/inittree.go
```diff
@@ -185,71 +185,57 @@ func (t *initTree) index(k string) *initTree {
...
+			value = fmt.Sprintf("%q", value)
+			t.typ.valFmt = bytesFmt()
+		}
+		// We need to quote string literals
+		if t.typ.prim == descriptor.FieldDescriptorProto_TYPE_STRING {
```
noahdietz: Same thing, `if pType ==`

cmd/gen-go-sample/main.go
```diff
@@ -81,60 +82,58 @@ func main() {
...
-				return errors.E(err, "value set: %s", vsID)
-			}
-			content, err := gen.commit(!nofmt, time.Now().Year())
-			if err != nil {
-				return err
+func genMethodSamples(gen *generator, sampConf schema_v1p2.SampleConfig, nofmt bool, outDir string) error {
+	for _, samp := range sampConf.Samples {
+
+		for _, iface := range gen.gapic.Interfaces {
+			if iface.Name == samp.Service {
```
noahdietz: I think it would be better to invert this to keep the body of this `if` statement less indented.

```go
if iface.Name != samp.Service {
  continue
}
```

cmd/gen-go-sample/main.go
```diff
@@ -81,60 +82,58 @@ func main() {
...
+				for _, meth := range iface.Methods {
+					if meth.Name == samp.Rpc {
```
noahdietz: Same comment here WRT inverting the conditional and indentation.

cmd/gen-go-sample/main.go
```diff
@@ -81,60 +82,58 @@ func main() {
...
-			gen.reset()
-			if err := gen.genSample(iface.Name, meth, sam.RegionTag, vs); err != nil {
-				return errors.E(err, "value set: %s", vsID)
-			}
-			content, err := gen.commit(!nofmt, time.Now().Year())
-			if err != nil {
-				return err
+func genMethodSamples(gen *generator, sampConf schema_v1p2.SampleConfig, nofmt bool, outDir string) error {
+	for _, samp := range sampConf.Samples {
+
```
noahdietz: please delete this line break

cmd/gen-go-sample/inittree.go
```diff
@@ -185,71 +185,57 @@ func (t *initTree) index(k string) *initTree {
...
+	if lv := t.leafVal; lv != "" {
+		return report(errors.E(nil, "value already set to %q", lv))
 	}
 
-	// TODO(pongad): handle resource names
-	switch r := sc.Scan(); r {
-	case scanner.Int, scanner.Float, scanner.String, scanner.Ident:
-		if lv := t.leafVal; lv != "" {
-			return report(errors.E(nil, "value already set to %q", lv))
+	if _, ok := t.typ.desc.(*descriptor.DescriptorProto); ok {
```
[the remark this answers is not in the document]
hzyi-google: Like dis?

cmd/gen-go-sample/inittree.go
hzyi-google: Good call! Added a `const emptyObjectLiteral = "{}"` to the top of the file.

cmd/gen-go-sample/inittree.go
hzyi-google: Done

cmd/gen-go-sample/inittree.go
hzyi-google: Done

cmd/gen-go-sample/main.go
hzyi-google: Done

cmd/gen-go-sample/main.go
hzyi-google: Done

cmd/gen-go-sample/main.go
hzyi-google: Done

cmd/gen-go-sample/inittree.go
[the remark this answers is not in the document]
noahdietz: Almost! Sorry I should've included an example. We can make life just a *tiny* bit easier by declaring a variable in the switch statement, precluding the need to caste the type manually into a local. [Example](https://github.com/googleapis/gapic-generator-go/blob/ca13906cc094887efa78d43d7d3bf8368d38f8f9/internal/gengapic/markdown.go#L52)

```go
switch typ := t.typ.desc.(type) {
...
}
```
Then `typ` will be automatically caste to the type `case` that it matches and you can use `typ` in the `case` statement like that type (pretty cool imo).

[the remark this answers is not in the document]
hzyi-google: Done! (I used `desc` instead of `typ`, because typ` is the parent of `desc` and make the enum `case` block a bit confusing).

PTAL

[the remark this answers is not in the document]
noahdietz: Awesome.

cmd/gen-go-sample/main.go
```diff
@@ -81,60 +82,61 @@ func main() {
...
+				content, err := gen.commit(!nofmt, time.Now().Year())
+				if err != nil {
+					return err
+				}
+				// Handle duplicate sample IDs
+				fname := samp.ID + ".go"
+				if err := ioutil.WriteFile(filepath.Join(outDir, fname), content, 0644); err != nil {
+					return err
+				}
+				break
```
vchudnov-g: Nit: I would add a comment to remind the reader which loop we're breaking, here and below: `break //for meth`

[This pattern of having to remember a `break` at the bottom of the loop makes me slightly uneasy. If only we could do something like `defer break`..... Personally I would just up skip the non-matching elements in the loop and do the processing for the matching element outside the loop, but that's just a matter of taste]

cmd/gen-go-sample/main.go
```diff
@@ -81,60 +82,61 @@ func main() {
...
 			}
+			break
```
vchudnov-g: Where do you check the case of no matching `meth.Name` and the case of no matching `iface.Name`? In either of those cases I think we should error. (I'll propose this in the spec)

cmd/gen-go-sample/main.go
```diff
@@ -225,77 +229,68 @@ func (g *generator) genSample(ifaceName string, methConf GAPICMethod, regTag str
...
 
 	// Make the RPC call and handle output
 	if meth.GetOutputType() == ".google.protobuf.Empty" {
-		err = g.emptyOut(meth, valSet)
+		err = g.emptyOut(meth, sampConf.Response)
 	} else if meth.GetOutputType() == ".google.longrunning.Operation" {
-		err = g.lro(meth, methConf, valSet)
+		err = g.lro(meth, methConf, sampConf.Response)
 	} else if meth.GetServerStreaming() || meth.GetClientStreaming() {
 		err = errors.E(nil, "streaming methods not supported yet")
```
vchudnov-g: This is simply due to us not implementing them yet in samplegen? If so, could you add a TODO here (so this is easy to find) referencing a GitHub issue to do this? Thanks!

cmd/gen-go-sample/main.go
```diff
@@ -338,63 +333,63 @@ func (g *generator) getInitInfo(inType pbinfo.ProtoType, methConf GAPICMethod, v
 	}
 
 	// Set up request object.
-	for _, def := range valSet.Parameters.Defaults {
-		if err := itree.parseInit(def, g.descInfo); err != nil {
-			return initInfo{}, errors.E(err, "can't set default value: %q", def)
+	for _, reqConf := range reqConfs {
```
vchudnov-g: how about `_, fieldConf := range reqConf`? (because each `RequestConfig` describes a field)
I think that makes for clearer local var and param names, and avoids them looking so similar

cmd/gen-go-sample/main.go
```diff
@@ -338,63 +333,63 @@ func (g *generator) getInitInfo(inType pbinfo.ProtoType, methConf GAPICMethod, v
...
 	// Some parts of request object are from arguments.
-	for _, attr := range valSet.Parameters.Attributes {
-		if attr.ReadFile {
+	for _, reqConf := range reqConfs {
+		if reqConf.ValueIsFile {
 			var varName string
 			var err error
-			if attr.SampleArgumentName != "" {
-				varName = snakeToCamel(attr.SampleArgumentName) + fileContentSuffix
+			if reqConf.InputParameter != "" {
```
vchudnov-g: You have this condition several times. I'd assign it its own variable for clarity: 
```
isSampleParam = (fieldConf.InputParameter != "")
```

cmd/gen-go-sample/main.go
```diff
@@ -338,63 +333,63 @@ func (g *generator) getInitInfo(inType pbinfo.ProtoType, methConf GAPICMethod, v
...
 				argNames = append(argNames, fileName)
-				flagNames = append(flagNames, attr.SampleArgumentName)
+				flagNames = append(flagNames, reqConf.InputParameter)
 				argTrees = append(argTrees, subTree)
 			}
 			files = append(files, &fileInfo{fileName, varName})
 			continue
 		}
-		if attr.SampleArgumentName == "" {
+		if reqConf.InputParameter == "" {
```
vchudnov-g: nit: blank line above

hzyi-google:
We could change all `meth` to just `m` (lol, jk, jk)

There are a lot of `meth`s in the project and I kinda just followed the pattern. I'll start to change them to `method` lazily.

cmd/gen-go-sample/inittree.go
```diff
@@ -185,71 +188,58 @@ func (t *initTree) index(k string) *initTree {
...
 	default:
-		return report(errors.E(nil, "expected value, found %q", sc.TokenText()))
-	}
-
-	if sc.Scan() != scanner.EOF {
-		return report(errors.E(nil, "expected EOF, found %q", sc.TokenText()))
+		pType := t.typ.prim
+		validPrim := validPrims[pType]
+		if validPrim == nil {
+			return report(errors.E(nil, "not a primitive type? %q", pType))
```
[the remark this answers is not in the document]
hzyi-google: This is copy-paste from Michael's code. I'll remove it.

cmd/gen-go-sample/main.go
```diff
@@ -81,60 +82,61 @@ func main() {
...
+				}
 
-			fname := iface.Name + "_" + meth.Name + "_" + vsID + ".go"
-			if err := ioutil.WriteFile(filepath.Join(outDir, fname), content, 0644); err != nil {
-				return err
+				content, err := gen.commit(!nofmt, time.Now().Year())
+				if err != nil {
+					return err
+				}
+				// Handle duplicate sample IDs
```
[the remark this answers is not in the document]
hzyi-google: Yes. Added TODO.

software-dov:
Lotta meth in the project. Must have been written in Albuquerque.

cmd/gen-go-sample/main.go
hzyi-google: Good point! Slightly tweaked this part of the code to achieve both this and the comment above.

cmd/gen-go-sample/main.go
hzyi-google: https://github.com/googleapis/gapic-generator-go/issues/177 Will also link in code.

cmd/gen-go-sample/main.go
hzyi-google: Changed to `fieldConf` and `fieldConfs`. (It's somewhat weird to use a singular noun `reqConf` as a name of a slice)

cmd/gen-go-sample/main.go
hzyi-google: Done. Changed the local var to `isInputParam` to be consistent with config schema. But I can change to `isSampleParam` too.

cmd/gen-go-sample/inittree.go
```diff
@@ -185,71 +188,58 @@ func (t *initTree) index(k string) *initTree {
 }
 
 // Just do simple structs for now.
-// TODO(pongad): allow map and array index.
-//
-// spec = path [ '=' value ] .
-func (t *initTree) parseInit(txt string, info pbinfo.Info) error {
+// TODO(hzyi): allow map index.
+func (t *initTree) parseInit(path string, value string, info pbinfo.Info) error {
```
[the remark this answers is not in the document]

hzyi-google:
PTAL

cmd/gen-go-sample/main.go
```diff
@@ -98,34 +98,45 @@ func main() {
...
-				if err != nil {
-					return err
-				}
-				// Handle duplicate sample IDs
-				fname := samp.ID + ".go"
-				if err := ioutil.WriteFile(filepath.Join(outDir, fname), content, 0644); err != nil {
-					return err
+		var iface GAPICInterface
+		var method GAPICMethod
+		for _, i := range gen.gapic.Interfaces {
```
vchudnov-g: I think you can just do
```
for _, iface = range ...
```
(no colon)
and this gets rid of the need for `i` and reassigning to `iface.

cmd/gen-go-sample/main.go
```diff
@@ -98,34 +98,45 @@ func main() {
...
+			if i.Name == samp.Service {
+				iface = i
+				for _, m := range iface.Methods {
```
vchudnov-g: likewise, `for _, method = range ...`

cmd/gen-go-sample/main.go
```diff
@@ -98,34 +98,45 @@ func main() {
...
+						method = m
+						break
+					}
 				}
 				break
 			}
-			break
+		}
+
+		if method.Name == "" {
```
vchudnov-g: Yay for taking everything below out of the loop! :+1:

cmd/gen-go-sample/main.go
```diff
@@ -98,34 +98,45 @@ func main() {
 }
 
 func genMethodSamples(gen *generator, sampConf schema_v1p2.SampleConfig, nofmt bool, outDir string) error {
+
```
vchudnov-g: nit: rm blank line

cmd/gen-go-sample/main.go
```diff
@@ -98,34 +98,45 @@ func main() {
...
+		if method.Name == "" {
+			return errors.E(nil, "generating sample %q: rpc %q not found", samp.ID, method.Name)
+		}
+		if iface.Name == "" {
+			return errors.E(nil, "generating sample %q: service %q not found", samp.ID, iface.Name)
+		}
+
+		gen.reset()
+		if err := gen.genSample(samp, method); err != nil {
+
```
vchudnov-g: nit: rm blank line

cmd/gen-go-sample/main.go
```diff
@@ -98,34 +98,45 @@ func main() {
...
+			err = errors.E(err, "generating: %s:%s", iface.Name+"."+method.Name, samp.ID)
```
vchudnov-g: Would be simpler to just do `%s.%s:%s` in the format string and omit string concat.

cmd/gen-go-sample/main.go
vchudnov-g: Yeah, in general. But I'm thinking as a user of a "request config" being comprised of several "field configs", so in that context IMO it makes sense for the name to be singular. But not a biggie either way. :man_shrugging:

cmd/gen-go-sample/main.go
vchudnov-g: Of course, I just realized my suggestion would require a different way of marking when a method and interface are not found. Maybe setting a simple found flag before the `break`? It's probably just fine to simply have a single error message that just says `{service}.{rpc} not found`....

hzyi-google: Done. We can still compare the method name and interface name with `samp.Rpc` and `samp.Service` outside the loop so we don't necessarily need a flag.

cmd/gen-go-sample/main.go

PR #179, opened 2019-08-21
SampleGen: CLI for sample configs

hzyi-google:
- Add a `--samples` flag which points at a sample config file or a directory containing sample config files
- Disambiguate sample IDs when there are multiple samples with the same ID.
- Fix a bunch of wrong yaml tags

Generated speech samples: https://github.com/hzyi-google/sample-staging/tree/master/20190821/go-output

Also, the output handling part is not handling repeated fields correctly. Bug filed: https://github.com/googleapis/gapic-generator-go/issues/180


cmd/gen-go-sample/primval.go
```diff
@@ -24,7 +24,7 @@ import (
 // Functions only perform cursory checking; since we use scanner for tokenizing,
 // they must already be a valid token of some type.
 var validPrims = [...]func(string) bool{
-	descriptor.FieldDescriptorProto_TYPE_BOOL: func(s string) bool { return s == "true" || s == "false" },
+	descriptor.FieldDescriptorProto_TYPE_BOOL: func(s string) bool { return s == "true" || s == "false" || s == "True" || s == "False" },
```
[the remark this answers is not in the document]
hzyi-google: This was added for debugging. Removing it.

cmd/gen-go-sample/main.go
```diff
@@ -203,6 +267,32 @@ func (g *generator) commit(gofmt bool, year int) ([]byte, error) {
...
 
+func (g *generator) disambiguateSampleIDs() error {
+	idCount := make(map[string]int)
+	samples := g.sampleConfig.Samples
+	for i := range samples {
+		// default ID to region tag
+		if samples[i].ID == "" {
+			samples[i].ID = samples[i].RegionTag
+		}
+		if samples[i].ID != "" {
```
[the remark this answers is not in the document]
hzyi-google: Yes. Both `ID` and `RegionTag` are optional.

hzyi-google:
PTAL
cc/ @googleapis/samplegen

cmd/gen-go-sample/main.go
```diff
@@ -79,32 +89,91 @@ func main() {
...
+		if err != nil {
+			return err
+		}
+
+	case mode.IsRegular():
+		f, err := os.Open(path)
+		if err != nil {
+			return errors.E(err, "cannot read sample config file: %s", path)
+		}
+		if err := yaml.NewDecoder(f).Decode(&gen.sampleConfig); err != nil {
```
software-dov: Should also check type and version.

hzyi-google: That's right! Updated.

hzyi-google:
PTAL

cmd/gen-go-sample/main.go
```diff
@@ -79,32 +93,97 @@ func main() {
...
+					// ignore directories and non-YAML files
+					return nil
+				}
+
+				f, err := os.Open(p)
+				if err != nil {
+					return errors.E(err, "cannot read sample config file: %s", p)
+				}
+
+				if err := yaml.NewDecoder(f).Decode(&sc); err != nil {
```
vchudnov-g: Please support multiple documents in the YAML file as per the spec. Sounds from the godoc like that simply entails calling `Decode` again until EOF. You'll want to check all the docs and append the matching ones before returning `nil`

cmd/gen-go-sample/main.go
```diff
@@ -79,32 +93,97 @@ func main() {
...
+		if err != nil {
+			return err
+		}
+
+	case mode.IsRegular():
+		f, err := os.Open(path)
+		if err != nil {
+			return errors.E(err, "cannot read sample config file: %s", path)
+		}
+		if err := yaml.NewDecoder(f).Decode(&gen.sampleConfig); err != nil {
```
vchudnov-g: Support multiple YAML documents in the file and error if none of them are a sample config.

cmd/gen-go-sample/main.go
```diff
@@ -203,6 +282,35 @@ func (g *generator) commit(gofmt bool, year int) ([]byte, error) {
...
+	}
+
+	for i := range samples {
+		// Generate a sample ID when the user-provided ID is empty not unique
+		if samples[i].ID == "" || idCount[samples[i].ID] > 1 {
+			jsonStr, err := json.Marshal(samples[i])
+			if err != nil {
+				return err
+			}
+			checkSum := sha256.Sum256([]byte(jsonStr))
```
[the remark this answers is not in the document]
hzyi-google: My reasoning was the same: In that case there are two exactly identical sample configuration, thus just generating one sample isn't a bad outcome.

cmd/gen-go-sample/main.go
hzyi-google: Ack

cmd/gen-go-sample/main.go
hzyi-google: Ack

cmd/gen-go-sample/main.go
```diff
@@ -79,32 +93,97 @@ func main() {
...
+				}
+				if sc.Type != expectedSampleConfigType {
+					// ignore non sample config files
+					return nil
+				}
+				if sc.Version != expectedSampleConfigVersion {
+					// ignore unsupported versions
+					return nil
+				}
+				gen.sampleConfig.Samples = append(gen.sampleConfig.Samples, sc.Samples...)
```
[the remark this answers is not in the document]
hzyi-google: Ack.

cmd/gen-go-sample/main.go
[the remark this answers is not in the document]
software-dov: The python microgenerator is susceptible to the same non-uniqueness scenario.
I would argue it's not _wrong_ to do the work twice and overwrite, but it it would definitely be more user friendly to fail loudly.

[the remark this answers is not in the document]
vchudnov-g: "Failing loudly" in this case would entail keeping the hash as is (without appending the `i`) and then checking that each hash is unique. (Agreed?)

I can see pros and cons. I wonder whether a warning rather than an error would be better: we still generate one sample but we warn the user that two separate configs resulted in the same sample. I can imagine that in some cases (eg during sample configuration, which might involve a lot of copying and pasting and testing) that might be acceptable.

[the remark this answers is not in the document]
software-dov: Pretty much; by "fail loudly" I mean raise a terminating error of some kind.
Not to get back on this soap box, but I think a warning is a compromise nobody wants. Silently swallowing duplicates and erroring out on duplicates are both consistent, reasonable choices.
Flagging warnings feels like a slippery slope to being too clever, like when you pointed out that calling form wasn't sufficient to disambiguate duplicated IDs. Instead of trying to auto-generate a descriptive, unique ID and falling back to appending a hash, it was a lot easier and less error prone to just disambiguate using hashes immediately.

cmd/gen-go-sample/main.go

cmd/gen-go-sample/main.go

cmd/gen-go-sample/main.go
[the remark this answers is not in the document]

hzyi-google:
PTAL

cmd/gen-go-sample/main.go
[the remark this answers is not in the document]

hzyi-google:
PTAL

cmd/gen-go-sample/testdata/sample_response_map_field.want
```diff
@@ -0,0 +1,60 @@
+// Copyright 2018 Google LLC
+//
```
vchudnov-g: Is this file meant to be part of this PR? Doesn't seem to deal with CLI for sample configs....

hzyi-google: OH NO! Must be added here accidentally during git rebase. Thanks for catching this!

PR #182, opened 2019-08-23
SampleGen: access repeated fields and map fields in response handling

### LINKED ISSUE #180: SampleGen: response handling: support indexing into repeated fields
See: https://github.com/hzyi-google/sample-staging/blob/df0e9d64782c6d3fc61be167c7129a6e614d29d3/20190821/go-output/speech_transcribe_sync_gcs.go#L58

hzyi-google:
- Supports indexing into repeated fields using `[]` in response handling. ([want](https://github.com/googleapis/gapic-generator-go/compare/master...hzyi-google:output_array_and_map?expand=1#diff-72c8ff6387e4af3b7fd6ac18d3bab5ccR42))(Fix #180)

- Supports indexing into map fields using `{}` in response handling. (now this has some caveats...)

**tl;dr**
  1) in a `define` statement, indexing into the map field can <b>only</b> be the last part of the accessors ([want](https://github.com/googleapis/gapic-generator-go/compare/master...hzyi-google:output_array_and_map?expand=1#diff-72c8ff6387e4af3b7fd6ac18d3bab5ccR42)).
  2) we support field paths in `file_name` and `contents` of `write_file` statement. These two do **not** allow map fields.
  3) we support field paths in `print` statements. `print` does **not** allow map fields.
  4) therefore, to reference a subfield of a map value object, users need to call `define` first and assign the map value to a local variable, and then reference that variable or fields of that variable in following `write_file` and `print`.

********
****** Here's more explaination: ******

> 1) in a `define` statement, indexing into the map field can <b>only</b> be the last part of the accessors

for example:
```yaml
- response
  - define: a = $resp.my_map{"my_key"}            # this is allowed
  - define: a = $resp.my_map{"my_key"}.sub_field  # this will error!
```

**Do not allow** the second statement because:

- In Go, looking up a map entry by key usually required checking the returned `ok` to see if the key exists, thus the accessor chain has to stop at a map field. There is no way to generate idiomatic Go code using the second statement above.
```go
a, ok := $resp.myMap["my_key"]
if !ok {
    // handle error
}
```
(We talked about this a few weeks ago and I think most (if not all) of us agree this is the most appropriate way to handle this case (though not ideal). We rejected a few other proposals:
- Let the chain continue, do not check `ok` 
  - ---> not the appropriate way to look up a map
- Let the generator break up the chain and come up a local variable for the map value object 
  - ---> might not be able generate something the user want, and introducing variables users did not specify is every error-prone
********

>   2) we support field paths in `file_name` and `contents` of `write_file` statement. These two do **not** allow map fields.
>   3) we support field paths in `print` statements. `print` does **not** allow map fields.

- Similar to 1), it's not possible to inline the accessors within `ioutil.WriteFile` or `fmt.Sprintf` and check for `ok` and the same time.
********


hzyi-google:
PTAL

hzyi-google:
cc/ @googleapis/samplegen

cmd/gen-go-sample/inittree.go
```diff
@@ -155,10 +173,10 @@ func (t *initTree) get(k string, info pbinfo.Info) (*initTree, error) {
 							}
 						}
 					}
+					v.typ.repeated = false
```
noahdietz: I'm curious if this is necessary. Per protobuf3 [docs](https://developers.google.com/protocol-buffers/docs/proto3#maps), maps   cannot be repeated. I'd have to assume this case is impossible as this tool consumes a descriptor and protoc would fail if it encountered a `repeated map` field. The above checking for the repeated label (and the following assignment) would produce the necessary `false` value.

Does that track or am I missing a scenario?

cmd/gen-go-sample/out.go
```diff
@@ -217,7 +230,8 @@ func writeLoop(l *schema_v1p2.LoopSpec, st *symTab, gen *generator) error {
 	p := gen.pt.Printf
 
 	sc, report := initScanner(l.Collection)
-	path, typ, err := writePath(sc, st, gen.descInfo)
+	path, itree, err := writePath(sc, st, gen.descInfo, false)
+	typ := itree.typ
```
noahdietz: Can we put  the `typ` assignment after the error check? I understand that `itree` won't be `nil`, but for style sake lets keep the  error check immediately after the originating call 😄

hzyi-google:
PTAL

cmd/gen-go-sample/inittree.go
```diff
@@ -56,37 +58,65 @@ type initType struct {
...
 // initTree represents a node in the initialization tree.
 type initTree struct {
 	typ initType
 
 	// initTree is either a composite value (struct/array/map), where keys and vals are set,
 	// or a simple value, where leafVal is set.
 
+	// vals and keys keep tracks of all the child trees of this inittree and their identifiers.
+	// If the child is a protobuf field, key is the field name.
+	// If the child is an array element, key is the index.
```
vchudnov-g: Just so I understand: for a repeated proto field, there's an `initTree` with the index, and the child has a key with the field name? Or vice-versa?

cmd/gen-go-sample/inittree.go
```diff
@@ -56,37 +58,65 @@ type initType struct {
...
+}
+
 // initTree represents a node in the initialization tree.
 type initTree struct {
 	typ initType
 
 	// initTree is either a composite value (struct/array/map), where keys and vals are set,
 	// or a simple value, where leafVal is set.
 
+	// vals and keys keep tracks of all the child trees of this inittree and their identifiers.
```
vchudnov-g: nit: s/`inittree`/`initTree`/

cmd/gen-go-sample/main_test.go
```diff
@@ -274,6 +275,102 @@ func TestMapOut(t *testing.T) {
...
+			Rpc:       "UnaryMethod",
+			RegionTag: "awesome_region",
+			Response:  []schema_v1p2.ResponseConfig{r},
+		}
+
+		err := g.genSample(sp, GAPICMethod{Name: "UnaryMethod"})
+		if err == nil {
+			t.Errorf("expected error from response config: %v", r)
+		}
+		if !strings.Contains(err.Error(), expectedErrs[i]) {
```
vchudnov-g: [Aside] Project for another day: typing the errors and checking the types rather than the text. This is simple to do, but not trivial (similar to Python, but with one intermediate embedded struct)

cmd/gen-go-sample/primval.go
```diff
@@ -45,7 +45,7 @@ var validPrims = [...]func(string) bool{
 }
 
 func validStrings(s string) bool {
-	return true
+	return s[0] == '"'
```
vchudnov-g: Should we check that the last character is a quote as well? I suspect this is fine as is, because this is an initial sanity check before we rely on the scanner to tokenize fully.

cmd/gen-go-sample/inittree.go
```diff
@@ -301,6 +351,33 @@ func (t *initTree) parsePathRest(sc *scanner.Scanner, info pbinfo.Info) (*initTr
...
+				return nil, r, errors.E(nil, "cannot index via {} into a non-map field")
+			}
+
+			var tokVal string
+			if r := sc.Scan(); r == scanner.Int || r == scanner.String || r == scanner.Ident {
+				// Note we don't yet support use a locally defined a variable (by the `define:` directive)
+				// as a map key. scanner.Ident is here to handle the case when key is a bool value, as
+				// `true` and `false` (without quote) are parsed as scanner.Ident.
+				tokVal = sc.TokenText()
+				validPrim := validPrims[t.typ.keyType.prim]
```
vchudnov-g: See my responses on the other comment thread about this naming. These are two different types that differ by just one letter. We should fix this in a follow-up PR.

cmd/gen-go-sample/inittree.go
hzyi-google: > for a repeated proto field, there's an initTree with the index, and the child has a key with the field name?
Yes!

cmd/gen-go-sample/main_test.go
hzyi-google: That makes sense!

cmd/gen-go-sample/primval.go
hzyi-google: Yeah, but that's a good point, it does not hurt to check the last character. I'll add it.

cmd/gen-go-sample/inittree.go
```diff
@@ -301,6 +337,33 @@ func (t *initTree) parsePathRest(sc *scanner.Scanner, info pbinfo.Info) (*initTr
...
+				return nil, r, errors.E(nil, "cannot index into a non-map field")
+			}
+
+			var tokVal string
+			if r := sc.Scan(); r == scanner.Int || r == scanner.String || r == scanner.Ident {
+				// Note we don't yet support use a locally defined a variable (by the `define:` directive)
+				// as a map key. scanner.Ident is here to handle the case when key is a bool value, as
+				// `true` and `false` (without quote) are parsed as scanner.Ident.
+				tokVal = sc.TokenText()
+				validPrim := validPrims[t.typ.keyType.prim]
```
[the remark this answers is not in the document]
hzyi-google: Whoops I don't know why I thought `valid` is a verb! My bad!

cmd/gen-go-sample/inittree.go
hzyi-google: Issue filed: https://github.com/googleapis/gapic-generator-go/issues/190

PR #183, opened 2019-08-29
SampleGen: generate comments on fields of the request object

hzyi-google:
- when `value_is_file: true`, the comments are above reading bytes from the file.
- when the field is a part of the resource name, the comments are above constructing the resource name.
- for other fields, the comments are above the field in constructing the request object.

Comments for sample function parameters are not handled yet.


hzyi-google:
cc/ @googleapis/samplegen

cmd/gen-go-sample/testdata/sample_enum.want
```diff
@@ -34,6 +34,7 @@ func sampleUnaryMethod(myFruit foopb.AType_FruitEnum) error {
 	}
 
 	// myFruit := foopb.AType_APPLE
+
```
noahdietz: I'm not sure if I like this line break. I feel like it should be "on top" of the usage of said variable...is it to separate the inline comment from the generated value comments?

cmd/gen-go-sample/inittree.go
```diff
@@ -511,3 +515,21 @@ func snakeToCamel(s string) string {
...
+		indent(w, ind)
+		w.WriteString("// ")
+		w.WriteString(c)
+		w.WriteByte('\n')
+	}
+}
+
+func indent(w *bufio.Writer, ind int) {
+	for i := 0; i < ind; i++ {
+		w.WriteByte('\t')
```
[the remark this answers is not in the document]
software-dov: +1 for spaces instead of tabs for consistency.

Minor aside: how important is it to be diligent about formatting? Is it better to just do best-effort and let gofmt do the heavy lifting?

cmd/gen-go-sample/testdata/sample_enum.want
software-dov: +1 to Noah's comment.

cmd/gen-go-sample/inittree.go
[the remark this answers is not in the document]
noahdietz: > how important is it to be diligent about formatting?

While most of the generation was written by Mike D., in gapic-generator-go we try to stay roughly aligned with proper formatting, including indentation of code blocks. It makes the generation code much easier to read/mentall parse. Something like an empty curly-brace pair being on two lines isn't a big deal, it gets fixed-up by gofmt.

cmd/gen-go-sample/bytes.go
```diff
@@ -36,6 +37,8 @@ type fileInfo struct {
 	fileName string
 	// varName is the name of the local variable to hold the bytes of the file.
 	varName string
+
+	comment string
```
vchudnov-g: add a field comment for `comment`
(also, for this comment and the previous one, use a preceding blank line to separate it visually from the previous field)

cmd/gen-go-sample/inittree.go
```diff
@@ -511,3 +515,21 @@ func snakeToCamel(s string) string {
...
+	}
+	for _, c := range strings.Split(strings.TrimRight(comment, "\n"), "\n") {
+		indent(w, ind)
+		w.WriteString("// ")
+		w.WriteString(c)
+		w.WriteByte('\n')
+	}
+}
+
+func indent(w *bufio.Writer, ind int) {
```
vchudnov-g: I would suggest `indent` return a `string`, and  `printCommentLines` could simply use that

cmd/gen-go-sample/inittree.go
```diff
@@ -511,3 +515,21 @@ func snakeToCamel(s string) string {
 	}
 	return sb.String()
 }
+
+func printCommentLines(w *bufio.Writer, comment string, ind int) {
+	if comment == "" {
+		return
+	}
+	for _, c := range strings.Split(strings.TrimRight(comment, "\n"), "\n") {
+		indent(w, ind)
```
vchudnov-g: How about simply
```
w.WriteString(fmt.Sprintf("%s// %c\n", indent(ind), c))
```

(with `indent` returning a `string`?)

cmd/gen-go-sample/inittree.go
```diff
@@ -511,3 +515,21 @@ func snakeToCamel(s string) string {
 	}
 	return sb.String()
 }
+
+func printCommentLines(w *bufio.Writer, comment string, ind int) {
+	if comment == "" {
```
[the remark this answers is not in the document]
vchudnov-g: Uhm, no? https://play.golang.org/p/Ybm3uddpCE2

cmd/gen-go-sample/testdata/sample_enum.want
vchudnov-g: I tend to agree the blank line looks weird here. I was wondering whether when we initialize multiple fields, it would make things more readable to have a blank line, but I'm not sure it will.

cmd/gen-go-sample/inittree.go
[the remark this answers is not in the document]
hzyi-google: @noahdietz  I got the same result as @vchudnov-g . The loop will execute when comment is an empty string.

cmd/gen-go-sample/main.go
```diff
@@ -440,6 +440,9 @@ func (g *generator) handleRequest(initInfo initInfo) error {
 		buf.WriteByte('\n')
 	}
 	prependLines(&buf, "// ", false)
+  if len(initInfo.argNames) > 0 {
```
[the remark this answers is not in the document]
hzyi-google: It looks quite OK on my side - There are three new lines of code here?

cmd/gen-go-sample/inittree.go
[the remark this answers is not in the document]
noahdietz: ah jeez i forgot to change my own test haha my bad!

cmd/gen-go-sample/inittree.go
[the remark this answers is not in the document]
hzyi-google: The entire sample gen project uses tabs to indent.

And I found this: 

> Indentation
> We use tabs for indentation and gofmt emits them by default. Use spaces only if you must.

source: https://stackoverflow.com/questions/19094704/indentation-in-go-tabs-or-spaces

cmd/gen-go-sample/inittree.go
hzyi-google: `indent` is also used elsewhere. (It's also not new code either).

cmd/gen-go-sample/inittree.go
hzyi-google: Same as below: `indent` is also used elsewhere. It does make the code shorter, but it will cause too much changes.

PR #216, opened 2019-09-30
SampleGen: Add function-level documentation

hzyi-google:
Generates function-level comments.
* a comment describing what the sample does. Generated from `description` field in sample config.
* comments describing what each parameter is. Generated from `request.comment` field in sample config.

Open questions:
* Not following GoDoc style?
  * @jadekler mentioned in Chat that samples don't necessarily need to follow GoDoc styles.
  * It's hard to implement because that will require either the generator to construct a grammar correct English sentence from user-specified sample config or enforcing some conventions to the `request.comment` field in sample config.
* Is it OK to use a `:` between the parameter name and the comment?

### DIFF
--- a/cmd/gen-go-sample/main.go
+++ b/cmd/gen-go-sample/main.go
@@ -353,7 +353,7 @@ func (g *generator) genSample(sampConf schema_v1p2.Sample, methConf GAPICMethod)
 
 	p := g.pt.Printf
 
-	// Region tag, function signature and new client
+	// Region tag
 	argStr, err := argListStr(initInfo, g)
 	if err != nil {
 		return err
@@ -361,6 +361,30 @@ func (g *generator) genSample(sampConf schema_v1p2.Sample, methConf GAPICMethod)
 	p("")
 	p("// [START %s]", sampConf.RegionTag)
 	p("")
+
+  // comments
+  emptyArr := make([]string, 0)
+  requiresNewLine := false
+  
+  if sampConf.Description != "" {
+  	writeComment(sampConf.Description, emptyArr, g)
+  	requiresNewLine = true
+  }
+
+  for i, argName := range initInfo.argNames {
+  	comment := initInfo.argTrees[i].comment
+  	if comment == "" {
+  		continue
+  	}
+  	if requiresNewLine {
+  		writeComment("", emptyArr, g)
+  	}
+  	requiresNewLine = true
+  	comment = fmt.Sprintf("%s: %s", argName, comment)
+  	writeComment(comment, emptyArr, g)
+  }
+
+  // function signature and initialize a new client
 	p("func sample%s(%s) error {", meth.GetName(), argStr)
 	p("  ctx := context.Background()")
 	p("  c, err := %s.New%sClient(ctx)", g.clientPkg.Name, pbinfo.ReduceServName(serv.GetName(), g.clientPkg.Name))
--- a/cmd/gen-go-sample/main.go
+++ b/cmd/gen-go-sample/main.go
@@ -362,17 +362,17 @@ func (g *generator) genSample(sampConf schema_v1p2.Sample, methConf GAPICMethod)
 	p("// [START %s]", sampConf.RegionTag)
 	p("")
 
-  // comments
-  emptyArr := make([]string, 0)
+  // generate comments above the sample function
+  var emptyArr []string = nil
   requiresNewLine := false
   
   if sampConf.Description != "" {
-  	writeComment(sampConf.Description, emptyArr, g)
+  	writeComment(strings.TrimSuffix(sampConf.Description, "\n"), emptyArr, g)
   	requiresNewLine = true
   }
 
   for i, argName := range initInfo.argNames {
-  	comment := initInfo.argTrees[i].comment
+  	comment := strings.TrimSuffix(initInfo.argTrees[i].comment, "\n")
   	if comment == "" {
   		continue
   	}
--- a/cmd/gen-go-sample/main.go
+++ b/cmd/gen-go-sample/main.go
@@ -372,7 +372,7 @@ func (g *generator) genSample(sampConf schema_v1p2.Sample, methConf GAPICMethod)
 	}
 
 	if sampConf.Description != "" {
-		writeCommentLines(sampConf.Description)
+		writeCommentLines(fmt.Sprintf("sample%s: %s", meth.GetName(), sampConf.Description))
 		requiresNewLine = true
 	}
 
@@ -385,8 +385,7 @@ func (g *generator) genSample(sampConf schema_v1p2.Sample, methConf GAPICMethod)
 			writeCommentLines("")
 		}
 		requiresNewLine = true
-		comment = fmt.Sprintf("%s: %s", argName, comment)
-		writeCommentLines(comment)
+		writeCommentLines(fmt.Sprintf("%s: %s", argName, comment))
 	}
 
 	// function signature and initialize a new client
--- a/cmd/gen-go-sample/main.go
+++ b/cmd/gen-go-sample/main.go
@@ -363,7 +363,7 @@ func (g *generator) genSample(sampConf schema_v1p2.Sample, methConf GAPICMethod)
 	p("// [START %s]", sampConf.RegionTag)
 	p("")
 
-	// generate comments above the sample function
+	// comments above the sample function
 	requiresNewLine := false
 
 	writeCommentLines := func(comment string) {
@@ -716,4 +716,43 @@ func prependLines(b *bytes.Buffer, prefix string, skipEmptyLine bool) {
 		}
 		b.WriteString(l)
 	}
+}
+
+// wrapComment wraps comment at 100 characters, iff comment does not contain any newline characters
+// and comment has more than 110 characters.
+//
+// comment cannot have leading or trailing white spaces.
+func wrapComment(comment string) string {
+	if strings.ContainsRune(comment, '\n') || len(comment) < 110 {
+		return comment
+	}
+
+	var output strings.Builder
+	s := 0
+	prev := -1
+
+	for true {
+		fmt.Println("what")
+		p := strings.IndexByte(comment[prev+1:], ' ')
+		// we reached the end of comment
+		if p < 0 {
+			output.WriteString(comment[s:])
+			return output.String()
+		}
+
+		p = p + prev + 1
+		if p-s > 100 {
+			// a single word has more than 100 characters
+			if prev == s-1 {
+				prev = p
+			}
+			// break the line at prev
+			output.WriteString(comment[s:prev])
+			output.WriteByte('\n')
+			s = prev + 1
+		}
+		prev = p
+	}
+
+	return ""
 }
--- a/cmd/gen-go-sample/main.go
+++ b/cmd/gen-go-sample/main.go
@@ -313,6 +313,7 @@ func (g *generator) disambiguateSampleIDs() error {
 	return nil
 }
 
+// TODO(hzyi): this method is getting long. Split it up.
 func (g *generator) genSample(sampConf schema_v1p2.Sample, methConf GAPICMethod) error {
 	ifaceName := sampConf.Service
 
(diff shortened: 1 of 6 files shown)


hzyi-google:
PTAL

hzyi-google:
cc/ @googleapis/samplegen

cmd/gen-go-sample/main.go
```diff
@@ -353,14 +353,38 @@ func (g *generator) genSample(sampConf schema_v1p2.Sample, methConf GAPICMethod)
...
+	// Region tag
 	argStr, err := argListStr(initInfo, g)
 	if err != nil {
 		return err
 	}
 	p("")
 	p("// [START %s]", sampConf.RegionTag)
 	p("")
+
+  // comments
```
noahdietz: Can we have a slightly more descriptive comment here please?

cmd/gen-go-sample/main.go
```diff
@@ -353,14 +353,38 @@ func (g *generator) genSample(sampConf schema_v1p2.Sample, methConf GAPICMethod)
...
+
+  for i, argName := range initInfo.argNames {
+  	comment := initInfo.argTrees[i].comment
+  	if comment == "" {
+  		continue
+  	}
+  	if requiresNewLine {
+  		writeComment("", emptyArr, g)
+  	}
+  	requiresNewLine = true
```
noahdietz: so this never gets set back to `false`, meaning that after a single comment, a new line will precede the next comment. Is my understanding correct? Is it *just* a new line for the next thing or a blank line? What if the comment has a new line at the end, does it double up? Or why couldn't we write the comment string and append a newline to it before writing? This just feels a little wonky to me and I'd like to talk it out.

cmd/gen-go-sample/main.go
```diff
@@ -353,14 +353,38 @@ func (g *generator) genSample(sampConf schema_v1p2.Sample, methConf GAPICMethod)
...
+  }
+
+  for i, argName := range initInfo.argNames {
+  	comment := initInfo.argTrees[i].comment
+  	if comment == "" {
+  		continue
+  	}
+  	if requiresNewLine {
+  		writeComment("", emptyArr, g)
```
noahdietz: Could we use `nil` instead of an empty slice? Typed `nil` variables will take the empty value of their type, in this case, an empty string slice. I think `writeComment` would work still. The `emptyArr` is just kind of confusing imo

Test: https://play.golang.org/p/n7GluDmGICm

cmd/gen-go-sample/main.go
hzyi-google: It is a blank line between the method comment and each parameter comment.

> What if the comment has a new line at the end, does it double up?
I think it shouldn't, mostly because multi-line strings in YAML always have a newline character in the end, but users don't want that to actually result in a new line. However, I apparently missed this case, and I'll add it shortly.

> Or why couldn't we write the comment string and append a newline to it before writing?
We can't append a newline to the last comment, that will cause an empty comment line between the last comment and the function signature.

cmd/gen-go-sample/main.go
```diff
@@ -353,14 +353,38 @@ func (g *generator) genSample(sampConf schema_v1p2.Sample, methConf GAPICMethod)
...
 	argStr, err := argListStr(initInfo, g)
 	if err != nil {
 		return err
 	}
 	p("")
 	p("// [START %s]", sampConf.RegionTag)
 	p("")
+
+  // generate comments above the sample function
+  var emptyArr []string = nil
```
vchudnov-g: Weird indentation. Run `gofmt` on our source files?

cmd/gen-go-sample/main.go
```diff
@@ -362,17 +362,17 @@ func (g *generator) genSample(sampConf schema_v1p2.Sample, methConf GAPICMethod)
...
 
-  // comments
-  emptyArr := make([]string, 0)
+  // generate comments above the sample function
+  var emptyArr []string = nil
   requiresNewLine := false
   
   if sampConf.Description != "" {
-  	writeComment(sampConf.Description, emptyArr, g)
+  	writeComment(strings.TrimSuffix(sampConf.Description, "\n"), emptyArr, g)
```
noahdietz: I might just use [`strings.TrimSpace`](https://golang.org/pkg/strings/#TrimSpace) to normalize the comment, removing leading & trailing white space. WDYT? Is leading white space necessary?

cmd/gen-go-sample/main.go
```diff
@@ -362,17 +362,17 @@ func (g *generator) genSample(sampConf schema_v1p2.Sample, methConf GAPICMethod)
...
   	requiresNewLine = true
   }
 
   for i, argName := range initInfo.argNames {
-  	comment := initInfo.argTrees[i].comment
+  	comment := strings.TrimSuffix(initInfo.argTrees[i].comment, "\n")
```
noahdietz: same thing here [`strings.TrimSpace`](https://golang.org/pkg/strings/#TrimSpace)

cmd/gen-go-sample/main.go
```diff
@@ -353,14 +353,38 @@ func (g *generator) genSample(sampConf schema_v1p2.Sample, methConf GAPICMethod)
...
+	// Region tag
 	argStr, err := argListStr(initInfo, g)
 	if err != nil {
 		return err
 	}
 	p("")
 	p("// [START %s]", sampConf.RegionTag)
 	p("")
+
+  // generate comments above the sample function
```
vchudnov-g: (P2, No need to do this now) This function is starting to get a bit long. I'd consider breaking each section into helper functions that you call sequentially.

cmd/gen-go-sample/testdata/sample_unary.want
```diff
@@ -28,6 +28,14 @@ foopb "path.to/pb/foo"
 
 // [START awesome_region]
 
+// Construct a complex request object,
+// send it to the server,
+// and inspect the response.
+//
+// theX: a single-line comment for an input parameter
```
vchudnov-g: If the comment is a single line but is really long, do you insert it as is or do you line-wrap it at some length?

cmd/gen-go-sample/main.go
noahdietz: I was more thinking you'd just supply `nil` to the function call `writeComment("", nil, g)` because `writeComment` will know how to handle it.

You could also wrap `writeComment` with a `writeCommentLine(gen *generator) error` method

```go
// writeCommentLine adds a commented line break to a comment block
func writeCommentLine(gen *generator) error {
  return writeComment("", nil, gen)
}
```


The latter suggestion might actually read better IMO. And wrapping the "generic" method to do something specific in its scope is totally reasonable.

WDYT?

cmd/gen-go-sample/testdata/sample_unary.want
```diff
@@ -28,6 +28,14 @@ foopb "path.to/pb/foo"
 
 // [START awesome_region]
 
+// Construct a complex request object,
```
vchudnov-g: Though I get the problem with changing the verb mood, one thing you could do is go half-way to the Go standard and include the `functionName + ": " + description`, so that this reads, for example,

```
// sampleUnaryMethod: Construct a complex request object
```

or **even better**, `functionName + " will " + lowerFirstChar(description)`:

```
// sampleUndaryMethod will construct a complex request object
```

I _think_ the latter we can do safely and always grammatically (assuming that the verb is always imperative as in this example ["Construct ..."] and not declarative [ie "Constructs ..." would break this]), and that ought to be sure to shut up `gofmt`. The former may or may not shut up `gofmt`.

Is there a standard/prevalence of imperative vs declarative in proto comments?

cmd/gen-go-sample/testdata/sample_unary.want
vchudnov-g: The colons are fine, I think. Maybe we have a preceding line saying `// Parameters:`? But I don't think that's strictly needed.

cmd/gen-go-sample/main.go
noahdietz: > I think it shouldn't...I'll add it shortly.

sgtm

> We can't append a newline to the last comment, that will cause an empty comment line between the last comment and the function signature.

Ack. There might be a more _explicit_ way to do this (i.e. break before the last item and write it immediately following the loop), but it would be a few more lines of code and it doesn't matter all that much.

cmd/gen-go-sample/testdata/sample_unary.want
hzyi-google: There are no standard for proto comments AFAIK. AIP however, has requirements (https://aip.dev/192#style).

I am onboard with function name + : + rest of comments. I am a bit reluctant to do function name + "will" + lowerFirstChar(description) because we are assuming the description starts with a verb and the verb is not in third-person singular form, the latter especially, because that style is fairly common (at least AIP and Java are two that I know of).

cmd/gen-go-sample/testdata/sample_unary.want
hzyi-google: Not doing line-wrapping right now. Good point. I'll add it.

cmd/gen-go-sample/main.go
hzyi-google: Yep I have the same feeling. I'll add a note.

cmd/gen-go-sample/main.go
hzyi-google: > break before the last item and write it immediately following the loop...
The problem is that not all parameters have comments. So this would require two passes through all the parameters: the first to find all parameters with comments, and the second generate them.

cmd/gen-go-sample/main.go

cmd/gen-go-sample/main.go

cmd/gen-go-sample/main.go

cmd/gen-go-sample/testdata/sample_unary.want
hzyi-google: @vchudnov-g @noahdietz Go doesn't seem to strictly enforce line length for comments (it is actually discouraged): https://github.com/golang/go/wiki/CodeReviewComments#line-length

> break lines because of the semantics of what you're writing (as a general rule) and not because of the length of the line. If you find that this produces lines that are too long, then change the names or the semantics and you'll probably get a good result.

Now that users can break lines themselves with multi-line YAML strings, I think it should be OK that we don't forcing line-wrapping in the generator. I am OK if we force newlines when the line becomes super long (e.g., 120 or 140 characters), but that to me is trying to be smart (and I don't like writing code that tries to be smart, because there are always weird edge cases).

noahdietz: For the reasons you've stated, I'm cool with not doing automatic line breaks

hzyi-google:
PTAL

cmd/gen-go-sample/testdata/sample_unary.want
vchudnov-g: I know Go *allows* long lines, but I don't want us to generate uncomfortably long lines. Here's why I think  I think we should consider wrapping:
- In the other languages, we wrap lines, right?
- So then users won't be thinking about automatically inserting new lines. This is good, less for them to think about in the usual case (say they're writing a paragraph).
- If we then force them to insert line breaks for Go, then they might do it at, say, column 100. But for some other languages the word wrap happens at column 80. So then for those languages, there would be an ~80 column line, followed by a ~20 column line, followed by an ~80 column line, all part of the same paragraph. That does not look idiomatic!

I would propose something like this for Go: if the configured comment is exactly one line of less than 110 characters, output it as one line. Otherwise, wrap all the logical lines of the comment at 100 characters. (Those numbers are arbitrary, but seem reasonable for ease of reading.)

Notice that I am implicitly making a distinction between *semantic* line breaks (which the use should configure in the input when formatting matters) from *cosmetic* line breaks (which the user should not have to worry about; they're just about making the comment easier to read for the end developer).

I'm fine deferring this to a separate PR, but I would ask that we file an issue to keep track of it and see what feedback we get about comment length. (I'm also fine with implementing this now if it's easy :smile: . )

cmd/gen-go-sample/testdata/sample_unary.want
vchudnov-g: Makes sense. The `:` solution sounds like the best compromise between what people might write in their protos and what Go idiomaticity requires.

cmd/gen-go-sample/main.go
```diff
@@ -353,14 +354,41 @@ func (g *generator) genSample(sampConf schema_v1p2.Sample, methConf GAPICMethod)
...
+			continue
+		}
+		if requiresNewLine {
+			writeCommentLines("")
+		}
+		requiresNewLine = true
+		writeCommentLines(fmt.Sprintf("%s: %s", argName, comment))
+	}
+
+	// function signature and initialize a new client
```
vchudnov-g: nit: parallel structure between comments and within this comment

How about `// function signature and client initialization`?

cmd/gen-go-sample/main.go
```diff
@@ -353,14 +354,41 @@ func (g *generator) genSample(sampConf schema_v1p2.Sample, methConf GAPICMethod)
...
+	// Region tag
 	argStr, err := argListStr(initInfo, g)
 	if err != nil {
 		return err
 	}
 	p("")
 	p("// [START %s]", sampConf.RegionTag)
 	p("")
+
+	// generate comments above the sample function
```
vchudnov-g: nit: parallel structure across comments

How about s/`generate `//?

cmd/gen-go-sample/testdata/sample_unary.want
hzyi-google: So there are several cases here:

1) User input is a very very long one-line comment, say, 200 characters. In this case it's preferable that the generator breaks the comment into multiple lines.

2) Use input follows Go patterns (i.e., break lines because of the semantics of what they're writing), and one of the lines is slightly long, say, 111 characters. In this case I would prefer to not to break lines in the generator, because that will cause cascading effects to all following lines, which might change the line breaking that the user tried to maintain.

When it comes to cosmetic formatting, I am more inclined to have the generator to "prevent very bad things" rather than "try to make every case as good as possible". So how about we only handle case 1) here?

If we had to wrap multi-line user input as well, then we would have two choices:
a) preserves user specified line breaks, and also breaks at maximum line length. This will potentially introduce some very short lines :(

b) ignore user specified line breaks, and always breaks at maximum line length. This is just doing the exact opposite of what Go guidance prefers. :(

I just don't like either of them.

vchudnov-g: > Use input follows Go patterns (i.e., break lines because of the semantics of what they're writing), 

The user usually won't if they're thinking about more than one language. That's what my example was about.

> and one of the lines is slightly long, say, 111 characters. In this case I would prefer to not to break lines in the generator, because that will cause cascading effects to all following lines, which might change the line breaking that the user tried to maintain.

That's why my proposal above said (emphasis added) "if the configured comment is **exactly one line** of less than 110 characters, output it as one line" 

> a) preserves user specified line breaks, and also breaks at maximum line length. This will potentially introduce some very short lines :(

That's why I proposed breaking only for single-paragraph comments. I'm guessing this will get us the 80% case.... And even if we do break for multi-paragraph comments, my proposal has some margin to prevent ultra-short lines: if the line is longer than 110 characters, break by the column 100. (Notice the difference in numbers.)

(I'd like to also be able to break for multi-line comments that don't rely on formatting, but determining that is a harder problem that is not a good use of our time right now, IMO.)

I agree, no on (b) ;-)

hzyi-google: Okay I think we are on the same page. I misunderstood "Otherwise" earlier, I thought that refers to "wrap all the logical lines of the comment at 100 characters when user input is either one line with more than 110 characters or multiple lines".

PR #218, opened 2019-10-02
Have gengapic protoc plugin invoke samplegen

hzyi-google:
- Move most of `cmd/gen-go-sample` to `intenal/gensample`, but leave `main.go` still in `cmd/gen-go-sample`. This makes sure that samplegen can continue to run as a standalone program, and gapic-gen and invoke samplegen

- Add `--sample`, `--gapic-config` and `--sample-only` flag to docker entry point, and pass them along to `gensample` and `gengapic`.
  - When `--sample-only` is present, toggle off `gengapic` and `gapic_validator_out`, so that we can generate samples for APIs that are not yet properly annotated.
  - `gensample` will consume gapic config and sample config.


internal/gensample/gensample.go
```diff
@@ -0,0 +1,651 @@
...
+
+	var wg sync.WaitGroup
+
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		if gapicFname != "" {
+			f, err := os.Open(gapicFname)
+			if err != nil {
+				log.Fatal(errors.E(err, "cannot read GAPIC config file: %q", gapicFname))
```
noahdietz: Can we refactor the goroutine erros to use a `chan` to pass errors back to to the main routine to be returned by the calling function (`InitGen`)? Since this code is a "library" now, it shouldn't panic (which `log.Fatal` does), but rather pass errors back up the call chain. This way, `gen-go-sample` and `protoc-gen-go_gapic` can handle the error however they'd like.

internal/gensample/gensample.go
```diff
@@ -0,0 +1,651 @@
...
+	"github.com/googleapis/gapic-generator-go/internal/license"
+	"github.com/googleapis/gapic-generator-go/internal/pbinfo"
+	"github.com/googleapis/gapic-generator-go/internal/printer"
+	yaml "gopkg.in/yaml.v2"
+)
+
+const expectedSampleConfigType = "com.google.api.codegen.samplegen.v1p2.SampleConfigProto"
+const expectedSampleConfigVersion = "1.2.0"
+
+func InitGen(desc []*descriptor.FileDescriptorProto, sampleFnames []string, gapicFname string, clientPkg string, nofmt bool) (*generator, error) {
```
noahdietz: godoc comment please

internal/gensample/gensample.go
```diff
@@ -0,0 +1,651 @@
...
+			log.Fatal(err)
+		}
+	}()
+
+	wg.Wait()
+	gen.nofmt = nofmt
+	return &gen, nil
+}
+
+func (gen *generator) GenMethodSamples() error {
```
noahdietz: godoc comment please

internal/gensample/gensample.go
hzyi-google: Yes!

cmd/gen-go-sample/main.go
```diff
@@ -15,34 +15,22 @@
...
-	"github.com/googleapis/gapic-generator-go/internal/pbinfo"
-	"github.com/googleapis/gapic-generator-go/internal/printer"
-	yaml "gopkg.in/yaml.v2"
+	"github.com/googleapis/gapic-generator-go/internal/gensample"
 )
 
-const expectedSampleConfigType = "com.google.api.codegen.samplegen.v1p2.SampleConfigProto"
-const expectedSampleConfigVersion = "1.2.0"
+const (
+	paramError = "need parameter in format: go-gapic-package=client/import/path;packageName"
```
vchudnov-g: This does not seem to be used

internal/gengapic/client_init_test.go
```diff
@@ -21,12 +21,12 @@ import (
 	"github.com/golang/protobuf/proto"
 	"github.com/golang/protobuf/protoc-gen-go/descriptor"
 	"github.com/golang/protobuf/ptypes/duration"
+	wrappers "github.com/golang/protobuf/ptypes/wrappers"
```
vchudnov-g: why the reordering? what's the sort criteria?

internal/gensample/gensample.go
vchudnov-g: +1
for most of the functions in this file, but especially the exported ones

internal/gengapic/client_init_test.go
hzyi-google: That's automatic changes by running `go fmt`

hzyi-google:
Merge conflicts solved! PTAL

I noticed that github is not showing diffs in `gensample.go` because it is duplicated from `cmd/main.go`. I can make a duplicate `gensample.go` from `main.go` without any change and submit it as a separate PR, so that the diffs will show up here.

Let me know what you folks think.

internal/gensample/gensample.go
```diff
@@ -0,0 +1,740 @@
...
+	}()
+
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		if err := gen.readSampleConfigFiles(sampleFnames); err != nil {
+			errChan <- err
+			return
+		}
+		errChan <- nil
```
noahdietz: this could probably just be 

```go
defer wg.Done()
errChan <- gen.readSampleConfigFiles(sampleFnames)
```

Or save the return value into `err` and send that to the `chan` if you'd prefer.

internal/gensample/gensample.go
```diff
@@ -0,0 +1,740 @@
...
+		return nil, errors.E(nil, "need -clientPkg in 'url/to/client/pkg;name' format, got %q", clientPkg)
+	}
+
+	var wg sync.WaitGroup
+	errChan := make(chan error)
+
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		if gapicFname != "" {
```
noahdietz: Let's invert this to reduce indentation

```go
if gapicFname == "" {
  errChan <- nil
  return
}
```

internal/gensample/gensample.go
```diff
@@ -0,0 +1,737 @@
...
+		argTrees:  argTrees,
+		flagNames: flagNames,
+		files:     files,
+		reqTree:   itree,
+	}
+
+	return initInfo, nil
+}
+
+func argListStr(init initInfo, g *generator) (string, error) {
```
vchudnov-g: function comment

internal/gensample/gensample.go
```diff
@@ -0,0 +1,737 @@
...
+
+	// main
+	if err := writeMain(g, initInfo.argNames, initInfo.flagNames, initInfo.argTrees, meth.GetName()); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (g *generator) getInitInfo(inType pbinfo.ProtoType, methConf GAPICMethod, fieldConfs []schema_v1p2.RequestConfig) (initInfo, error) {
```
vchudnov-g: function comment: summarize the purpose of this

internal/gensample/plugin.go
```diff
@@ -0,0 +1,78 @@
...
+
+	plugin "github.com/golang/protobuf/protoc-gen-go/plugin"
+	"github.com/googleapis/gapic-generator-go/internal/errors"
+)
+
+const (
+	paramError = "need parameter in format: go-gapic-package=client/import/path;packageName"
+)
+
+// PluginEntry is the entry point of SampleGen as a protoc plugin, together with CodeGen
```
vchudnov-g: What does "together with CodeGen" mean? Clarify or omit....

internal/gensample/plugin.go
```diff
@@ -0,0 +1,78 @@
...
+func PluginEntry(genReq *plugin.CodeGeneratorRequest) (*plugin.CodeGeneratorResponse, error) {
```
vchudnov-g: We are having one protoc plugin only that can generate one or both of the library and the sample, right? I thought you were keeping samplegen as its own binary, but we're having only one plugin?

hzyi-google: Yeah, samplegen can continue to run as a standalone go binary. (We already have this, and installing/running go binaries is easy, so why not?)

But when running the microgenerator as a plugin, if sample configs are specified, samples are also generated. (The setting is very similar to Python micro).
 
In an ideal world, when all APIs are annotated, DPEs just need to run the microgenerator (as a docker or a plugin), and can get all the code and samples for an API.

hzyi-google: (That reminds me that I should have @software-dov for review, to ensure the consistency between Go and Python micro generators)

internal/gensample/gensample.go
```diff
@@ -0,0 +1,737 @@
...
+	return nil
+}
+
+type generator struct {
+	desc         []*descriptor.FileDescriptorProto
+	descInfo     pbinfo.Info
+	gapic        GAPICConfig
+	sampleConfig schema_v1p2.SampleConfig
+	clientPkg    pbinfo.ImportSpec
+	nofmt        bool
```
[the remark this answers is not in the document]
hzyi-google: `nofmt` has been used a lot in the repo. It'll be a lot of change but definitely doable. I don't think it's too bad considering the existence of `gofmt` command which `nofmt` looks very similar to, but happy to do a global replacement if you insist.

internal/gensample/gensample.go
```diff
@@ -0,0 +1,737 @@
...
+}
+
+func (g *generator) reset() {
+	g.pt.Reset()
+	for imp := range g.imports {
+		delete(g.imports, imp)
+	}
+}
+
+func (g *generator) commit(gofmt bool, year int) ([]byte, error) {
```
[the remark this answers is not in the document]
hzyi-google: I didn't write this function but I actually think this is a great name, because this method does is exactly to commit what is around in the buffer inside `generator` to bytes.

But I agree this is a bit confusing for folks who haven't been working closely with this part of the code. I'll add documentation to list out what `commit` does.

internal/gensample/gensample.go
```diff
@@ -0,0 +1,737 @@
...
+	"github.com/googleapis/gapic-generator-go/internal/license"
+	"github.com/googleapis/gapic-generator-go/internal/pbinfo"
+	"github.com/googleapis/gapic-generator-go/internal/printer"
+	yaml "gopkg.in/yaml.v2"
+)
+
+const expectedSampleConfigType = "com.google.api.codegen.samplegen.v1p2.SampleConfigProto"
+const expectedSampleConfigVersion = "1.2.0"
+
+// InitGen creates a new sample generator.
```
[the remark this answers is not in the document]
hzyi-google: Done.

internal/gensample/gensample.go
```diff
@@ -0,0 +1,737 @@
...
+		gen.clientPkg = pbinfo.ImportSpec{Path: (clientPkg)[:p], Name: (clientPkg)[p+1:]}
+	} else {
+		return nil, errors.E(nil, "need -clientPkg in 'url/to/client/pkg;name' format, got %q", clientPkg)
+	}
+
+	var wg sync.WaitGroup
+	errChan := make(chan error)
+
+	wg.Add(1)
+	go func() {
```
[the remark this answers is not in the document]
hzyi-google: Did the second suggestion (love the modularity!)

internal/gensample/gensample.go
```diff
@@ -0,0 +1,737 @@
...
+	}()
+
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		errChan <- gen.readSampleConfigFiles(sampleFnames)
+	}()
+
+	gen.nofmt = nofmt
+	if err := <-errChan; err != nil {
```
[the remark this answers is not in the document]
hzyi-google: Done.

internal/gensample/gensample.go
```diff
@@ -0,0 +1,737 @@
...
+	sampleConfig schema_v1p2.SampleConfig
+	clientPkg    pbinfo.ImportSpec
+	nofmt        bool
+
+	pt      printer.P
+	imports map[pbinfo.ImportSpec]bool
+	Outputs map[string][]byte
+}
+
+// readSampleConfigFiles loads sample configs from local files and store
```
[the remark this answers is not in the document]
hzyi-google: Done.

internal/gensample/gensample.go
```diff
@@ -0,0 +1,737 @@
...
+
+func (g *generator) reset() {
+	g.pt.Reset()
+	for imp := range g.imports {
+		delete(g.imports, imp)
+	}
+}
+
+func (g *generator) commit(gofmt bool, year int) ([]byte, error) {
+	// We'll gofmt unless user asks us to not, so no need to think too hard about sorting
```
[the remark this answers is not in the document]
hzyi-google: Done.

internal/gensample/gensample.go
```diff
@@ -0,0 +1,737 @@
...
+		if samples[i].ID == "" {
+			samples[i].ID = samples[i].RegionTag
+		}
+		if samples[i].ID != "" {
+			idCount[samples[i].ID]++
+		}
+	}
+
+	for i := range samples {
+		// Generate a sample ID when the user-provided ID is empty not unique
```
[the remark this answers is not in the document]
hzyi-google: Done.

internal/gensample/gensample.go
```diff
@@ -0,0 +1,737 @@
...
+		if samples[i].ID == "" || idCount[samples[i].ID] > 1 {
+			jsonStr, err := json.Marshal(samples[i])
+			if err != nil {
+				return err
+			}
+			checkSum := sha256.Sum256([]byte(jsonStr))
+			encodedStr := base32.StdEncoding.EncodeToString(checkSum[:])
+			suffix := string([]rune(encodedStr)[0:8])
+			if _, found := hashes[suffix]; found {
+				return errors.E(nil, "unable to get a unique hash, multiple samples with identical contents?")
```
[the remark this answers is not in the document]
hzyi-google: Done.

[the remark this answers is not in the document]
hzyi-google: Done.

internal/gensample/gensample.go
hzyi-google: Done.

internal/gensample/gensample.go
hzyi-google: Done.

internal/gensample/plugin.go
hzyi-google: Updated.

hzyi-google:
PTAL

internal/gensample/gensample.go
```diff
@@ -0,0 +1,757 @@
...
+		fmt.Fprintf(&file, "%s %q\n", imp.Name, imp.Path)
+	}
+	file.WriteString(")\n")
+	file.Write(g.pt.Bytes())
+
+	b := file.Bytes()
+	if gofmt {
+		b2, err := format.Source(b)
+		if err != nil {
+			return nil, errors.E(err, "syntax error, run with -nofmt to find out why?")
```
noahdietz: please remove the question mark from the error message

internal/gensample/gensample.go
```diff
@@ -0,0 +1,757 @@
...
+		g.pt.Printf("fmt.Println(resp)")
+		g.imports[pbinfo.ImportSpec{Path: "fmt"}] = true
+	}
+	return nil
+}
+
+// readSampleConfig loads sample configs from a file.
+func readSampleConfig(decoder *yaml.Decoder, fname string) (schema_v1p2.SampleConfig, error) {
+	var config schema_v1p2.SampleConfig
+	for true {
```
noahdietz: don't need the `true` just do `for{`

internal/gensample/gensample.go
```diff
@@ -0,0 +1,757 @@
...
+	}
+	if len(config.Samples) == 0 {
+		return config, errors.E(nil, "Found no valid sample config in %q", fname)
+	}
+	return config, nil
+}
+
+// prependLines adds prefix to every line in b. A line is defined as a possibly empty run
+// of non-newlines terminated by a newline character.
+// If b doesn't end with a newline, prependLines panics.
```
noahdietz: Can we return an error? I might've mentioned it before but using `panic` in a "library" is an anti-pattern. The caller should be able to handle the error as they see fit. 

Also, why can't a newline be appended in the event that it doesn't end with one?

internal/gensample/gensample.go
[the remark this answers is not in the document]
vchudnov-g: Not my favorite, but if it's a widespread change let's not do it now. But yes to the godoc ;-)

internal/gensample/gensample.go
```diff
@@ -41,12 +41,17 @@ const expectedSampleConfigType = "com.google.api.codegen.samplegen.v1p2.SampleCo
 const expectedSampleConfigVersion = "1.2.0"
 
 // InitGen creates a new sample generator.
+// sampleFnames is the filenames of all sample config files.
+// gapicFname is the filename of gapic config.
+// clientPkg is the Go package of the generated gapic client library.
+// nofmt set to true will instruct the generator not to format the generated code. This could be useful for debugging purpose.
```
vchudnov-g: nit: s/purpose/purposes/

internal/gensample/gensample.go
```diff
@@ -56,41 +61,19 @@ func InitGen(desc []*descriptor.FileDescriptorProto, sampleFnames []string, gapi
 	}
 
 	var wg sync.WaitGroup
-	errChan := make(chan error)
+	errChan := make(chan error, 2)
```
vchudnov-g: Uhm, where do you read from `errChan`?  It seems like you're always returning `err==nil` below

internal/gensample/gensample.go
```diff
@@ -140,19 +123,29 @@ func (gen *generator) GenMethodSamples() error {
 }
 
 type generator struct {
+	// desc is the proto descriptors of the API
```
vchudnov-g: nit: "...is the set of..."

internal/gensample/gensample.go
```diff
@@ -170,15 +163,40 @@ func (gen *generator) readSampleConfigFiles(paths []string) error {
...
+
 func (g *generator) reset() {
 	g.pt.Reset()
 	for imp := range g.imports {
 		delete(g.imports, imp)
 	}
 }
 
+// commit writes the internal data representation of the generator to bytes.
+// What it does can be summarized as:
```
vchudnov-g: nit: suggest replacing this line with "It does the following:"

internal/gensample/gensample.go
```diff
@@ -504,6 +523,7 @@ func (g *generator) getInitInfo(inType pbinfo.ProtoType, methConf GAPICMethod, f
 	return initInfo, nil
 }
 
+// argListStr returns a comma-separated string of the list of arguments that sample function takes.
```
vchudnov-g: nit: s/that/that the/

internal/gensample/plugin.go
```diff
@@ -27,7 +27,11 @@ const (
 	paramError = "need parameter in format: go-gapic-package=client/import/path;packageName"
 )
 
-// PluginEntry is the entry point of SampleGen as a protoc plugin, together with CodeGen
+// PluginEntry is the entry point of SampleGen as a protoc plugin.
+// When running gapic-generator-go as a docker image or a protoc plugin, both gapic clients
```
vchudnov-g: The explanation is very helpful. Thanks!

However, I am uncomfortable including details here of this function's callers. It seems like this raises the probability this comment will go stale.

How about saying here that if a protoc plugin is invoked that wishes to generate samples, it will eventually call this function to do so?

And then move the actual specifics you currently have (the configuration presence as a signal, the `main()` filename) to the main plugin source file and/or a README or other doc that summarizes the code layout.

hzyi-google: Done. I'll update readme separately.

internal/gensample/gensample.go
hzyi-google: This function is mostly implemented by pongad@ but my guess is:

> why it panics instead of returning an error:

This is a pure internal helper function. The input of the function is always constructed by other functions in this program, independent of customer inputs. And therefore any error produced by this function is an internal error (someone screwed up and had a bug), and users can do nothing about it.

> why can't a newline be appended in the event that it doesn't end with one?

There isn't such a use case. This function is only used to prepend `//` to convert some code into comments, or prepend whitespace for indentation. The chunk of code to be prepended has to end with a newline, because otherwise following code will continue at the end of the last line of the code prepended, which is always wrong. We could force this function to append a newline character if it did not see one at the end of the buffer, but that might potentially result in more confusion in some corner cases in the future.

Anyway, I'd be happy to convert the panic into returning an error, but automatically adding a newline is a bit unnecessary for now IMHO. WDYT?

noahdietz: Thanks for the detailed breakdown!

> and users can do nothing about it.

I'd say this is a reason _not_ to panic. If the calling code (i.e. the plugin) panics, that's different. I would be quite frustrated as a user if doing a normal invocation the program just crashes, especially without much context. I'd look to the calling functions to wrap the error in the call's context for reporting. Also it is harder to test edge cases when the code crashes.

>  I'd be happy to convert the panic into returning an error

Doesn't have to happen in this PR, but yes lets refactor this to return an error

internal/gensample/gensample.go
hzyi-google: Duh I was dumb! Added.

internal/gensample/gensample.go
hzyi-google: Just used three times altogether, I'll fix them here :)

hzyi-google:
PTAL

internal/gensample/gensample.go
```diff
@@ -0,0 +1,769 @@
...
+	go func() {
+		defer wg.Done()
+		errChan <- gen.readSampleConfigFiles(sampleFnames)
+	}()
+	wg.Wait()
+
+	close(errChan)
+	for err := range errChan {
+		if err != nil {
+			return nil, err
```
vchudnov-g: Do you want to accumulate all the errors so you report them at once?

hzyi-google: I think that is the best way because all errors are reported, but I do not know how to do that. Can you give me an example?

PR #219, opened 2019-10-02
gapic: add release-level option

noahdietz:
* adds `release-level` plugin option
* generates release level documentation for `alpha` and `beta` based on [this snippet](https://github.com/googleapis/gapic-generator/blob/61977cce087339396a29cf5889c70cfac7ca9d60/src/main/resources/com/google/api/codegen/go/doc.snip#L102-L110)
* update README


noahdietz:
cc @jadekler for generated output review

PR #231, opened 2019-10-10
Add build tag for samples

hzyi-google:
Add `// +build sample` build tag to samples, so that running `go install ./...` from client library rood directory will ignore the samples. We will have sample-tester to test the samples so it is safe to do that.


hzyi-google:
cc/ @googleapis/samplegen

internal/gensample/gensample.go
```diff
@@ -233,6 +233,7 @@ func (g *generator) commit(gofmt bool, year int) ([]byte, error) {
 
 	var file bytes.Buffer
 	fmt.Fprintf(&file, license.Apache, year)
+	fmt.Fprintf(&file, "// +build sample\n\n")
```
noahdietz: By convention, build constraints are typically placed at the top of the file on the first line with a new line following. I know the docs don't say this specifically, but it is the common convention and what you will find in google-cloud-go

hzyi-google: Got it. Will update it shortly.

hzyi-google:
I don't think so. I tried locally and it worked. I could have missed something for sure, but I think we can solve issues when we get one.

PR #254, opened 2019-11-26
gapic: fix parsing nested request params

noahdietz:
Nested field names for header request params were not being recognized by the parsing regex. This fixes the regex and updates the test.

### DIFF
--- a/internal/gengapic/gengapic.go
+++ b/internal/gengapic/gengapic.go
@@ -49,7 +49,7 @@ const (
 	beta       = "beta"
 )
 
-var headerParamRegexp = regexp.MustCompile(`{([_\.a-z]+)=`)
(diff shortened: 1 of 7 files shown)


internal/gengapic/gengapic.go
```diff
@@ -49,7 +49,7 @@ const (
 	beta       = "beta"
 )
 
-var headerParamRegexp = regexp.MustCompile(`{([_a-z]+)=`)
+var headerParamRegexp = regexp.MustCompile(`{([_\.a-z]+)=`)
```
lukesneeringer:
```suggestion
var headerParamRegexp = regexp.MustCompile(`{([_.a-z]+)=`)
```

Issue #236, opened 2019-10-15
chore: CI should test with bazel


noahdietz:
In order to validate that the bazel build still works, CI should also test via `bazel test...`

vam-google:
@noahdietz Which CI are you planning to use (CircleCI, Kokoro, something else)?

noahdietz:
@vam-google we use CircleCI in this repo.

We should discuss this, possibly with other microgen authors adding bazel rules. I'd like to not only run `bazel test //...` but also build a GAPIC to test that the generation rule works. 

One idea is to add these rules to Showcase, pull that repo as part of CI and attempt to generate micro-GAPICs via bazel from there as the smoketest. Just an idea.

PR #256, opened 2019-11-27
bazel: add service_yaml attribute

noahdietz:
* update gapic-generator dependency
* update README with optional attributes

Tested locally (going to add bazel testing to CI per #236).

### DIFF
--- a/rules_go_gapic/go_gapic.bzl
+++ b/rules_go_gapic/go_gapic.bzl
@@ -15,14 +15,24 @@
 load("@io_bazel_rules_go//go:def.bzl", "go_library")
 load("@com_google_api_codegen//rules_gapic:gapic.bzl", "proto_custom_library", "unzipped_srcjar")
 
-def go_gapic_library(name, srcs, importpath, deps, release_level="", grpc_service_config = None, **kwargs):
+def go_gapic_library(
+  name,
+  srcs,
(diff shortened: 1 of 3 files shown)


rules_go_gapic/go_gapic.bzl
```diff
@@ -15,14 +15,24 @@
...
   file_args = {}
 
   if grpc_service_config:
-    file_args = {
-      grpc_service_config: "grpc-service-config",
-    }
+    file_args[grpc_service_config] =  "grpc-service-config"
+
+  if service_yaml:
+    file_args[service_yaml] = "gapic-service-config"
```
vam-google: Service config yaml is not gapic-specific (i.e. makes sense and existed before gapic became a thing). Is there a reason why "gapic-" prefix is used here?

noahdietz: Yeah, the thought was that if other microgenerators needed the service yaml, they'd all key in on the name `gapic-service-config`, the "gapic-" prefix making it language agnostic (as opposed to `go-gapic-package` for example which should not be used for a python package). Luke and I settled on it, but this option is going to be removed once I figure out how to get what I need in it (some doc strings) from another source, so I'm not too worried about the name. It is also already being used though, so if possible I'd like not to change it.

PR #258, opened 2019-12-03
gapic: fix lro handling Empty

noahdietz:
The monolith generated `Wait` methods with an `error` return type when `google.protobuf.Empty` was the `response_type`. To avoid a breaking change, this adopts that approach.

@jadekler


noahdietz:
LGTM'd offline

PR #259, opened 2019-12-03
gapic: fix lro wait returns

noahdietz:
#258 neglected to update the corresponding `return` statements based on the `response_type`. This fixes that.

It also neglected to fix the example files, which would fail to compile. This fixes that.

It also updates all of the relevant tests, adding more for the special cases.

### DIFF
--- a/internal/gengapic/gengapic.go
+++ b/internal/gengapic/gengapic.go
@@ -45,6 +45,8 @@ const (
 	emptyType = ".google.protobuf.Empty"
 	lroType   = ".google.longrunning.Operation"
 	// used when google.protobuf.Empty is the value of an annotation, which isn't resolved by protoc
+	//
+	// TODO(ndietz): https://github.com/googleapis/gapic-generator-go/issues/260
 	emptyValue = "google.protobuf.Empty"
 	paramError = "need parameter in format: go-gapic-package=client/import/path;packageName"
 	alpha      = "alpha"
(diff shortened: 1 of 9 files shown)


internal/gengapic/gengapic.go
```diff
@@ -42,8 +42,10 @@ import (
 
 const (
 	// protoc puts a dot in front of name, signaling that the name is fully qualified.
-	emptyType  = ".google.protobuf.Empty"
-	lroType    = ".google.longrunning.Operation"
+	emptyType = ".google.protobuf.Empty"
+	lroType   = ".google.longrunning.Operation"
+	// used when google.protobuf.Empty is the value of an annotation, which isn't resolved by protoc
+	emptyValue = "google.protobuf.Empty"
```
jadekler: Kind of a bummer to have two of the same import. Is there a way to tidy that?

noahdietz: Yeah it is a bummer. Unfortunately, the preceding `.` has semantic meaning, it means protoc has identified it as a fully resolved name, which influences some code paths in the generator. However, when a user sets an annotation, which can be just a string value, to `"google.protobuf.Empty"`, this isn't mutated by protoc but influences different code paths in the generator. 

I will file a bug to look at refactoring it, but I'd prefer not to do it in this CL.

noahdietz: Added a tracking bug comment.

PR #273, opened 2019-12-13
chore: add bazel test to CI

### LINKED ISSUE #236: chore: CI should test with bazel

noahdietz:
Runs `bazel test` as part of CI

Fixes #236


.circleci/config.yml
```diff
@@ -61,6 +65,27 @@ jobs:
             mv googleapis-master googleapis
             export PATH=$PATH:$(pwd)/protobuf/bin
             GOOGLEAPIS=googleapis OUT=$GOPATH/src ./test.sh
+  bazel-test:
+    environment:
+      BAZEL_VERSION: 1.2.1
```
vam-google: Please target bazel version 2+ (it is the newest verson, googleapis CI is build using this bazel version)

.circleci/config.yml
```diff
@@ -61,6 +65,27 @@ jobs:
...
+      PYTHON_VERSION: 3.5.2
```
vam-google: 3.5 is generally ok, but our workstations on 3.7. Please consider changing this to 3.7 to minimize number of differences (monolyth gapic-generator CI still uses these versions, but it has to be updated as well, I just haven't a chance to make that small PR).

noahdietz: Done!

.circleci/config.yml
noahdietz: Done!

PR #311, opened 2020-02-14
gapic: fix spelling of canceled

noahdietz:
Go uses the American-English spelling `CANCELED`, while the gRPC code name uses the British-English spelling `CANCELLED`.

### DIFF
--- a/internal/gengapic/client_init.go
+++ b/internal/gengapic/client_init.go
@@ -24,6 +24,7 @@ import (
 	conf "github.com/googleapis/gapic-generator-go/internal/grpc_service_config"
 	"github.com/googleapis/gapic-generator-go/internal/pbinfo"
 	"google.golang.org/genproto/googleapis/api/annotations"
+	"google.golang.org/genproto/googleapis/rpc/code"
 )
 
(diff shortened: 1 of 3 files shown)


internal/gengapic/client_init.go
```diff
@@ -160,7 +160,13 @@ func (g *generator) clientOptions(serv *descriptor.ServiceDescriptorProto, servN
 				p("gax.WithRetry(func() gax.Retryer {")
 				p("  return gax.OnCodes([]codes.Code{")
 				for _, c := range rp.GetRetryableStatusCodes() {
-					p("    codes.%s,", snakeToCamel(c.String()))
+					// Go uses the American-English spelling with a single "L"
+					code := c.String()
+					if code == "CANCELLED" {
+						code = "CANCELED"
```
alexander-fenster: Just wondering, can we use that `grpc.status` enum directly without having a string constant?  In JavaScript, we can:

```
> grpc.status.CANCELLED
1
```

noahdietz: Great idea! Done.

PR #314, opened 2020-02-20
chore: update bazel repositories.bzl

noahdietz:
Update of Go deps based on `go.mod` using some manual editing after:
`bazel run //:gazelle -- update-repos -from_file=go.mod`


noahdietz:
This change demonstrates how dependency on `google.golang.org/genproto` brings in all of these grpc dependencies. Also it changes super frequently.

We really need to get api-common-protos on an automatic update and a separate api-common-go repo with *just* proto bindings for the annotations (no grpc) or something  of the sort to depend on.

PR #337, opened 2020-03-23
Use jsonpb to marshal json message output

### LINKED ISSUE #336: Json output is not jsonpb formatted, so is incompatible with from_file input
The output messages printed using the `--json` flag use raw json encoding, whereas the parser used when providing a json file via `--from_file` uses a jsonpb to unmarshal.
Usually this works fine except for well known messages such as [Duration](https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/duration.proto#L92)  which should encode into a json string instead of just the object, leading to an output which doesn't work as an input.

Eg for a type foo with a duration field, `gapicclient get-foo --name=foo1 > foo.json` followed by `gapic update-foo --from_file=foo.json` will fail with `json: cannot unmarshal object into Go value of type string`.

PR to follow.

tprice11:
This means outputs can also be used as inputs via the --from_file flag.

Fixes #336.

### DIFF
--- a/internal/gencli/root_file.go
+++ b/internal/gencli/root_file.go
@@ -26,8 +26,8 @@ const (
 package main
 
 import (
+	"bytes"
 	"context"
-	"encoding/json"
 	"fmt"
 	"os"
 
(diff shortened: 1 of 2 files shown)


internal/gencli/root_file.go
```diff
@@ -72,11 +72,14 @@ func printVerboseInput(srv, mthd string, data interface{}) {
...
-		d, _ := json.MarshalIndent(data, "", "  ")
-		s = string(d)
-	} else if msg, ok := data.(proto.Message); ok {
-		s = msg.String()
+	if msg, ok := data.(proto.Message); ok {
+		if OutputJSON {
+			var b bytes.Buffer
+			marshaler.Marshal(&b, msg)
+			s = b.String()
+		} else {
```
noahdietz: Can you make the `else` case the "default" and move it above the `if` statement?  I like to reduce indentation and avoid `else` statements when possible.

```go
s = msg.String()
if OutputJSON {
    var b bytes.Buffer
    marshaler.Marshal(&b, msg)
    s = b.String()
}
```

tprice11: Sounds good! Done.

tprice11:
Thanks for the super fast reviews, and for building this out! We're excited to integrate it into our development practices!

noahdietz:
I'm glad you're using it. Is your project OSS? It doesn't handle every API perfectly, but does a decent job with most...and thanks for the PR.

Give me a little while to release the change, I want to see if there are other changes I can include.

PR #342, opened 2020-03-26
gapic: tighten reference link parser regexp

noahdietz:
I needed to tighten the regexp added in #341. It was capturing too much, incorrectly.

### DIFF
--- a/internal/gengapic/markdown.go
+++ b/internal/gengapic/markdown.go
@@ -27,7 +27,7 @@ import (
 )
 
 var linkParser = regexp.MustCompile(`<a href=["'](.+)["']`)
-var referenceParser = regexp.MustCompile(`\[(.+)\]\[(.*)\]`)
+var referenceParser = regexp.MustCompile(`\[([a-zA-Z1-9\.\_]+)\]\[[a-zA-Z1-9\.\_]*\]`)
 
 func MDPlain(s string) string {
 	var mdr mdRenderer
--- a/internal/gengapic/markdown.go
+++ b/internal/gengapic/markdown.go
@@ -27,7 +27,7 @@ import (
 )
 
 var linkParser = regexp.MustCompile(`<a href=["'](.+)["']`)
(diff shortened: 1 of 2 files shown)


internal/gengapic/markdown.go
```diff
@@ -27,7 +27,7 @@ import (
 )
 
 var linkParser = regexp.MustCompile(`<a href=["'](.+)["']`)
-var referenceParser = regexp.MustCompile(`\[(.+)\]\[(.*)\]`)
+var referenceParser = regexp.MustCompile(`\[([a-zA-Z1-9\.\_]+)\]\[[a-zA-Z1-9\.\_]*\]`)
```
tbpg: I don't think you need to escape the period or underscore inside a character class (`[...]`).

```suggestion
var referenceParser = regexp.MustCompile(`\[([a-zA-Z1-9._]+)\]\[[a-zA-Z1-9._]*\]`)
```

internal/gengapic/markdown_test.go
```diff
@@ -74,8 +74,8 @@ func TestMDPlain(t *testing.T) {
 			want: "html with a softbreak (at /link/to/some#thing) \n test",
 		},
 		{
-			in:   "link to [a search engine](https://www.google.com) with request type [Search][foo.bar.v1.Search].",
-			want: "link to a search engine (at https://www.google.com) with request type Search.",
+			in:   "link to [a search engine](https://www.google.com) with request type [Search][foo.bar_baz.v1.Search], [biz][].",
```
tbpg: Would you mind adding another test case for something that was matched before this change, but isn't matched now?

internal/gengapic/markdown.go
noahdietz: Done! Thanks, good catch.

internal/gengapic/markdown_test.go
noahdietz: Done! I expanded it some more. But basically, if there were multiple references in a block, the previous regex would match the link text to keep as everything between the first square bracket and the closing bracket of the last set, keeping essentially everything.

PR #343, opened 2020-03-27
chore: automate release notes generation

### LINKED ISSUE #338: chore: automate release notes creation
A simple tag push results in automation creating a GitHub Release for the newly pushed tag. But there are no release notes and the Release title is not ideal (the entire message of the latest commit). 

We should make release notes generated into the release based on the commit messages of those commits since the last release.

noahdietz:
Adds functionality to the release tool that collects the commit messages since the previous tag, aggregates dependency updates into one message, and includes formatted release notes when creating a new release.

Fixes #338


util/cmd/release/main.go
```diff
@@ -109,9 +109,98 @@ func main() {
...
 }
+
+// notes collects the commit messages since the previous tag, aggregates
+// dependency updates into one message, and returns formatted release notes.
+func notes(version string) string {
+	previous := util.ExecuteWithOutput(
+		"git",
+		"describe",
+		"--abbrev=0",
+		"--tags",
```
tbpg: Does this need `--no-pager`?

util/cmd/release/main.go
```diff
@@ -109,9 +109,98 @@ func main() {
...
+		version+"^",
+	)
+	previous = strings.TrimSpace(previous)
+
+	output := util.ExecuteWithOutput(
+		"git",
+		"log",
+		previous+"..HEAD",
+		"--oneline",
+		"--pretty=format:%s",
```
tbpg: Same comment.

util/cmd/release/main.go
```diff
@@ -109,9 +109,98 @@ func main() {
...
+		case "gencli":
+			gencli = append(gencli, "*"+split[1])
+		case "chore(deps)":
+			hasDeps = true
+		case "chore":
+			chore = append(chore, "*"+split[1])
+		case "samples":
+			samples = append(samples, "*"+split[1])
+		default:
+			other = append(other, "* "+msg)
```
tbpg: Why a space after this `*` and not the others?

util/execute.go
```diff
@@ -25,3 +25,12 @@ func Execute(args ...string) {
 		log.Fatalf("%s", output)
 	}
 }
+
+// ExecuteWithOutput runs the given strings as a command and returns the output.
+func ExecuteWithOutput(args ...string) string {
```
tbpg: Would love a better spot for this than `package util`. "Avoid meaningless package names. Packages named util, common, or misc provide clients with no sense of what the package contains." --https://blog.golang.org/package-names

If these are only used in main.go, we could just move them there.

tbpg: Would propose following the pattern of `exec.Command`: make the first string argument required. Makes this harder to misuse and the same level of easiness to use correctly.

util/cmd/release/main.go
```diff
@@ -109,9 +109,98 @@ func main() {
...
+
+	output := util.ExecuteWithOutput(
+		"git",
+		"log",
+		previous+"..HEAD",
+		"--oneline",
+		"--pretty=format:%s",
+	)
+
+	var hasDeps bool
```
codyoss: Could this be broken out into a function and tested so it is easier to understand what the release notes would look like.

util/cmd/release/main.go
noahdietz: Nope, this returns a single tag, just the tag preceding `version`.

For example, in this repo:
```
$ git describe --abbrev=0 --tags v0.12.5^
v0.12.4
```

Number of `^` used indicates how many to jump back to.
```
$ git describe --abbrev=0 --tags v0.12.5^^
v0.12.3
```

At least, that is what I've seen in local testing.

util/cmd/release/main.go
noahdietz: There doesn't seem to be a `--no-pager` flag for the `log` cmd ([docs](https://git-scm.com/docs/git-log)). But in my local testing, there hasn't been a problem for the script with longer output.

util/cmd/release/main.go
noahdietz: Hm this might be a brittle assumption. But it's based on the assumption that one of the known components will be formatted with a space following the colon `gapic: fix some stuff`. That space shows up in `split[1]` as the first char. 

However, with unknown components or just malformed messages that are put in "other", the entire commit message is used (it's not split on the colon), so a space is added to separate the full `msg` from the `*`. 

Do you think I should strip white space on the `split` components and manually add the space? For example `samples = append(samples, "* " + strings.TrimSpace(split[1]))`.

util/execute.go
noahdietz: Ah yes, it is only used in `cmd/release` so I will move it there. I copied a lot of this stuff from another repo that had more utilities depending on `execute.go`. Didn't think twice about it when I moved it over.

noahdietz: Good idea. Will do.

util/cmd/release/main.go
noahdietz: Good idea. Will do!

util/cmd/release/main.go
tbpg: I think so. Minor change, but it will help in the accidental case.

util/cmd/release/main_test.go
```diff
@@ -0,0 +1,63 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"testing"
+)
+
+func Test_format(t *testing.T) {
```
tbpg:
```suggestion
func TestFormat(t *testing.T) {
```

I don't know why it's a pattern for integration tests. I guess for easy separation from other tests?

util/cmd/release/main_test.go
```diff
@@ -0,0 +1,63 @@
...
+* update dependencies (see history)
+
+# other
+
+* unknown: foo bar baz
+* malformed message`
+
+	got := format(in)
+	if got != want {
+		t.Errorf("format(%q)=%q, want %q", in, got, want)
```
tbpg: Might want to use `if diff := cmp.Diff(got, want); diff != ""` to make it easier to find the wrong part if this fails.

noahdietz: Done.

util/cmd/release/main_test.go
noahdietz: Done!

util/cmd/release/main.go
noahdietz: Done! Did it with a little less repeated code though :)

util/cmd/release/main.go
```diff
@@ -94,24 +93,143 @@ func main() {
...
+	commits := strings.Split(output, "\n")
+
+	return format(commits)
+}
+
+// format parses the given commits messages and formats them into release notes
+// organized by component.
+func format(commits []string) string {
+	var hasDeps bool
+	var gapic, bazel, gencli, chore, samples, other []string
```
tbpg: Sorry, I just thought of this, so feel free to ignore it.

What if we did a `map[string][]string`? Then we wouldn't need to duplicate all of the `append` and `WriteString` calls. Need to be careful with sorting and preventing other categories. So, it might end up being more complicated.

noahdietz: That is a great idea, but I like the simplicity of the current approach so I think I'll stick with that :)

PR #356, opened 2020-04-07
gencli: JSON output of paged responsed

ericwenn:
Paged responses does not implemented `proto.Message` and the output was not JSON encoded.

### DIFF
--- a/internal/gencli/cmd_file.go
+++ b/internal/gencli/cmd_file.go
@@ -276,32 +276,16 @@ var {{$methodCmdVar}} = &cobra.Command{
 		{{ end }}
 		{{ end }}
 		{{ else if and .Paged (not .IsLRO ) }}
-		// get requested page
-		var items []interface{}
-		data := make(map[string]interface{})
-
-		// PageSize could be an integer with a specific precision.
-		// Doing standard i := 0; i < PageSize; i++ creates i as
-		// an int, creating a potential type mismatch. 
-		for i := {{ .InputMessageVar }}.PageSize; i > 0; i-- {
-			item, err := iter.Next()
-			if err == iterator.Done {
-				err = nil
-				break
-			} else if err != nil {
-				return err
-			}
-
-			items = append(items, item)
+		// populate iterator with a page
+		_, err = iter.Next()
+		if err != nil && err != iterator.Done {
+			return err
 		}
-
-		data["page"] = items
-		data["nextToken"] = iter.PageInfo().Token
 
 		if Verbose {
 			fmt.Print("Output: ")
 		}
-		printMessage(data)
+		printMessage(iter.Response)
 		{{ else if not .IsLRO }}
 		if Verbose {
 			fmt.Print("Output: ")
--- a/internal/gencli/root_file.go
+++ b/internal/gencli/root_file.go
@@ -79,8 +79,6 @@ func printMessage(data interface{}) {
 			marshaler.Marshal(&b, msg)
 			s = b.String()
 		}
-	} else if page, ok := data.(map[string]interface{}); ok {
-		s = fmt.Sprintf("%v", page)
 	}
 
 	fmt.Println(s)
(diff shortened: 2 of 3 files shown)


ericwenn:
@googlebot I signed it!

odsod:
Nice one @ericwenn !

noahdietz:
Thanks @ericwenn for catching that! I wonder if there is a way we can make this align with the rest of the `jsonpb` output that was incorporated by @tprice11 in #337.

The generated Iterator type contains a `Response interface{}` field ([example](https://github.com/googleapis/google-cloud-go/blob/master/translate/apiv3/translation_client.go#L594-L597)). It is the raw response message of the current page, updated when the page needs more items and makes a RPC to fetch the next page. 

I wonder if we should remove the `PageSize`-based loop that invokes `iter.Next()` to collect the items, and just print the `iter.Response` instead - `printMessage(iter.Response)`. It contains the page of items and the `next_page_token` so it is just as good.

We could replace this entire [chunk of the template](https://github.com/googleapis/gapic-generator-go/blob/master/internal/gencli/cmd_file.go#L280-L304) with the following:

```go
// populate iterator with a page
_, err := iter.Next()
if err != nil && err != iterator.Done {
  return err
}

if Verbose {
  fmt.Print("Output: ")
}
printMessage(iter.Response)
```

`printMessage` properly casts the given `iter.Response` to a `proto.Message` and prints with `jsonpb`.

I _think_ that should work.

WDYT?

ericwenn:
Sounds good to me! 
TBH I'm not overly familiar with the generated iterators/client's and don't know if using the response directly (instead of iterator) messes with some other use-case. If you feel comfortable with it, I'll go ahead and update the PR.

ericwenn:
Updated based on your suggestion @noahdietz. Removed extra if-case in printMessage for paged responses.

internal/gencli/cmd_file.go
```diff
@@ -276,32 +276,15 @@ var {{$methodCmdVar}} = &cobra.Command{
...
-			if err == iterator.Done {
-				err = nil
-				break
-			} else if err != nil {
-				return err
-			}
-
-			items = append(items, item)
+		// populate iterator with a page
+		_, iterErr := iter.Next()
```
noahdietz, line 280: Can we change this to `_, err = iter.Next()` rather than creating a new error variable?

internal/gencli/cmd_file.go
```diff
@@ -276,32 +276,15 @@ var {{$methodCmdVar}} = &cobra.Command{
...
+		if iterErr != nil && iterErr != iterator.Done {
+			return iterErr
 		}
-
```
noahdietz, line 297: Please keep the new line between the error check closing brace and the beginning if the `if Verbose {`

internal/gencli/cmd_file.go
ericwenn: Wasn't sure if `err` was always declared earlier, hence `listErr` to be safe. Could change to `err = ` instead

noahdietz:
Works fine with our [gapic-showcase](https://github.com/googleapis/gapic-showcase) CLI. Great stuff.

internal/gencli/cmd_file.go
noahdietz: `err` is a [named return value](https://github.com/googleapis/gapic-generator-go/blob/master/internal/gencli/cmd_file.go#L116) of the encapsulating function. So we are good to use it here!

internal/gencli/cmd_file.go
ericwenn: Fixed

ericwenn:
Works with my projects as well!

noahdietz:
Released in https://github.com/googleapis/gapic-generator-go/releases/tag/v0.13.1

ericwenn:
@noahdietz Thanks!

PR #359, opened 2020-04-09
bazel: fix assembly_pkg archive contents path

noahdietz:
Today, `go_gapic_assembly_pkg` produces an invalid archive. The proper files are in the archive, but the file paths are prefixed with `../`. This makes extraction fail. 

Also, the `chmod` for the GAPIC code fails because it attempts to use the raw importpath, which includes a trailing `;package`. There isn't such a directory, so this strips that suffix, resulting in the proper directory being targeted.

### DIFF
--- a/rules_go_gapic/go_gapic_pkg.bzl
+++ b/rules_go_gapic/go_gapic_pkg.bzl
@@ -32,7 +32,7 @@ def _go_gapic_src_pkg_impl(ctx):
                 if dep_file.extension == "srcjar":
                     srcjars.append(dep_file)
 
-    paths = construct_package_dir_paths(ctx.attr.package_dir, ctx.outputs.pkg, ctx.label.name)
+    paths = construct_package_dir_paths(None, ctx.outputs.pkg, ctx.label.name)
     script = """
     for srcjar in {srcjars}; do
         mkdir -p {package_dir_path}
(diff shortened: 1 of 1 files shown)


rules_go_gapic/go_gapic_pkg.bzl
```diff
@@ -32,7 +32,7 @@ def _go_gapic_src_pkg_impl(ctx):
                 if dep_file.extension == "srcjar":
                     srcjars.append(dep_file)
 
-    paths = construct_package_dir_paths(ctx.attr.package_dir, ctx.outputs.pkg, ctx.label.name)
+    paths = construct_package_dir_paths(None, ctx.outputs.pkg, ctx.label.name)
```
vam-google, line 35: Does the `../` prefix issue happen on Mac? We had discovered same issue debuggin PHP stuff, but it is reproducible only on mac (probably due to some differences in tar implementation between linux and mac platforms). The archive is not broke-broken, it is "damaged", but it still can be extracted by adding `-P` command line arg to `tar` and then checking the output one level above from the dir where tar is executed (because of `../` prefix inside tar).

If it is mac-specific, then we can use this fix to fix other targerts as well.

noahdietz: I just built a go_assembly_pkg on my workstation without any of these changes. It looks like `tar` removes the `../` prefixes:

```
tar: Removing leading `../' from member names
tar: Removing leading `../' from hard link targets
```

And a `tar -tf` shows a clean set:
```
gapi-cloud-language-v1-go/cloud.google.com/
gapi-cloud-language-v1-go/cloud.google.com/go/
gapi-cloud-language-v1-go/cloud.google.com/go/language/
gapi-cloud-language-v1-go/cloud.google.com/go/language/apiv1/
gapi-cloud-language-v1-go/cloud.google.com/go/language/apiv1/language_client_example_test.go
gapi-cloud-language-v1-go/cloud.google.com/go/language/apiv1/language_client.go
gapi-cloud-language-v1-go/cloud.google.com/go/language/apiv1/doc.go
...
```

I also tried it _with_ this change and it worked fine as well. The `./` prefix was present, but that's OK.

So this seems like a mac-specific issue/fix, but it worked fine on my workstation too.

PR #362, opened 2020-04-14
chore: change plugin use of Unmarshal

noahdietz:
This changes the use of `Unmarshal` from the `CodeGeneratorRequest` to the general `proto.Unmarshal`. The former is removed in golang/protobuf v1.4.0 and it is breaking builds that are pulling this tool and latest golang/protobuf.


noahdietz:
> Does seem rather odd that this got removed in a minor release... Have you reached out to that team?

I haven't, I think I will, maybe I missed something though.

I don't mind doing the change though. Thanks for the review.

PR #363, opened 2020-04-14
bazel: use rules_go provided protobuf plugin dep

noahdietz:
Changes plugin `go_library` targets to use `rules_go`-provided protobuf plugin dependency. This gets ride of import shadowing warnings.

Also results from `bazel run //:gazelle`:
* adds the `util/cmd/release` and `util/cmd/license` packages to Bazel gazelle ignore list
* adds a `go_test` target to  internal/gensample/schema 
* removes grpc plugin form grpc_service_config go_proto_library target as it is unnecessary
* refactor some dep lists


noahdietz:
> Also this will transfer the "version ownership" from us to the rules_go and make it less flexible for us

Mmm good point. And as it stands, [rules_go is behind](https://github.com/bazelbuild/rules_go/blob/b64542950a81e86c37e6bc5335060b70c251cf5e/go/private/repositories.bzl#L129) on the version. Is it possible for me to provide the version of golang/protobuf with my own http_archive call that precedes ruels_go in my workspace? I'm afraid that if I do that, the patches executed by rules_go won't be executed and I won't be able to use the dependency. I just wanted to align this dependency with the "well known types" in rules_go...

> potentially inconsistent with the other versions for other langauges.

This is for github.com/golang/protobuf, the Go implementation of protobuf & the plugin.pb.go interface, not the github.com/protocolbuffers/protobuf repo that has every other language in it. Go is not included in that repo. So it already is different from the other languages in a sense.

noahdietz:
Actually, I'm not sure that this is a problem, because we declare our own dependency on github.com/golang/protobuf in the [repositories.bzl](https://github.com/googleapis/gapic-generator-go/blob/master/repositories.bzl#L82-L87).

The [commentary in rules_go](https://github.com/bazelbuild/rules_go/blob/b64542950a81e86c37e6bc5335060b70c251cf5e/go/private/repositories.bzl#L108-L111) indicates users should specify their own dep. Maybe what we are already doing is sufficient? WDYT?

PR #379, opened 2020-05-04
gapic: flag support for proto3_optional

noahdietz:
* add SupportedFeatures flag to CodeGeneratorResponse
* update showcase integation go.mod
* add proto3_optional enablement flag to showcase script
* update repositories.bzl from go.mod
* upgrade version of protoc in CI

**Note:** Bazel build breaks because rules_go forces the golang/protobuf version to v1.3.3, which doesn't have the API necessary for the plugin to flag proto3_optional support. So, it has been [overridden](https://github.com/bazelbuild/rules_go/blob/master/go/dependencies.rst#overriding-dependencies) using the patches found on HEAD of rules_go (meant for golang/protobuf v1.4.0, but they work for v1.4.1 as well). This will be reverted once rules_go releases with the sufficient dependency versions included. The use of latest rules_go also meant taking its new latest minimum bazel version - 2.2.0.


.circleci/config.yml
```diff
@@ -54,7 +54,7 @@ jobs:
           name: Run Showcase
           command: |
             mkdir protobuf
-            curl -sSL https://github.com/protocolbuffers/protobuf/releases/download/v3.11.4/protoc-3.11.4-linux-x86_64.zip > protobuf/protoc.zip
+            curl -sSL https://github.com/protocolbuffers/protobuf/releases/download/v3.12.0-rc1/protoc-3.12.0-rc-1-linux-x86_64.zip > protobuf/protoc.zip
```
vam-google, line 57: Please add TODOs here and in other places, which either depend on `rc` verisons of dependencies or result of "hacks" (like overriding protobuf from `1.3.3` to `1.4.0`) briefly explaining when and it should be replaced with proper GA versions.

noahdietz: Done. Good idea, thanks.

noahdietz:
> Maybe just wait a bit until the proper dependencies are released

As I was doing this I contacted the rules_go maintainer and he said the golang protobuf deps would be updated in the next release v0.23.0 sometime next week. I'm fine with the churn here to unblock gapic-showcase development and have green CI.

As far as protobuf version, the golang protobuf deps released versions based on it, so I don't mind doing so as well. But I do plan on changing it to the stable release as soon as possible.

Issue #411, opened 2020-06-19
gapic: clients should set a default timeout in the generated code based on the generation config (cloud spanner)


hengfengli:
## Issue

We found that the generated code does not have any default timeout settings from config files. 

Config files: 

* https://github.com/googleapis/googleapis/blob/master/google/spanner/v1/spanner_grpc_service_config.json
* https://github.com/googleapis/googleapis/blob/master/google/spanner/admin/database/v1/spanner_admin_database_grpc_service_config.json
* https://github.com/googleapis/googleapis/blob/master/google/spanner/admin/instance/v1/spanner_admin_instance_grpc_service_config.json

Generated code: 

- Spanner client: https://github.com/googleapis/google-cloud-go/blob/master/spanner/apiv1/spanner_client.go#L68
- DB admin client: https://github.com/googleapis/google-cloud-go/blob/master/spanner/admin/database/apiv1/database_admin_client.go#L74
- Instance admin client: https://github.com/googleapis/google-cloud-go/blob/master/spanner/admin/instance/apiv1/instance_admin_client.go#L67

## Actual behaviour

No default timeouts. 

## Expected behaviour

All grpc methods should respect the default timeout settings from the config files. 

## Notes

I had a brief discussion with @noahdietz: "It is possible to explore a solution that if the context does not have any user-specified timeout, we should assign a default timeout. However, this could be perceived as a breaking change for Go GAPICs - where an RPC would happily hang until completion, now it is canceling and returning an error. Users might not be prepared for that."

We expect the timeout/retrying settings should be consistent across all languages. However, in terms of timeout settings, Go client lib is inconsistent. The spanner client lib dev team hopes this can be fixed.

noahdietz:
Hi @hengfengli thanks for opening this so that we can formally discuss it. I would like to clarify that this is not an "issue" (or bug) as stated. It is a design decision based on the common practice of Go users managing their `Context` and deadlines.

Now, supporting this using [`context.WithDeadline()`](https://pkg.go.dev/context?tab=doc#WithDeadline) might be OK, per this part of the docs:

    If the parent's deadline is already earlier than d, WithDeadline(parent, d) is semantically equivalent to parent

So we won't disturb anyone's existing deadline by setting a default one. But, like I said offline, this could be considered a breaking change in behavior for those not managing their `Context` deadline.

What do others think? @broady @tbpg @codyoss

noahdietz:
[...]

PR #422, opened 2020-08-05
gapic: add default method deadlines

noahdietz:
This adds consumption of the `timeout` in the generation config, using it to impose a default `context` deadline on client methods. This is an _overall deadline_ that encompasses all _RPC attempts_ made by the method.

Rough design:
* A `context` deadline set by a user is respected always
* the context derived from adding a deadline is canceled in the following scenarios:
  * in the event of an error in any method on the top-level client ~or a stream-wrapper client~
  * when a Unary method returns (using `defer`)
  * ~when a server-streaming client wrapper `Recv` receives an `io.EOF`~
  * ~when a client-streaming client wrapper `CloseAndRecv`~
  * ~when a bi-directional streaming client wrapper has both ends (client send & recv) of the client closed (via `CloseSend` & `Recv`-with-`io.EOF` respectively)~
* ~streaming methods return a generated wrapper that handles cancelation logic and passes through to the gRPC stub for all other work~

_Note: this will **not** support methods that return an `Iterator` at the outset. There needs to be more discussion around Iterator vs. per-Iteration deadlines._ 

_Update: this will **not** support streaming methods at the outset. There needs to be more discussion around the use case. The corresponding code was removed from this PR._

_Update: there will be an opt-out environment variable that's generated too._

Addresses part of #411


internal/gengapic/testdata/method_BidiThings.want
```diff
@@ -1,15 +1,71 @@
...
+		}
+		resp = &fooBidiThingsClient{
+			Foo_BidiThingsClient: stub,
+			cancel: cancel,
+		}
+		return nil
 	}, opts...)
 	if err != nil {
+		if cancel != nil {
+			cancel()
```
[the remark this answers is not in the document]
codyoss: Are there cases where the context will have not been canceled by this point already?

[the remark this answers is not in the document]
noahdietz: > Should this also be canceled for the happy path?

This same question came up in the doc. This specific `cancel` call is for when `gax` returns an error while executing `Invoke` i.e. during stream creation. The happy path is covered by the follwing stream client wrapper (e.g. `fooBidiThingsClient`) that `cancel`s in the success scenarios described in the description (and in the doc).

Does that help?

internal/gengapic/gengapic.go
```diff
@@ -481,6 +489,34 @@ func (g *generator) emptyUnaryCall(servName string, m *descriptor.MethodDescript
 	return nil
 }
 
+func (g *generator) deadline(s, m string) {
+	t, ok := g.grpcConf.Timeout(s, m)
+	if !ok {
+		return
+	}
+
+	g.printf("if _, set := ctx.Deadline(); !set {")
```
tbpg, line 498: Nit: I think `ok` is usually used as the variable name? Probably doesn't matter.

internal/gengapic/gengapic.go
```diff
@@ -481,6 +489,34 @@ func (g *generator) emptyUnaryCall(servName string, m *descriptor.MethodDescript
...
+	g.printf("  cc, cancel := context.WithTimeout(ctx, %d * time.Millisecond)", t)
```
tbpg, line 499: Nit: Perhaps `cctx`?

noahdietz: Yeah, that's better! Done.

internal/gengapic/gengapic.go
noahdietz: I'll go with `ok` to stick to common practice. Done.

noahdietz:
> Can we do any sort of extra (?) leak detection here? 

Yeah, I agree, we should. I will take a look at adding leak detection to the Showcase-based integration tests. I've never done that before, but I see there might be an example in google-cloud-go "leackcheck"?

noahdietz:
Opened #426 to add basic `leakcheck` to the integration tests. I will rebase this change once that is merged in order to help verify.

noahdietz:
I rebased this PR to include the addition of the `leakcheck` tool in our integration test module.

tbpg:
Thanks! LGTM!

noahdietz:
Rebased the branch to pull in the retry/deadline test suite from #428 and remove the default deadline changes for streaming methods. 

Added the opt-out env var code gen and a test.

Added more default deadline integration tests.

internal/gengapic/gengapic.go
```diff
@@ -47,6 +47,7 @@ const (
 	paramError = "need parameter in format: go-gapic-package=client/import/path;packageName"
 	alpha      = "alpha"
 	beta       = "beta"
+	disableDeadlinesVar = "GOOGLE_API_GO_EXPERIMENTAL_DISABLE_DEFAULT_DEADLINE"
```
tbpg, line 50: Is this a pattern for env vars for this repo? I'm OK with it, just curious.

noahdietz: Pattern for the env var name? or for how I'm implementing it? The name was suggested by cbro.

noahdietz: To be clear, this is the name of the env var that _generated clients_ will load for the opt-out

tbpg: Yeah, I wasn't sure if `GOOGLE_API_GO_EXPERIMENTAL_` was a known, common, prefix for env vars for clients. Just looked and I don't see it anywhere. So, I don't think so.

noahdietz: Yeah, I looked around google-api-go-client and google-cloud-go for any sort of pattern in env var naming that we could follow here...there was none, all kind of adhoc.

PR #448, opened 2020-09-22
gapic: add module generator option

### LINKED ISSUE #447: protoc-gen-go_gapic: support strip-module prefix option
Could you support `module` option like in `protoc-gen-go` and `protoc-gen-go-grpc` plugins.

```
	protoc \
		-I ${API_COMMON_PROTOS} \
		-I ${EXAMPLEAPIS} \
		--go_gapic_out=. \
                --go_gapic_opt=module=github.com/example/repo \
		--go_gapic_opt="go-gapic-package=github.com/example/repo/myapi/apiv1"';myapi' \
		${EXAMPLEAPIS}/myapi/v1/*.proto
```

should generate next structure:

```
.
|_myapi
   |_apiv1
      |_doc.go
      |_example_client.go
      |_example_client_example_test.go
```

but in the same time `example_client_example_test.go` should have next import:

```
import myapi "github.com/example/repo/myapi/apiv1"
```

instead of

```
import myapi "myapi/apiv1"
```

noahdietz:
Adds a generator option to strip the given module prefix from the `go-gapic-package` value when used in the generated filenames. Similar to the [protoc-gen-go option](https://go-review.googlesource.com/c/protobuf/+/219298/). So, if the plugin output directory is the module root, the generated code will directly overwrite the existing files.

Fixes #447

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,6 +78,9 @@ The configuration supported by the plugin option includes:
     **Note:** Idiomatically the name is last element of the path but it need not be.
     For instance, the last element of the path might be the package's version, and the package would benefit
     from a more descriptive name.
+
+  * `module`: prefix to be stripped from the `go-gapic-package` used in the generated filenames.
+     * _Note: This option is not supported from the Bazel interface._
   
   * `grpc-service-config`: the path to a gRPC ServiceConfig JSON file.
     * This is used for client-side retry configuration in accordance with [AIP-4221](http://aip.dev/4221)
--- a/internal/gengapic/gengapic.go
+++ b/internal/gengapic/gengapic.go
@@ -73,6 +73,10 @@ func Gen(genReq *plugin.CodeGeneratorRequest) (*plugin.CodeGeneratorResponse, er
 		}
 
 		key, val := s[:e], s[e+1:]
+		if val == "" {
+			return &g.resp, errors.E(nil, "invalid plugin option value, missing value in key=value: %s", s)
+		}
+
 		switch key {
 		case "go-gapic-package":
 			p := strings.IndexByte(s, ';')
--- a/internal/gengapic/gengapic.go
+++ b/internal/gengapic/gengapic.go
@@ -66,7 +66,8 @@ func Gen(genReq *plugin.CodeGeneratorRequest) (*plugin.CodeGeneratorResponse, er
 		if e < 0 {
 			e = len(s)
 		}
-		switch s[:e] {
+		key, val := s[:e], s[e+1:]
+		switch key {
 		case "go-gapic-package":
 			p := strings.IndexByte(s, ';')
 
@@ -78,7 +79,7 @@ func Gen(genReq *plugin.CodeGeneratorRequest) (*plugin.CodeGeneratorResponse, er
 			pkgName = s[p+1:]
 			outDir = filepath.FromSlash(pkgPath)
 		case "gapic-service-config":
-			f, err := os.Open(s[e+1:])
+			f, err := os.Open(val)
 			if err != nil {
 				return &g.resp, errors.E(nil, "error opening service config: %v", err)
 			}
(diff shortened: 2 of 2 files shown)


internal/gengapic/gengapic.go
```diff
@@ -66,7 +66,8 @@ func Gen(genReq *plugin.CodeGeneratorRequest) (*plugin.CodeGeneratorResponse, er
 		if e < 0 {
 			e = len(s)
 		}
-		switch s[:e] {
+		key, val := s[:e], s[e+1:]
```
tbpg, line 69: `e` can be set to `len(s)` above. So, this can panic.

https://play.golang.org/p/6V_5jjOBokQ

README.md
```diff
@@ -78,6 +78,9 @@ The configuration supported by the plugin option includes:
     **Note:** Idiomatically the name is last element of the path but it need not be.
     For instance, the last element of the path might be the package's version, and the package would benefit
     from a more descriptive name.
+
+  * `module`: prefix to be stripped from the `go-gapic-package` used in the generated filenames.
```
tbpg, line 82: Is this `modulePrefix` or just `module`?

tbpg (changes requested):

README.md
valery-barysok: actually prefix, i can provide something like this in case of `protoc-gen-go`:

```
		--go_out=new_folder \
		--go_opt=module=${DELOREAN_GENPROTO_GO_MODULE}/old_folder \
```

internal/gengapic/gengapic.go
```diff
@@ -108,6 +111,10 @@ func Gen(genReq *plugin.CodeGeneratorRequest) (*plugin.CodeGeneratorResponse, er
 		return &g.resp, errors.E(nil, paramError)
 	}
 
+	if g.modulePrefix != "" && strings.HasPrefix(outDir, g.modulePrefix) {
```
valery-barysok, line 114: What about to handle wrong prefix like here: https://go-review.googlesource.com/c/protobuf/+/219298/4/compiler/protogen/protogen.go#417

README.md
noahdietz: The command line option is named `module`, as @valery-barysok indicates, I wanted to align with what already exists in other plugins. I named the containing variable in the code `modulePrefix` because I felt it was more readable, and to no get confused with the `go-gapic-package` value. I can change the variable to just `module` if you think that's better, but I think the cli option should stay as `module`.

internal/gengapic/gengapic.go
noahdietz: Good point. There is only one plugin option that isn't a `key=value` pair, `sample-only`, so I'll just move that check up and return an error when the format is incorrect. Thanks for catching that.

internal/gengapic/gengapic.go
noahdietz: Done.

internal/gengapic/gengapic.go
noahdietz: Done.

internal/gengapic/gengapic.go
```diff
@@ -62,11 +62,18 @@ func Gen(genReq *plugin.CodeGeneratorRequest) (*plugin.CodeGeneratorResponse, er
...
+		}
+
 		e := strings.IndexByte(s, '=')
 		if e < 0 {
-			e = len(s)
+			return &g.resp, errors.E(nil, "invalid plugin option format, must be key=value: %s", s)
 		}
-		switch s[:e] {
+
+		key, val := s[:e], s[e+1:]
```
tbpg, line 75: Check `val == ""`, just in case? Won't panic, but won't be correct. Could also do `if e < 0 || e == len(s) - 1` above.

tbpg (approved):

internal/gengapic/gengapic.go
noahdietz: I went with a `val == ""` check. Done.

PR #451, opened 2020-10-02
chore: wrap gazelle update-repo in make

noahdietz:
This wraps the `gazelle` `update-repos` command in Make target to simplify updating the Bazel repositories with the Go modules.

`gazelle` will only generate `go_repository` rules to load Go module dependencies, but we like to wrap our dependencies with a `_maybe` macro, in case another section earlier in a WORKSPACE has already imported the dependency. So, I've defined a `go_repository` macro that invokes `gazelle`'s `go_repository` with our `_maybe`. `gazelle` `update-repos` continues to generate `go_repository` rules from our `go.mod`. 

We also have to do a little find-remove because `gazelle` generates a load statement for `go_repository` even though we have already loaded it with an alias (`gazelle_go_repository = "go_repository"`).

Finally, the last hack, the `go.sum` declares `cloud.google.com/go` as a transitive dependency brought in at some point in the tree, thus making `gazelle` include it as a `go_repository`. The `gapic-generator-go` does not depend on this module at all and it shouldn't be imported as a build dependency of the generator. So, we allow `gazelle` to include the dep, but we use our `go_repository` wrapper to explicitly ignore it. This is easier than more find-remove scripting.

### DIFF
--- a/repositories.bzl
+++ b/repositories.bzl
@@ -19,7 +19,7 @@ load(
     gazelle_go_repository = "go_repository",
 )
 
-def go_modules():
+def com_googleapis_gapic_generator_go_repositories():
     go_repository(
         name = "co_honnef_go_tools",
         importpath = "honnef.co/go/tools",
@@ -94,7 +94,8 @@ def go_modules():
     )
     go_repository(
         name = "com_github_jhump_protoreflect",
-        # Manually added in order to disable testproto BUILD file generation.
+        # Added in order to disable testproto BUILD file generation.
+        # This should be retained by gazelle.
         build_file_proto_mode = "disable",
         importpath = "github.com/jhump/protoreflect",
         sum = "h1:qJ7piXPrjP3mDrfHf5ATkxfLix8ANs226vpo0aACOn0=",
@@ -195,6 +196,12 @@ def go_modules():
         importpath = "gitlab.com/opennota/wd",
         sum = "h1:uPZaMiz6Sz0PZs3IZJWpU5qHKGNy///1pacZC9txiUI=",
         version = "v0.0.0-20180912061657-c5d65f63c638",
+    )
+    _maybe(
+        http_archive,
+        name = "com_google_api_codegen",
+        strip_prefix = "gapic-generator-2.4.5",
+        urls = ["https://github.com/googleapis/gapic-generator/archive/v2.4.5.zip"],
     )
     go_repository(
         name = "com_google_cloud_go",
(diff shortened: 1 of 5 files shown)


codyoss (approved):
LGTM

repositories.bzl
```diff
@@ -13,351 +13,341 @@
...
     if repo_name in native.existing_rules():
         return
     repo_rule(name = repo_name, **kwargs)
+
+# Redefine gazelle's go_repository rule with our own that will wrap go_repository
+# with the _maybe macro. gazelle update-repos can only generate go_reposoritory
+# targets. This way we get both the _maybe functionality and still use
+# gazelle update-repos. The real go_repository rule is loaded with an alias:
+# gazelle_go_repository.
+def go_repository(name, importpath, sum, version, build_file_proto_mode = ""):
```
vam-google, line 336: Nice trick!

repositories.bzl
```diff
@@ -13,351 +13,341 @@
...
         name = "com_github_gordonklaus_ineffassign",
         importpath = "github.com/gordonklaus/ineffassign",
         sum = "h1:vc7Dmrk4JwS0ZPS6WZvWlwDflgDTA26jItmbSj83nug=",
         version = "v0.0.0-20200309095847-7953dde2c7bf",
     )
-    _maybe(
-        go_repository,
+    go_repository(
         name = "com_github_jhump_protoreflect",
+        # Manually added in order to disable testproto BUILD file generation.
```
vam-google, line 97: just a qq - is this file (or at least the content of this file) generated? If yes, how do we make sure that this "disable" parameter does not get overridden?

vam-google (approved):
LGTM

repositories.bzl
noahdietz: It is retained by `gazelle`. I've done some tests and it seems to be kept. My guess is because `gazelle` does some merging of the existing rules with `buildozer`. I also updated the comment to remove the implication that I manually update that attribute every time - I've done it once, and `gazelle` keeps it now :)

repositories.bzl
noahdietz: Thanks! A little hacky, but seems simple enough.

repositories.bzl
noahdietz: And yes, the contents of the _macro_ is generated - except for the `gapic-generator` dependency, that was added by me, but is retained by `gazelle`.

PR #457, opened 2020-10-14
bazel: fix empty release-level param

noahdietz:
Recent changes to plugin param parsing made sending an empty `release-level` param an error. The bazel macro defaults the `release_level` attribute to `""` and was supplying `"release-level="` if it went unset, which would produce an error even though this attribute/param is optional. There is no bug in googleapis today as the param parsing changes haven't been released, but updating this generator in the WORKSPACE would've caused the error to happen.

Tested this fix against all of googleapis.


codyoss (approved):
LGTM

PR #458, opened 2020-10-15
bazel: use WKT for extra gapic deps

noahdietz:
Use the rules_go WKT for some common types instead of directly depending on the Go module that owns them for deps included by the generator on behalf of the generated client.

Cleans up some import conflict warnings.


vam-google (approved):
LGTM

codyoss (approved):

codyoss:
@noahdietz Might be good to add vam as a codeowner here too. No need to block on just yoshi-go

PR #464, opened 2020-10-21
chore: add showcase integration test readme

noahdietz:
Adds a README to the showcase integration testing submodule.

### DIFF
--- a/showcase/README.md
+++ b/showcase/README.md
@@ -7,31 +7,38 @@ gapic-generator-go, as well as the script used to setup and execute them.
 
 The tests can be run out-of-the-box in the normal `go test` fashion, so long as
 there is a `gapic-showcase` server already running locally. This does not test
-the a client generated with the locally checked out generator - it would use the
-released version of the GAPIC. Running `make test` will execute the tests
-against the locally install generator.
+the Showcase client generated with the locally installed generator - it would
+use the released version of the GAPIC. Running `make test` will execute the
+tests against the locally installed generator.
 
-To test the local generator, the `showcase.bash` script first downloads the
-Showcase artifacts associated with the targeted version. These are the compiled
-proto descriptor set, the retry configuration, and a pre-compiled server binary.
+To test the local generator, use the `showcase.bash` script. It does the
+following:
 
-Then using protoc and the retrieved artifacts as input, the Go protobuf/gRPC
+1. Downloads the Showcase artifacts associated with the targeted version.
+These are the compiled proto descriptor set, the retry configuration, and a
+pre-compiled server binary.
+
+1. Using protoc and the retrieved artifacts as input, the Go protobuf/gRPC
 bindings, and the Go GAPIC are generated, the latter with the
 **locally installed** gapic-generator-go. Make sure to `make install` so that
 any new changes are utilized during generation.
 
-This submodule's `go.mod` is temporarily edited to replace the remote dependency
-on `gapic-showcase` with the locally generated artifacts.
+1. The submodule's `go.mod` is temporarily edited to replace the remote
+dependency on `github.com/googleapis/gapic-showcase` with the locally generated
+artifacts.
 
-The server binary is run in the background and the tests in this directory are
-executed against it.
+1. The server binary is started in the background.
+
+1. The tests in this directory are executed against the locally running server.
+
+1. The server process is stopped.
 
 ### Adding tests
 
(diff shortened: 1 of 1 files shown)


codyoss (approved):

showcase/README.md
```diff
@@ -0,0 +1,49 @@
+# Showcase integration testing
+
+This submodule contains the gapic-showcase integration tests for
+gapic-generator-go, as well as the script used to setup and execute them.
+
+## How the tests work
+
+The tests can be run out-of-the-box in the normal `go test` fashion, so long as
+there is a `gapic-showcase` server already running locally. This does not test
+the a client generated with the locally checked out generator - it would use the
```
vchudnov-g, line 10: typo: "the a"

showcase/README.md
```diff
@@ -0,0 +1,49 @@
...
+released version of the GAPIC. Running `make test` will execute the tests
+against the locally install generator.
```
vchudnov-g, line 12: "installed"

showcase/README.md
vchudnov-g, line 10: hyphen: "checked-out"

vchudnov-g, line 10: I would clarify "Showcase client"

showcase/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+To test the local generator, the `showcase.bash` script first downloads the
+Showcase artifacts associated with the targeted version. These are the compiled
+proto descriptor set, the retry configuration, and a pre-compiled server binary.
+
+Then using protoc and the retrieved artifacts as input, the Go protobuf/gRPC
```
vchudnov-g, line 18: nit: comma after "Then"

showcase/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+## How the tests work
+
+The tests can be run out-of-the-box in the normal `go test` fashion, so long as
+there is a `gapic-showcase` server already running locally. This does not test
+the a client generated with the locally checked out generator - it would use the
+released version of the GAPIC. Running `make test` will execute the tests
+against the locally install generator.
+
+To test the local generator, the `showcase.bash` script first downloads the
```
vchudnov-g, line 14: I would restructure this and the following paragraphs to be more list-like and active voice. WDYT?

```
To test the local generator, use the showcase.bash script. It does the following:
1. It downloads the ....
2. It generates the ... using protoc and ....
  - make sure to make install first....
3. It edits this submodule's go.mod....
4. It runs the server in the background
5. It runs the tests against the server
6. It stops the server once the tests are finished.
```

showcase/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+The server binary is run in the background and the tests in this directory are
+executed against it.
+
+### Adding tests
+
+Adding tests for an existing service is as easy as adding a new Go test to the
+respective file: `echo_test.go`, `identity_test.go`, or `sequence_test.go`.
+
+Adding a new service for testiing requires creating an appropriate
```
vchudnov-g, line 34: sp: "testing"

showcase/README.md
```diff
@@ -0,0 +1,49 @@
...
+`{service}_test.go` file and adding client creation to the `main_test.go`.
+
+Any newly added tests should make sure to add a `leakcheck` statement to the
+test function: `defer check(t)`.
+
+### Updating the Showcase version
```
vchudnov-g, line 40: This is probably not worth adding in this doc, but I'll bring it up since it's on my mind: to use a local version of Showcase, one would also need to do a replace in go.mod to point to the local repo.

vchudnov-g (approved):
This will be very helpful. Thank you!

I just have some typo fixes and a minor suggestion.

showcase/README.md
noahdietz: Done.

showcase/README.md
noahdietz: Good idea! Done.

showcase/README.md
noahdietz: This is not easy to do atm. This script depends on the compiled proto descriptor, which is created by the release process. Locally, you'd have to compile that descriptor, then we'd have to wire up the script to check if some environment variable is set to point it at the locally compiled descriptor (or you'd just copy it into this dir and comment out the `curl`).

Using a local version of showcase for any of this is not set up to be easy atm. But that can change!

PR #465, opened 2020-10-21
gapic: use DefaultEndpoint and DefaultMTLSEndpoint for default options

andyrzhao:
The option "WithEndpoint" is intended for user override only, and specifying it will interfere with mTLS endpoint selection. See details regarding mTLS at https://google.aip.dev/auth/4114


noahdietz (commented):
LGTM but I'd like @codyoss to review/approve as well.

codyoss (approved):
LGTM

PR #475, opened 2020-11-03
feat: generate a default audience for clients

codyoss:
This option will be used to generate self-signed JWTs when the
client is authenticated with a service account and the user does
not provide additional scopes.

Also, now marking the generated scopes as default so we can
distinguish between generated and user provided scopes.

Related: googleapis/google-api-go-client#738


codyoss:
Before this PR is merged:

1. The associated PR needs to be merged.
2. A new version of google-api-go-client needs to be released with those changes
3. google-cloud-go needs to be updated to depend on the new version.

codyoss:
cc @bshaffer

internal/gengapic/client_init.go
```diff
@@ -310,3 +311,22 @@ func generateDefaultMTLSEndpoint(defaultEndpoint string) string {
 	}
 	return defaultEndpoint
 }
+
+// generateDefaultAudience transforms a host into a an audience that can be used
+// as the `aud` claim in a JWT.
+func generateDefaultAudience(host string) string {
```
vchudnov-g, line 317: Are we guaranteed that the host is a proper host (ie doesn't have a path)? I'd be inclined to defensively remove it (and include that in tests).

internal/gengapic/client_init.go
```diff
@@ -310,3 +311,22 @@ func generateDefaultMTLSEndpoint(defaultEndpoint string) string {
...
+	aud := host
+	// Add a scheme if not present.
+	if !strings.Contains(aud, "://") {
+		aud = "https://" + aud
+	}
+	// Remove port if present.
+	if strings.Count(aud, ":") > 1 {
+		aud = aud[:strings.LastIndex(aud, ":")]
```
vchudnov-g, line 325: nit: How about truncating from the second colon, not from the last, in case the host includes a path with a colon? (Maybe a non-issue, but it would make this function more portable.)

internal/gengapic/client_init_test.go
```diff
@@ -208,3 +208,25 @@ func TestClientInit(t *testing.T) {
...
+	want := "https://foo.googleapis.com/"
+	tests := []struct {
+		name string
+		host string
+	}{
+		{name: "plain host", host: "foo.googleapis.com"},
+		{name: "host with port", host: "foo.googleapis.com:443"},
+		{name: "host with scheme", host: "https://foo.googleapis.com"},
+		{name: "host with scheme and port", host: "https://foo.googleapis.com:1234"},
+		{name: "host is a proper audience", host: want},
```
vchudnov-g, line 222: I'd also test that a custom scheme is preserved: `snazzy://foo.example.org:534`

vchudnov-g (approved):
No comment on the JWT semantics, but the Go looks fine. Some minor, non-blocking suggestions.

internal/gengapic/client_init.go
```diff
@@ -67,8 +67,9 @@ func (g *generator) clientOptions(serv *descriptor.ServiceDescriptorProto, servN
 		p("  return []option.ClientOption{")
 		p("    internaloption.WithDefaultEndpoint(%q),", host)
 		p("    internaloption.WithDefaultMTLSEndpoint(%q),", generateDefaultMTLSEndpoint(host))
+		p("    internaloption.WithDefaultAudience(%q),", generateDefaultAudience(host))
 		p("    option.WithGRPCDialOption(grpc.WithDisableServiceConfig()),")
-		p("    option.WithScopes(DefaultAuthScopes()...),")
+		p("    internaloption.WithDefaultScopes(DefaultAuthScopes()...),")
```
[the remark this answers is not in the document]
codyoss: Done.

internal/gengapic/client_init_test.go
codyoss: Done.

internal/gengapic/client_init.go
codyoss: Guaranteed, no. But it does not happen today I can fairly certainly say. On L63 of this file we append `:443` to the the end of the host. If there were path fragments this should be an invalid host and the client would not be able to talk to the backend.

internal/gengapic/client_init.go
codyoss: Sure, done.

vchudnov-g (approved):
Looks great! Thanks!

PR #479, opened 2020-11-13
feat: generator options are parsed into a standalone struct

software-dov:
Lifts some generator options into a separate structure.
Lifts option parameter parsing logic into a standalone function.
Adds parsing logic for permitted transports.

Also adds unit tests for parsing transport options


internal/gengapic/gengapic.go
```diff
@@ -52,82 +51,164 @@ const (
...
+		if !strings.HasPrefix(opts.outDir, opts.modulePrefix) {
+			return nil, errors.E(nil, "go-gapic-package %q does not match prefix %q", opts.outDir, opts.modulePrefix)
 		}
-		outDir = strings.TrimPrefix(outDir, g.modulePrefix+"/")
+		opts.outDir = strings.TrimPrefix(opts.outDir, opts.modulePrefix+"/")
 	}
 
+	// Default is just grpc
+	if opts.transports == nil {
+		opts.transports = []Transport{grpc}
```
miraleung, line 171: Consider moving this assignment to line 102 to better group the initialization together.

internal/gengapic/gengapic.go
```diff
@@ -52,82 +51,164 @@ const (
...
+type options struct {
+	pkgPath           string
+	pkgName           string
+	outDir            string
+	relLvl            string
+	modulePrefix      string
+	grpcConfPath      string
+	serviceConfigPath string
+	sampleOnly        bool
+	transports        []Transport
```
miraleung, line 70: I believe Java assumes that either gRPC or REST will be used at a particular time. Does this array imply that callers can pass in both transport types?

internal/gengapic/gengapic.go
```diff
@@ -149,30 +230,30 @@ func Gen(genReq *plugin.CodeGeneratorRequest) (*plugin.CodeGeneratorResponse, er
 		// Keep the current behavior for now, but we could revisit this later.
 		outFile := pbinfo.ReduceServName(s.GetName(), "")
 		outFile = camelToSnake(outFile)
-		outFile = filepath.Join(outDir, outFile)
+		outFile = filepath.Join(g.opts.outDir, outFile)
 
 		g.reset()
-		if err := g.gen(s, pkgName); err != nil {
+		if err := g.gen(s, g.opts.pkgName); err != nil {
```
vchudnov-g, line 236: Here and below, do we want to pass `g.opts.pkgName` to `g.gen()`, or have the method look at its `g.opts` directly? The latter might be simpler while not hindering testing via fake values, etc.

internal/gengapic/gengapic.go
```diff
@@ -149,30 +230,30 @@ func Gen(genReq *plugin.CodeGeneratorRequest) (*plugin.CodeGeneratorResponse, er
...
+		g.commit(outFile+"_client_example_test.go", g.opts.pkgName+"_test")
 	}
 
 	g.reset()
 	scopes, err := collectScopes(genServs, g.serviceConfig)
 	if err != nil {
 		return &g.resp, err
 	}
-	g.genDocFile(pkgPath, pkgName, time.Now().Year(), scopes)
+	g.genDocFile(g.opts.pkgPath, g.opts.pkgName, time.Now().Year(), scopes)
```
vchudnov-g, line 254: Similarly here: maybe we can omit the `g.opts.xxx` args when calling the `g.yyy()` method.

internal/gengapic/gengapic.go
software-dov: Yes, that is the implication. [Python](https://github.com/googleapis/gapic-generator-python/pull/688) allows multiple transports to be generated for the surface with runtime choice via dependency injection. This is supported by the spec: see [this](https://docs.google.com/document/d/1u0WGcF6woB9HnyUK5J-GaBCRUc9G4znwbvcB4QdYq8s/edit#heading=h.lebdm4farlvy) heading.

internal/gengapic/gengapic_test.go
```diff
@@ -499,3 +499,55 @@ func Test_isLRO(t *testing.T) {
...
 	}
 }
+
+func Test_transportParse(t *testing.T) {
+	for _, tst := range []struct {
+		param        string
+		expectedOpts options
+		expectErr    bool
+	}{
+		{
```
vchudnov-g, line 509: Maybe you also want to test more  error conditions (not just transport): missing `pkgPath`, `pkgName`, `outDir` and non-matching `modulePrefix`

internal/gengapic/gengapic_test.go
```diff
@@ -499,3 +499,55 @@ func Test_isLRO(t *testing.T) {
...
+			expectedOpts: options{
+				transports: []Transport{grpc},
+				pkgPath:    "path",
+				pkgName:    "pkg",
+				outDir:     "path",
+			},
+		},
+	} {
+		opts, err := ParseOptions(&tst.param)
+		if tst.expectErr {
```
vchudnov-g, line 545: You should also check you don't get an error when you don't expect one. Maybe `if test.ExpectErr != (err!=nil)` ?

internal/gengapic/gengapic_test.go
```diff
@@ -499,3 +499,46 @@ func Test_isLRO(t *testing.T) {
...
+		},
+		{
+			param:        "transport=tcp,go-gapic-package=path;pkg",
+			expectedOpts: options{},
+			expectErr:    true,
+		},
+	} {
+		opts, err := ParseOptions(&tst.param)
+		if tst.expectErr {
+			if err == nil {
```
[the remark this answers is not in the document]
vchudnov-g: `if tst.expectErr && (err == nil)` doesn't work? Weird....

internal/gengapic/gengapic.go
```diff
@@ -52,82 +51,164 @@ const (
...
+	outDir            string
+	relLvl            string
+	modulePrefix      string
+	grpcConfPath      string
+	serviceConfigPath string
+	sampleOnly        bool
+	transports        []Transport
+}
+
+func TransportSliceEqual(a, b []Transport) bool {
```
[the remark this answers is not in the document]
software-dov: My cursory doc readthrough missed `reflect.DeepEqual`. I'll use that instead.

internal/gengapic/gengapic_test.go
software-dov: Good point. Will add.

software-dov:
It's quite inexplicable and caused me at least an hour of frustration before I just gave up.

internal/gengapic/gengapic.go
software-dov: I think the current placement better communicates the intention that this is a default value if the user doesn't provide a `transport=` argument.

internal/gengapic/gengapic.go
software-dov: Good point. My original intention was to limit the blast radius of changed code, but it's better and more consistent to just use the attribute.

internal/gengapic/doc_file.go
```diff
@@ -29,14 +29,14 @@ import (
...
-func (g *generator) genDocFile(pkgPath, pkgName string, year int, scopes []string) {
+func (g *generator) genDocFile(year int, scopes []string) {
 	p := g.printf
 
 	p(license.Apache, year)
 	p("")
 
 	if g.apiName != "" {
-		p("// Package %s is an auto-generated package for the", pkgName)
+		p("// Package %s is an auto-generated package for the", g.opts.pkgName)
```
vchudnov-g, line 39: nit: space after `for the`?

internal/gengapic/gengapic_test.go
```diff
@@ -499,3 +500,86 @@ func Test_isLRO(t *testing.T) {
 		}
 	}
 }
+
+func Test_transportParse(t *testing.T) {
```
vchudnov-g, line 504: maybe change the name to `Test_optionParse` since you're testing other options as well.

internal/gengapic/gengapic_test.go
```diff
@@ -499,3 +500,86 @@ func Test_isLRO(t *testing.T) {
...
+			param:     "module=different_path,go-gapic-package=path;pkg",
+			expectErr: true,
+		},
+	} {
+		opts, err := ParseOptions(&tst.param)
+		if tst.expectErr {
+			if err == nil {
+				t.Errorf("ParseOptions(%s) expected error", tst.param)
+			}
+		} else {
```
vchudnov-g, line 576: idiomaticity: Go prefers exiting early and avoiding nested code. Son this case, maybe
```
if tst.ExpectErr {
 ...
  continue
}
if err != nil {
  ...
  continue  // <- you probably want this, but maybe not
}
if !reflect... {
  ...
}
}
```

internal/gengapic/gengapic_test.go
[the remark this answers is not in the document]
vchudnov-g: Suggestion for debugging this: have the test print `("%v", err)`. I worry this may hide a simple bug, so we should try to resolve this.

vchudnov-g (approved):
Just minor comments, nothing blocking.

(Although we should find out what's behind needing to do a nested if in the test, rather than a conjunction: I fear it hides a simple but potentially important bug.)

internal/gengapic/gengapic_test.go
software-dov: Yeah, the scope of the test really expanded.

internal/gengapic/doc_file.go
software-dov: Good catch

internal/gengapic/gengapic_test.go
software-dov: I am absolutely behind less nesting, but my first inclination is to try to do without `continue`. If my attempts in that area are unsuccessful I'll lean towards shallow+continue.

Tangent: one of my least favorite things in reading code, either code I've written or something other people have written, is complicated control flow. It's been big source of difficult to diagnose "push the bubble in the wallpaper" bugs where fixing one exposes a new problem. A big part of my python style is trying to avoid explicit control flow, and I need to retrain those muscles for golang's idioms.

internal/gengapic/gengapic_test.go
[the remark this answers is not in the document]
software-dov: I think I figured it out: the test struct is storing the expected options by value, but ParseOptions is returning by pointer, so the comparison fails because sometimes the returned options is nil, and the expected options cannot be nil. Changing the expected options field to a pointer fixed this.

PR #495, opened 2021-01-06
fix: relax paging missing repeated

### LINKED ISSUE #493: do not fail if pagination response field does not have repeated fields
[AIP-4233](https://google.aip.dev/4233) does not require the generator to fail if the pagination settings cannot be inferred from the request and response messages. If a response method does not contain any repeated fields, my understanding is that it just should not be considered a paginated method.

Code in question:

https://github.com/googleapis/gapic-generator-go/blob/483fa656ddea2ccf18d68e615d9ae78c12cdd4fe/internal/gengapic/paging.go#L148

https://github.com/googleapis/gapic-generator-go/blob/483fa656ddea2ccf18d68e615d9ae78c12cdd4fe/internal/gensample/paging.go#L71

Internal question ID: 9137275878353403904

noahdietz:
Previously, if an RPC had all of the necessary Pagination fields according to AIP-4233, but was missing a repeated field, the generator would fail. This is not a requirement of AIP-4233, instead it should treat the RPC as if it wasn't paginated, and generate a standard method.

Fixes #493

### DIFF
--- a/internal/gengapic/paging.go
+++ b/internal/gengapic/paging.go
@@ -141,12 +141,13 @@ func (g *generator) pagingField(m *descriptor.MethodDescriptorProto) (*descripto
 			elemFields = append(elemFields, f)
 		}
 	}
-	if !hasSize || !hasToken || !hasNextToken {
+
+	// If any of the required pagination fields are missing or if the response
+	// has no repeated fields, it is not a fully conforming paginated method.
+	if !hasSize || !hasToken || !hasNextToken || len(elemFields) == 0 {
 		return nil, nil
 	}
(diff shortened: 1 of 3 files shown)


vchudnov-g (approved):
lgtm. Thanks for doing this!

I would add at least a TODO in `gapic-generator-go/internal/gensample/paging.go` (I know that code is not in use and may wind up being removed, but while it's around, let's at least flag that a similar change needs to be made)

noahdietz:
Ah good call, I added a TODO comment :) Thanks Victor.

vchudnov-g (approved):

PR #499, opened 2021-01-07
chore: run basic tests on Actions

noahdietz:
Adds a GitHub Actions config for running basic linting, unit tests and bazel testing.

To remove circleci we'd need to port: 
- Showcase integration tests
- Docker image creation + publication
- release script execution (this should probably be ported to release-please)

Once Actions is confirmed to work, I will remove the duplicate jobs from the CircleCI config if possible

cc @alexander-fenster: thanks for the example to work from. I copied some of Luke's configs from the api-linter (a Go project) for basic testing/linting.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,0 +1,63 @@
+---
+name: Generator tests
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+
+jobs:
+  unit-tests:
+    runs-on: ubuntu-latest
+    container: golang:1.13
+    steps:
+      - uses: actions/checkout@v2
+      - run: go test -p 1 ./...
+  bazel-build:
+    runs-on: ubuntu-latest
+    container: gcr.io/gapic-images/googleapis-bazel:20210105
+    # Dockerfile for this image: https://github.com/googleapis/googleapis-discovery/blob/master/Dockerfile
+    # If you update its version, please also update it below in
+    # 'Cache Bazel files' - unfortunately it cannot accept variables at this
+    # time.
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Cache Bazel files
+      id: cache-bazel
+      uses: actions/cache@v2
+      with:
+        path: ~/.cache/bazel
+        key: ${{ runner.os }}-bazel-20210105-${{ secrets.CACHE_VERSION }}
+
+    - name: Cache not found
+      if: steps.cache-bazel.outputs.cache-hit != 'true'
+      run: |
+        echo "No cache found."
+
+    - name: Cache found
+      if: steps.cache-bazel.outputs.cache-hit == 'true'
+      run: |
+        echo -n "Cache found. Cache size: "
+        du -sh ~/.cache/bazel
+        echo "If the cache seems broken, update the CACHE_VERSION secret in"
+        echo "https://github.com/googleapis/googleapis-discovery/settings/secrets/actions"
+        echo "(use any random string, any GUID will work)"
+        echo "and it will start over with a clean cache."
+        echo "The old one will disappear after 7 days."
+
+    - name: Run bazel build
+      run: bazel build '//...'
+
+    - name: Run bazel test
+      run: bazel test '//...'
+  lint:
+    runs-on: ubuntu-latest
+    container: golang:1.13
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install golangci-lint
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yaml
```diff
@@ -0,0 +1,65 @@
...
+      - master
+  pull_request:
+
+jobs:
+  unit-tests:
+    runs-on: ubuntu-latest
+    container: golang:1.13
+    steps:
+      - uses: actions/checkout@v2
+      - run: go test -p 1 ./...
```
codyoss, line 15: Might want a actions/setup-go@v2 and pin a Go version. I am not sure what we be used otherwise.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,65 @@
...
+        echo -n "Cache found. Cache size: "
+        du -sh ~/.cache/bazel
+        echo "If the cache seems broken, update the CACHE_VERSION secret in"
+        echo "https://github.com/googleapis/googleapis-discovery/settings/secrets/actions"
+        echo "(use any random string, any GUID will work)"
+        echo "and it will start over with a clean cache."
+        echo "The old one will disappear after 7 days."
+
+    - name: Run bazel build
+      run: bazel build '//...'
```
vchudnov-g, line 51: Q: This is running `bazel build` at the top of the Showcase repo? If so, where is it specified what version of Go it's using? (the `unit-tests` job has a `golang` container, but I'm not seeing a Go version specified for this job's container (https://github.com/googleapis/googleapis-discovery/blob/master/Dockerfile)

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,65 @@
...
+  unit-tests:
+    runs-on: ubuntu-latest
+    container: golang:1.13
+    steps:
+      - uses: actions/checkout@v2
+      - run: go test -p 1 ./...
+  bazel-build:
+    runs-on: ubuntu-latest
+    container: gcr.io/gapic-images/googleapis-bazel:20210105
+    # Dockerfile for this image: https://github.com/googleapis/googleapis-discovery/blob/master/Dockerfile
```
vchudnov-g, line 19: Maybe we should move to the Dockerfile to a more central place, not one that's tied to `googleapis-discovery` (That's outside the scope of this PR; just suggesting this)

.github/workflows/ci.yaml
noahdietz: I'm not not following, this isn't the Showcase repo, its the Go micro-generator.

The bazel build is using the version of Go specified in the WORKSPACE here https://github.com/googleapis/gapic-generator-go/blob/eadb3631a0520ea9b439d8d7ec4885122589e003/WORKSPACE#L30.

That one is on 1.15.6 because of requirements in the version of rules_go we are using. We need that version of rules_go because of changes in the protobuf and google.golang.org/api repos that we need. The bazel deps are a little bit of a mess tbh.

vchudnov-g: Ah, OK. 
- Sorry, I should have said generator-go; my mistake.
- OK, the WORKSPACE file pins the version. sgtm

vchudnov-g (approved):

.github/workflows/ci.yaml
noahdietz: Good idea, much clearer. Done.

.github/workflows/ci.yaml
noahdietz: cc @alexander-fenster 

Yeah maybe so. We might start using it more broadly if generator repos run bazel builds.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,71 @@
...
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Cache Bazel files
+      id: cache-bazel
+      uses: actions/cache@v2
+      with:
+        path: ~/.cache/bazel
+        key: ${{ runner.os }}-bazel-20210105-${{ secrets.CACHE_VERSION }}
```
alexander-fenster, line 49: Right now there is no secret so it probably uses an empty string. That's OK! If we ever find that the cache is broken and we need to invalidate it, we'll add a `CACHE_VERSION` secret (any random string) and it will start over. This is a semi-official hack since there is no way to force cache removal.

Alternatively, you can just remove this step altogether if you don't need caching.

alexander-fenster (approved):
LGTM! Let me know when this is merged, I'll remove Circle tasks from the required list.

.github/workflows/ci.yaml
noahdietz: I just removed all of the caching stuff.

PR #504, opened 2021-01-08
chore: fix release action

noahdietz:
Attempting to fix the releasing action.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -12,8 +12,9 @@ jobs:
       - uses: actions/setup-go@v2
         with:
           go-version: '^1.13.1'
+      - name: Parse tag name
+        run: echo "RELEASE_VERSION=${GITHUB_REF#refs/*/}" >> $GITHUB_ENV
       - name: Create GitHub release
-        run: go run ./util/cmd/release \
-              -version=${GITHUB_REF#refs/tags/} \
-              -commitish=${{ github.sha }} \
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -12,8 +12,9 @@ jobs:
...
+      - name: Parse tag name
+        run: echo "RELEASE_VERSION=${GITHUB_REF#refs/*/}" >> $GITHUB_ENV
       - name: Create GitHub release
-        run: go run ./util/cmd/release \
-              -version=${GITHUB_REF#refs/tags/} \
-              -commitish=${{ github.sha }} \
-              -token=${{ github.token }}
+        run: | 
+          echo "Creating release $RELEASE_VERSION at ${{ github.sha }}"
+          go run ./util/cmd/release -version=$RELEASE_VERSION -commitish=${{ github.sha }} -token=${{ github.token }}
```
alexander-fenster, line 20: Just to simplify debugging - can you add literally the same line but with `echo` appended, before `go run`?

```
echo go run ./util/cmd/release -version=$RELEASE_VERSION -commitish=${{ github.sha }} -token=${{ github.token }}
```

alexander-fenster (approved):

.github/workflows/release.yaml
noahdietz: Sure thing!

vchudnov-g (approved):

Issue #514, opened 2021-01-14
metadata headers: numbers in field names not read; non-string fields not converted


vchudnov-g:
As part of https://github.com/googleapis/gapic-showcase/pull/532, I am adding an RPC with http rules for path params:

```
  rpc RepeatDataSimplePath(RepeatRequest) returns (RepeatResponse) {
    option (google.api.http) = {
      get: "/v1beta1/repeat/{info.f_string}/{info.f_int32}/{info.f_double}/{info.f_bool}:simplepath"
    };
```

which causes gapic-generator-go to generate the following code (formatted for readability here):
```
func (c *ComplianceClient) RepeatDataSimplePath(ctx context.Context, req *genprotopb.RepeatRequest, opts ...gax.CallOption) (*genprotopb.RepeatResponse, error) {
	md := metadata.Pairs("x-goog-request-params", 
                    fmt.Sprintf("%s=%v&%s=%v&%s=%v&%s=%v", 
                      "info.f_string", url.QueryEscape(req.GetInfo().GetFString()), 
                      "info.f_int", url.QueryEscape(req.GetInfo().GetFInt()), 
                      "info.f_double",  url.QueryEscape(req.GetInfo().GetFDouble()), 
                      "info.f_bool", url.QueryEscape(req.GetInfo().GetFBool())))
…
}
```

This client leads to the following compile-time errors:

- [x] `req.GetInfo().GetFInt undefined (type *genproto.ComplianceData has no field or method GetFInt)`
  - There are fields `f_int32` and `f_int64`, but no field `f_int`. I suspect we're only expecting/parsing letters in the field params
- [x] `cannot use req.GetInfo().GetFDouble() (type float64) as type string in argument to url.QueryEscape`
   - I suspect this has only been tested on path params for string fields in the request
- [x] `cannot use req.GetInfo().GetFBool() (type bool) as type string in argument to url.QueryEscape`
   - ditto

For the time being, I am working around this by commenting out the offending lines via Showcases's `compile_protos.go` utility.

noahdietz:
For the first one it is definitely a regexp issue. Only capturing letters, not numbers. Easily fixed.

noahdietz:
For the other two, we can add a check if the "field type != string" and if not, we can either:
A. generate a call to `strconv.Format(Bool|Float|Int)` around the `req.GetInfo().GetF{Double|Int64|Bool}()`
[...]

PR #516, opened 2021-01-15
fix: handle non-string header field types

noahdietz:
This adds support for non-`string` field types that are declared as header request params. For non-`string` types, the value at runtime is not `url.QueryEscape`'d. For proto field types that translate to a `float64`, the `%g` formatter is used to format the value in either exponent or decimal form depending on what is best. See https://golang.org/pkg/fmt/ for more on format modes. The resulting formatted value is `QueryEscape`'d in case there is a `+` in the exponent section.

A helper `lookupFieldType` in order to learn the type of the field declared as a header request param.

Furthermore, the existing tests were updated to account for the new lookup of field types in header request param construction. New tests were also added to cover the new methods.

Open TODO to handle `[]byte`.

#514

### DIFF
--- a/internal/gengapic/gengapic.go
+++ b/internal/gengapic/gengapic.go
@@ -619,12 +619,31 @@ func (g *generator) insertMetadata(m *descriptor.MethodDescriptorProto) error {
 			}
 			seen[field] = true
 
+			typ := g.lookupFieldType(m.GetInputType(), field)
+
+			accessor := fmt.Sprintf("req%s", buildAccessor(field))
+			if typ == descriptor.FieldDescriptorProto_TYPE_STRING {
+				accessor = fmt.Sprintf("url.QueryEscape(%s)", accessor)
+			}
+
 			// URL encode key & values separately per aip.dev/4222.
 			// Encode the key ahead of time to reduce clutter
 			// and because it will likely never be necessary
-			fmt.Fprintf(&values, " %q, url.QueryEscape(req%s),",
-				url.QueryEscape(field), buildAccessor(field))
-			formats.WriteString("%s=%v&")
+			fmt.Fprintf(&values, " %q, %s,", url.QueryEscape(field), accessor)
+
+			// The defualt format specifiers for bool, string, and int are
+			// sufficient.
+			//
+			// TODO(noahdietz): need to handle []byte for TYPE_BYTES.
+			formatter := "%v"
+			if typ == descriptor.FieldDescriptorProto_TYPE_DOUBLE || typ == descriptor.FieldDescriptorProto_TYPE_FLOAT {
+				// Use the format mode 'f' to use decimal format, which should
+				// be safe for headers/URLs.
+				// See golang.org/pkg/fmt for more information.
+				formatter = ("%f")
+			}
+			// Produces either "%s=%v&" or "%s=%f&" depending on the field type.
+			formats.WriteString("%s=" + formatter + "&")
 		}
 		f := formats.String()[:formats.Len()-1]
 		v := values.String()[:values.Len()-1]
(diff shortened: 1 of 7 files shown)


internal/gengapic/gengapic.go
```diff
@@ -619,11 +619,26 @@ func (g *generator) insertMetadata(m *descriptor.MethodDescriptorProto) error {
...
+			// TODO(noahdietz): need to handle []byte for TYPE_BYTES.
+			if typ == descriptor.FieldDescriptorProto_TYPE_STRING {
+				accessor = fmt.Sprintf("url.QueryEscape(%s)", accessor)
+			} else if typ == descriptor.FieldDescriptorProto_TYPE_DOUBLE || typ == descriptor.FieldDescriptorProto_TYPE_FLOAT {
+				// Format the floating point value with mode 'g' to allow for
+				// exponent formatting when necessary, and decimal when adequate.
+				// QueryEscape the resulting string in case there is a '+' in the
+				// exponent.
+				// See golang.org/pkg/fmt for more information on formatting.
+				accessor = fmt.Sprintf("url.QueryEscape(fmt.Sprintf(%q, %s))", "%g", accessor)
```
vchudnov-g, line 634: nit: consider using `%%g` directly for readability

internal/gengapic/gengapic.go
```diff
@@ -619,11 +619,26 @@ func (g *generator) insertMetadata(m *descriptor.MethodDescriptorProto) error {
...
+			}
+
 			// URL encode key & values separately per aip.dev/4222.
 			// Encode the key ahead of time to reduce clutter
 			// and because it will likely never be necessary
-			fmt.Fprintf(&values, " %q, url.QueryEscape(req%s),",
-				url.QueryEscape(field), buildAccessor(field))
+			fmt.Fprintf(&values, " %q, %s,", url.QueryEscape(field), accessor)
+			// Produces either "%s=%v&" or "%s=%f&" depending on the field type.
```
vchudnov-g, line 641: I'm confused by this comment: the string below is not conditional

vchudnov-g (commented):
looks good!

internal/gengapic/gengapic.go
noahdietz: Cool! Done

internal/gengapic/gengapic.go
noahdietz: Oops it was a comment I forgot to delete when changing the code. Deleted!

internal/gengapic/gengapic.go
```diff
@@ -619,11 +619,25 @@ func (g *generator) insertMetadata(m *descriptor.MethodDescriptorProto) error {
...
+			// TODO(noahdietz): need to handle []byte for TYPE_BYTES.
+			if typ == descriptor.FieldDescriptorProto_TYPE_STRING {
+				accessor = fmt.Sprintf("url.QueryEscape(%s)", accessor)
+			} else if typ == descriptor.FieldDescriptorProto_TYPE_DOUBLE || typ == descriptor.FieldDescriptorProto_TYPE_FLOAT {
+				// Format the floating point value with mode 'g' to allow for
+				// exponent formatting when necessary, and decimal when adequate.
+				// QueryEscape the resulting string in case there is a '+' in the
+				// exponent.
+				// See golang.org/pkg/fmt for more information on formatting.
+				accessor = fmt.Sprintf("url.QueryEscape(fmt.Sprintf(\"%%g\", %s))", accessor)
```
tbpg, line 634: Consider using \` to avoid having to escape the double quotes.

internal/gengapic/gengapic.go
```diff
@@ -652,6 +666,33 @@ func buildAccessor(field string) string {
 	return ax.String()
 }
 
+func (g *generator) lookupFieldType(msgName, field string) (typ descriptor.FieldDescriptorProto_Type) {
```
tbpg, line 669: Consider not using naked returns for this function. See https://github.com/golang/go/wiki/CodeReviewComments#named-result-parameters, in particular, "Naked returns are okay if the function is a handful of lines. Once it's a medium sized function, be explicit with your return values."

noahdietz: Done

internal/gengapic/gengapic.go
noahdietz: Done

vchudnov-g (approved):
Thanks for doing this!

tbpg (approved):

PR #518, opened 2021-01-20
bazel: use protobuf-go v2 common types

noahdietz:
The generator bazel rule needs to provide the protobuf-go v2 common type imports because protobuf changed the `go_package` of the protos to point at protobuf-go v2. See https://github.com/bazelbuild/rules_go/issues/2721

We can't upgrade the version of protobuf in googleapis because of that change.


vam-google (approved):
LGTM

vchudnov-g (approved):

PR #519, opened 2021-01-21
feat: add an abstract client interface

software-dov:
The grpc client currently implements this interface, and the eventual
rest+json client will implement it as well.

The purpose of this interface is to allow runtime selection and/or
fallback between different transport types.

Further PRs will add an implementation for a rest+json client that adheres to the interface.

### DIFF
--- a/internal/gengapic/client_init.go
+++ b/internal/gengapic/client_init.go
@@ -206,7 +206,7 @@ func (g *generator) grpcClientInit(serv *descriptor.ServiceDescriptorProto, serv
 
 	// client struct
 	p("// %sClient is a client for interacting with %s over gRPC transport.", servName, g.apiName)
-	p("// It satisfies the %sAbstractClient interface", servName)
+	p("// It satisfies the %sAbstractClient interface.", servName)
 	p("//")
 	p("// Methods, except Close, may be called concurrently. However, fields must not be modified concurrently with method calls.")
 	p("type %sClient struct {", servName)
--- a/internal/gengapic/client_init.go
+++ b/internal/gengapic/client_init.go
@@ -135,9 +135,208 @@ func (g *generator) clientOptions(serv *descriptor.ServiceDescriptorProto, servN
 	return nil
 }
 
-func (g *generator) clientInit(serv *descriptor.ServiceDescriptorProto, servName string) error {
+func (g *generator) abstractClientIntfInit(serv *descriptor.ServiceDescriptorProto, servName string) error {
+	p := g.printf
+	p("// %sAbstractClient is an interface that defines the methods availaible from %s.", servName, g.apiName)
+	p("// It is agnostic as to the underlying transport, i.e. json+http, gRPC, or other.")
+	p("type %sAbstractClient interface {", servName)
+	for _, m := range serv.Method {
+		inType := g.descInfo.Type[*m.InputType]
+		inSpec, err := g.descInfo.ImportSpec(inType)
+		if err != nil {
+			return err
+		}
+		outType := g.descInfo.Type[*m.OutputType]
+		outSpec, err := g.descInfo.ImportSpec(outType)
+		if err != nil {
+			return err
+		}
+
+		if m.GetOutputType() == emptyType {
+			p("%s(context.Context, *%s.%s, ...gax.CallOption) error",
+				m.GetName(),
+				inSpec.Name,
+				inType.GetName())
+			continue
+		}
+
+		if pf, err := g.pagingField(m); err != nil {
+			return err
+		} else if pf != nil {
+			iter, err := g.iterTypeOf(pf)
+			if err != nil {
+				return err
+			}
+			p("%s(context.Context, *%s.%s, ...gax.CallOption) *%s",
+				m.GetName(),
+				inSpec.Name,
+				inType.GetName(),
+				iter.iterTypeName)
+			continue
+		}
+
+		switch {
+		case g.isLRO(m):
+			// Unary call where the return type is a wrapper of
+			// longrunning.Operation and more precise types
+			p("%s(context.Context, *%s.%s, ...gax.CallOption) (*%s, error)",
+				m.GetName(), inSpec.Name, inType.GetName(), lroTypeName(m.GetName()))
+		case m.GetClientStreaming():
+			// Handles both client-streaming and bidi-streaming
+			p("%s(context.Context, ...gax.CallOption) (%s.%s_%sClient, error)",
+				m.GetName(), inSpec.Name, serv.GetName(), m.GetName())
+		case m.GetServerStreaming():
+			// Handles _just_ server streaming
+			p("%s(context.Context, *%s.%s, ...gax.CallOption) (%s.%s_%sClient, error)",
+				m.GetName(), inSpec.Name, inType.GetName(), inSpec.Name, serv.GetName(), m.GetName())
+		default:
+			// Regular, unary call
+			p("%s(context.Context, *%s.%s, ...gax.CallOption) (*%s.%s, error)",
+				m.GetName(), inSpec.Name, inType.GetName(), outSpec.Name, outType.GetName())
+		}
+	}
+	p("}")
+	p("")
+
+	return nil
+}
+
+func (g *generator) grpcClientInit(serv *descriptor.ServiceDescriptorProto, servName string, imp pbinfo.ImportSpec, hasLRO bool) {
 	p := g.printf
 
+	// client struct
+	p("// %sClient is a client for interacting with %s.", servName, g.apiName)
+	p("//")
+	p("// Methods, except Close, may be called concurrently. However, fields must not be modified concurrently with method calls.")
+	p("type %sClient struct {", servName)
+
+	p("// Connection pool of gRPC connections to the service.")
+	p("connPool gtransport.ConnPool")
+	p("")
+
+	p("// flag to opt out of default deadlines via %s", disableDeadlinesVar)
+	p("disableDeadlines bool")
+	p("")
+
+	p("// The gRPC API client.")
+	p("%s %s.%sClient", grpcClientField(servName), imp.Name, serv.GetName())
+	p("")
+
+	if hasLRO {
+		p("// LROClient is used internally to handle longrunning operations.")
+		p("// It is exposed so that its CallOptions can be modified if required.")
+		p("// Users should not Close this client.")
+		p("LROClient *lroauto.OperationsClient")
+		p("")
+
+		g.imports[pbinfo.ImportSpec{Name: "lroauto", Path: "cloud.google.com/go/longrunning/autogen"}] = true
+	}
+
+	p("// The call options for this service.")
+	p("CallOptions *%sCallOptions", servName)
+	p("")
+
+	p("// The x-goog-* metadata to be sent with each request.")
+	p("xGoogMetadata metadata.MD")
+	p("}")
+	p("")
+
+	g.imports[pbinfo.ImportSpec{Path: "google.golang.org/grpc"}] = true
+	g.imports[pbinfo.ImportSpec{Path: "google.golang.org/grpc/metadata"}] = true
+
+	g.grpcClientUtilities(serv, servName, imp, hasLRO)
+}
+
+func (g *generator) grpcClientUtilities(serv *descriptor.ServiceDescriptorProto, servName string, imp pbinfo.ImportSpec, hasLRO bool) {
+	p := g.printf
+
+	clientName := camelToSnake(serv.GetName())
+	clientName = strings.Replace(clientName, "_", " ", -1)
+
+	// Factory function
+	p("// New%sClient creates a new %s client.", servName, clientName)
+	p("//")
+	g.comment(g.comments[serv])
+	p("func New%[1]sClient(ctx context.Context, opts ...option.ClientOption) (*%[1]sClient, error) {", servName)
+	p("  clientOpts := default%sClientOptions()", servName)
+	p("")
+	p("  if new%sClientHook != nil {", servName)
+	p("    hookOpts, err := new%sClientHook(ctx, clientHookParams{})", servName)
+	p("    if err != nil {")
+	p("      return nil, err")
+	p("    }")
+	p("    clientOpts = append(clientOpts, hookOpts...)")
+	p("  }")
+	p("")
+	p("  disableDeadlines, err := checkDisableDeadlines()")
+	p("  if err != nil {")
+	p("    return nil, err")
+	p("  }")
+	p("")
+	p("  connPool, err := gtransport.DialPool(ctx, append(clientOpts, opts...)...)")
+	p("  if err != nil {")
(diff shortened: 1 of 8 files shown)


software-dov:
Note: the diff seems to have picked up other changes that have already been made due to fat fingering git commands. I can disambiguate what changes are part of the PR and which are not.

showcase/go.mod
```diff
@@ -11,3 +11,5 @@ require (
 	google.golang.org/genproto v0.0.0-20201201144952-b05cb90ed32e
 	google.golang.org/grpc v1.33.2
 )
+
+replace github.com/googleapis/gapic-showcase => ./gen/github.com/googleapis/gapic-showcase
```
noahdietz, line 15: Please revert this change - `make clean` will do so. :)

internal/gengapic/client_init.go
```diff
@@ -135,155 +135,236 @@ func (g *generator) clientOptions(serv *descriptor.ServiceDescriptorProto, servN
...
-		p("")
+	for _, v := range g.opts.transports {
+		switch v {
+		case grpc:
+			g.grpcClientInit(serv, servName, imp, hasLRO)
+		case rest:
+			// TODO(dovs): add rest client struct initialization
+			continue
+		default:
+			return fmt.Errorf("unexpected transport variant: %d", v)
```
noahdietz, line 366: Nit: maybe added supported values to the error message?

internal/gengapic/client_init.go
```diff
@@ -135,155 +135,236 @@ func (g *generator) clientOptions(serv *descriptor.ServiceDescriptorProto, servN
...
-
-	var hasLRO bool
+	p("// %sAbstractClient is an interface that defines the methods availaible from %s.", servName, g.apiName)
+	p("// It is agnostic as to the underlying transport, i.e. json+http, gRPC, or other.")
+	p("type %sAbstractClient interface {", servName)
 	for _, m := range serv.Method {
-		if g.isLRO(m) {
-			hasLRO = true
-			break
+		inType := g.descInfo.Type[*m.InputType]
```
noahdietz, line 144: Nit: Please use the field accessor instead of direct access.
```suggestion
		inType := g.descInfo.Type[m.GetInputType()]
```

internal/gengapic/client_init.go
```diff
@@ -135,155 +135,236 @@ func (g *generator) clientOptions(serv *descriptor.ServiceDescriptorProto, servN
...
+		inSpec, err := g.descInfo.ImportSpec(inType)
+		if err != nil {
+			return err
+		}
+		outType := g.descInfo.Type[*m.OutputType]
```
noahdietz, line 149: Nit: Please use the field accessor instead of direct access.
```suggestion
		outType := g.descInfo.Type[m.GetOutputType()]
```

noahdietz (commented):
Looks great, @software-dov ! Let's have @codyoss review as well.

Generally, I wonder if we should have a `regapic-devel` branch to merge progress into...merging things into master piece-meal means if we need to cut a generator release for something else, we start generating fragments of REGAPIC support.

vchudnov-g:
I'm planning on TAL tomorrow, but on the branch issue I suggest either 

(a) Being very careful to make sure any PRs against master do not alter usage of the generator today. That should be the case given that we shouldn't be generating REST transport without the `--transport=rest` flag
(b) Having PRs that get merged into a branch, as Noah suggests. This would enable us to turn on everything at once, but we can drift from master, which means either frequent rebases or dealing with merge conflicts. (That said, I guess master isn't changing a whole lot)

I'm fine with (a) given the `transport` flag, and suspect it may make things simpler. Happy to discuss. (And no strong feelings; both options seem viable.)


[On a separate note, fyi, what I switched to doing on Showcase is uploading my work to my development branch on the main repo (rather than my fork) and generating PRs from that. This allows dependent PRs (which can't happen across forks). Something to think about, but I don't know whether it's needed, given quick review turnarounds. I came close to using a dependent PR once, but didn't need to in the end.]

noahdietz:
Well my concrete concern is that this PR introduces this new AbstractClient interface. If we were to merge this, then say fix a bug elsewhere and cut a release, suddenly we are generating this new interface in all the clients. Once it is out there, we cannot take it back. Users may start coding against the interface and then 1. when an API adds a new method, we break them (because a new method is added to the interface they were attempting to implement when we hadn't fully rolled out the point of this feature) 2. we may want to change this idea at some point (e.g. the name of it), and then we break users again. 

We just have to be very deliberate about introducing new, external types because they are forever once released.

vchudnov-g:
Fair enough. I'm fine with working on a branch until things are more solid.

internal/gengapic/client_init.go
```diff
@@ -135,155 +135,236 @@ func (g *generator) clientOptions(serv *descriptor.ServiceDescriptorProto, servN
 	return nil
 }
 
-func (g *generator) clientInit(serv *descriptor.ServiceDescriptorProto, servName string) error {
+func (g *generator) abstractClientIntfInit(serv *descriptor.ServiceDescriptorProto, servName string) error {
 	p := g.printf
-
-	var hasLRO bool
+	p("// %sAbstractClient is an interface that defines the methods availaible from %s.", servName, g.apiName)
```
codyoss, line 140: Our approach for the Go client libraries thus far has been to avoid interfaces, especially public one. If a new method `Zap` got added this is technically a breaking change to the interface. I don't think the outside world needs to know about this though, so could we just make the interface private. Only the package itself would ever use this, right?

codyoss: nit: maybe just `fooClient` vs `fooAbstractClient`. All interfaces are abstract and the inclusion of the word feels strongly OOP/Java

internal/gengapic/client_init.go
```diff
@@ -135,155 +135,236 @@ func (g *generator) clientOptions(serv *descriptor.ServiceDescriptorProto, servN
...
+func (g *generator) grpcClientInit(serv *descriptor.ServiceDescriptorProto, servName string, imp pbinfo.ImportSpec, hasLRO bool) {
+	p := g.printf
 
 	// client struct
-	{
-		p("// %sClient is a client for interacting with %s.", servName, g.apiName)
-		p("//")
-		p("// Methods, except Close, may be called concurrently. However, fields must not be modified concurrently with method calls.")
-		p("type %sClient struct {", servName)
+	p("// %sClient is a client for interacting with %s.", servName, g.apiName)
```
vchudnov-g, line 208: Consider adding "...over gRPC transport."
Maybe also "It satisfies the XXX interface"

internal/gengapic/client_init.go
```diff
@@ -135,155 +135,236 @@ func (g *generator) clientOptions(serv *descriptor.ServiceDescriptorProto, servN
...
+	p("")
+
+	p("// The gRPC API client.")
+	p("%s %s.%sClient", grpcClientField(servName), imp.Name, serv.GetName())
+	p("")
 
-		p("// The gRPC API client.")
-		p("%s %s.%sClient", grpcClientField(servName), imp.Name, serv.GetName())
+	if hasLRO {
+		p("// LROClient is used internally to handle longrunning operations.")
```
vchudnov-g, line 226: nit: `-` in "long-running"

internal/gengapic/client_init.go
```diff
@@ -135,155 +135,236 @@ func (g *generator) clientOptions(serv *descriptor.ServiceDescriptorProto, servN
...
+		}
+
+		if pf, err := g.pagingField(m); err != nil {
+			return err
+		} else if pf != nil {
+			iter, err := g.iterTypeOf(pf)
+			if err != nil {
+				return err
+			}
+			p("%s(context.Context, *%s.%s, ...gax.CallOption) *%s",
```
vchudnov-g, line 170: Since the idea is for apps developers to code against the interface, maybe it's worth generating a brief comment for each method in the generated interface? (matching the method comments in the concrete clients?). WDYT? (We can defer this to a later PR)

vchudnov-g (commented):
Looking good!

Respondinging to @codyoss :
* wdym about the public interface? I think the idea here is for app developers to be able to code almost exclusively against the interface, after getting a concrete gRPC or REST client. So the interface needs to be public, right?
* On the naming of the interface, I agree with your sentiment, but I know one of the issues is that for backwards compatibility with existing apps, we need the concrete gRPC client to remain named as it is.

internal/gengapic/client_init.go
codyoss: @vchudnov-g 

> wdym about the public interface? I think the idea here is for app developers to be able to code almost exclusively against the interface, after getting a concrete gRPC or REST client. So the interface needs to be public, right?

Why could they not keep coding against the concrete struct there is today and we just handle what transport is used under the hood? Like I said above, adding a method to an interface is a breaking change whereas adding one to a struct is not. This is why the clients did not provide an interface to begin with. In Go the general sentiment is if you want an interface for a struct you should declare it in your(the caller code). Also that functions should be returning concrete types and accepting interfaces. This is all done(mainly) for the sake of compatibility.

software-dov:
I'm generally more in favor of feature toggles over dev/feature branches because it makes smaller, easier merges more fluid. The commandline flag itself makes for a straightforward toggle, and we can hide Rest behind some early alpha/WIP warnings. If other people are unconvinced I'll bow to the consensus.

internal/gengapic/client_init.go
software-dov: Doh!

internal/gengapic/client_init.go
software-dov: +1 to both.

internal/gengapic/client_init.go
software-dov: I'm definitely in favor of that. To the extent that the code is readable it will be used as de facto documentation. As responsible engineers we should lean into that.

internal/gengapic/client_init.go
software-dov: You make a good point about adding a method being a technically breaking change, but IIUC this doesn't have serious impact in this particular instance. We're not expecting our users to define structs that implement the interface, and (both) structs in the generated library are structs that we control. If the API adds a `Zap` method, the gRPC client, the rest client, and the interface would all be updated simultaneously during regeneration.

Victor proposed a couple of usecases where application code does want the client/transport types to be polymorphic. Runtime transport choice was one, as well as dynamic fallback with one client of each kind. In both of these cases the user could definitely make their own interface to wrap both structs and obviate us from making this change. That feels like an unnecessary duplication of effort, though.

The workaround that Jean de Klerk mentions in [this post](https://blog.golang.org/module-compatibility) seems like overkill for our usecase here.

As for the `fooAbstractClient` naming convention, the `fooClient` moniker is already taken by the gRPC client. Reappropriating the name for the interface would itself be a breaking change.

The client struct as it is today is heavily gRPC centric. It seemed like it would be a lot of extra effort to partition off the gRPC fields, add rest fields, and maintain state that determined which kind of client our instance was. For Python GAPIC, the client receives (or creates) an injectable transport, but reworking that for golang GAPIC would be another breaking change.

showcase/go.mod
software-dov: Doh!

internal/gengapic/client_init.go
noahdietz: Personally, this feels like something we need to discuss in a doc/meeting. It doesn't feel like we are all on the same page with how the generated surface should change to support REGAPIC and how we can do that within best practices.

internal/gengapic/client_init.go
software-dov: Done

internal/gengapic/client_init.go
software-dov: Done

internal/gengapic/client_init.go
software-dov: Done

internal/gengapic/client_init.go
software-dov: I made a design doc a while ago describing alternatives and tradeoffs. I'll send it around for another round of scrutiny.

vchudnov-g: I'm fine with meeting to discuss this. A couple of points to bring up:
- as Dov said, we're not expecting users to implement the interface, but just to use it. When our generators change the interface, they'll be changing its concrete implementations as well.
- it just occurred to me that maybe we also have the alternative of using embedded structs (I haven't thought this out fully). In that case, the abstract interface becomes an abstract struct with dummy methods that is embedded in the concrete structs. The concrete structs can define their own methods with the same names as the abstract struct, thus shadowing them. Any concrete method not such defined will fall back to the abstract struct's.  That said, I still prefer the interface approach, which seems simpler.

But yes, let's discuss! Certainly having the transport selection/use happen under the hood is another way to approach the same goal, and one that other languages have been using. I'm open to that as well; let's figure out what that could look like and whether it's preferable.

internal/gengapic/client_init.go
```diff
@@ -135,155 +135,238 @@ func (g *generator) clientOptions(serv *descriptor.ServiceDescriptorProto, servN
...
+	p("")
+	p("    %s: %s.New%sClient(connPool),", grpcClientField(servName), imp.Name, serv.GetName())
+	p("  }")
+	p("  c.setGoogleClientInfo()")
+	p("")
+
+	if hasLRO {
+		p("  c.LROClient, err = lroauto.NewOperationsClient(ctx, gtransport.WithConnPool(connPool))")
 		p("  if err != nil {")
+		p("    // This error \"should not happen\", since we are just reusing old connection pool")
```
miraleung, line 294: For my Golang edification: When no variables are needed, what's the motivation for using single-line string values instead of a multi-line literal?

internal/gengapic/gengapic.go
```diff
@@ -70,6 +70,19 @@ type options struct {
...
 }
 
+// Utility function for stringifying the Transport enum
+func (t Transport) String() string {
+	switch t {
+	case grpc:
+		return "grpc"
+	case rest:
+		return "rest"
+	default:
```
miraleung, line 80: If we currently don't have other transport types and don't plan to support arbitrary ones, can we return one of the defined values by default?

vchudnov-g: This is a string representation of the current value, so unless we change the value to be valid (which we should do elsewhere, when we initially read it from the flag), I would keep the current code as correct.

internal/gengapic/client_init.go
vchudnov-g: This is just the convention we've been using in the Go generator and in Showcase since the beginning. If we were starting from scratch, I'd certainly support multi-line literals, especially if we used them with the dedent package. Given the convention, though, I suggest not changing things right now. (Possibly in a separate, dedicated PR, if we decide it's worth the effort.)

BTW, multi-line literals can have variables, though since they're not labeled inline it gets confusing fast when the format specifier is far removed from its corresponding parameter.

software-dov:
Closing this PR and republishing in order to clean up commit history.

PR #523, opened 2021-02-03
gapic: rename gapic-service-config to api-service-config

noahdietz:
This renames a flag used to supply the API Service config file path from `gapic-service-config` to `api-service-config`. This is a much more accurate name. It keeps support for `gapic-service-config` by falling through to `api-service-config` in the parsing logic, but we should remove it before we make this repo go to v1.0.0.

This also removes the deprecation notice from the `api-service-config` flag, because we have further uses for it now.


codyoss (approved):
LGTM

PR #524, opened 2021-02-04
gapic: refactor to use generated Service type

noahdietz:
Long ago, as a workaround during generator migration, I used a hand-spun type to represent the Service config proto in its very limited use in the micro-generator. Going forward we will have more use of the Service config so I refactored that hand-spun code to use the published, generated Go type for the Service config.


codyoss (approved):
LGTM

PR #527, opened 2021-02-09
gapic: add gapic_metadata.json generation

noahdietz:
Adds generation of  the `gapic_metadata.json` file containing the proto-to-client mapping of the generated code.

The schema for GAPIC Metadata is defined in [googleapis](https://github.com/googleapis/googleapis/blob/master/gapic/metadata/gapic_metadata.proto).

Enable generation via the plugin option `metadata`. See README for more info.

Note: I needed to override the `go-genproto` dependency because `rules_go` brings an older version that is missing the `GapicMetadata` proto. See #529.

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,7 +82,7 @@ The configuration supported by the plugin option includes:
   * `module`: prefix to be stripped from the `go-gapic-package` used in the generated filenames.
      * _Note: This option is not supported from the Bazel interface._
 
-  * `metadata`: enable generation of [GapicMetadata](https://github.com/googleapis/googleapis/blob/master/gapic/metadata/gapic_metadata.proto) in JSON form.
+  * `metadata`: enable generation of [GapicMetadata](https://github.com/googleapis/googleapis/blob/master/gapic/metadata/gapic_metadata.proto) in JSON form. The default is `false`.
   
   * `grpc-service-config`: the path to a gRPC ServiceConfig JSON file.
     * This is used for client-side retry configuration in accordance with [AIP-4221](http://aip.dev/4221)
@@ -161,7 +161,7 @@ following attributes:
 
   * `sample_only`: if present, directs the generator to forgo client generation and generate only samples.
 
-  * `metadata`: if `True`, [GapicMetadata](https://github.com/googleapis/googleapis/blob/master/gapic/metadata/gapic_metadata.proto) will be generated in JSON form.
+  * `metadata`: if `True`, [GapicMetadata](https://github.com/googleapis/googleapis/blob/master/gapic/metadata/gapic_metadata.proto) will be generated in JSON form. The default is `False`.
 
 Docker Wrapper
 --------------
--- a/README.md
+++ b/README.md
@@ -81,6 +81,8 @@ The configuration supported by the plugin option includes:
 
   * `module`: prefix to be stripped from the `go-gapic-package` used in the generated filenames.
      * _Note: This option is not supported from the Bazel interface._
+
+  * `metadata`: enable generation of [GapicMetadata](https://github.com/googleapis/googleapis/blob/master/gapic/metadata/gapic_metadata.proto) in JSON form.
   
   * `grpc-service-config`: the path to a gRPC ServiceConfig JSON file.
     * This is used for client-side retry configuration in accordance with [AIP-4221](http://aip.dev/4221)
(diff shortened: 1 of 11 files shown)


tbpg (approved):

internal/gengapic/metadata.go
```diff
@@ -0,0 +1,52 @@
...
+
+package gengapic
+
+import (
+	"google.golang.org/genproto/googleapis/gapic/metadata"
+	"google.golang.org/protobuf/encoding/protojson"
+)
+
+func (g *generator) genGapicMetadataFile() error {
+	data, err := protojson.MarshalOptions{Multiline: true}.Marshal(g.metadata)
```
codyoss, line 23: This is fine for now, but if it becomes an issue down the road we may want to stabilize this output to reduce diffs.

noahdietz: I _think_ this writes maps by the keys in lexicographical order. But agreed, the public docs state not to depend on stable output.  Let's keep an eye on it as you've suggested.

README.md
```diff
@@ -81,6 +81,8 @@ The configuration supported by the plugin option includes:
 
   * `module`: prefix to be stripped from the `go-gapic-package` used in the generated filenames.
      * _Note: This option is not supported from the Bazel interface._
+
+  * `metadata`: enable generation of [GapicMetadata](https://github.com/googleapis/googleapis/blob/master/gapic/metadata/gapic_metadata.proto) in JSON form.
```
vchudnov-g, line 85: Should we say that the default is `false`?

README.md
```diff
@@ -159,6 +161,8 @@ following attributes:
 
   * `sample_only`: if present, directs the generator to forgo client generation and generate only samples.
 
+  * `metadata`: if `True`, [GapicMetadata](https://github.com/googleapis/googleapis/blob/master/gapic/metadata/gapic_metadata.proto) will be generated in JSON form.
```
vchudnov-g, line 164: Mention default?

internal/gengapic/options.go
```diff
@@ -72,6 +74,9 @@ func parseOptions(parameter *string) (*options, error) {
 		// check for the boolean flag, sample-only, that disables client generation
 		if s == "sample-only" {
 			return &options{sampleOnly: true}, nil
+		} else if s == "metadata" {
```
vchudnov-g, line 77: We can remove the `else` and outdent the `if` statement, right? Seems more idiomatic and clearer

rules_go_gapic/go_gapic.bzl
```diff
@@ -86,6 +86,7 @@ def go_gapic_library(
   gapic_yaml = None,
   samples = [],
   sample_only = False,
+  metadata = False,
```
vchudnov-g, line 89: Cool. We'll have to remember to set this to true in our standard publication pipeline. Is there an issue to track this?

vchudnov-g (approved):
Looks good!. Thanks for doing this!

I only have very minor comments.

README.md
noahdietz: Yep, done!

README.md
noahdietz: Yep, done!

internal/gengapic/options.go
noahdietz: Sure thing. Done!

rules_go_gapic/go_gapic.bzl
noahdietz: Good idea. Here is one for `build_gen` https://github.com/googleapis/rules_gapic/issues/10 and another in https://github.com/googleapis/google-cloud-go/issues/3696 (where we have the production pipeline).

internal/gengapic/metadata_test.go
```diff
@@ -0,0 +1,172 @@
...
+		service, lib string
+		init, want   *metadatapb.GapicMetadata
+	}{
+		{
+			service: "LibraryService",
+			lib:     "LibraryServiceClient",
+			init: &metadatapb.GapicMetadata{
+				Services: make(map[string]*metadatapb.GapicMetadata_ServiceForTransport),
+			},
+			want: &metadatapb.GapicMetadata{
```
vam-google, line 37: minor: I'm not expert on Go by any means, but I have to say this has way too many indentations =) Is it ok/idiomatic for go? Maybe it can be reduced (number of indentations) by introducing intermediate variables somewhere?

vam-google (approved):
LGTM

internal/gengapic/metadata_test.go
vchudnov-g: lol.... This is totally idiomatic for Go. I guess I've gotten used to this pattern and actually kinda like it.

noahdietz: I understand where you are coming from, and this specific case deep indention is normal. This is a Table Driven Test (https://github.com/golang/go/wiki/TableDrivenTests), but one with the "row" type declared in the `range` statement. What makes this worse is the many nested messages we have to initialize :)

PR #528, opened 2021-02-09
bazel: update Go deps from go.mod

noahdietz:
Update `repositories.bzl` with `go.mod` via `make update-bazel-repos`


codyoss (approved):
LGTM

PR #530, opened 2021-02-12
bazel: ignore metadata.srcjar if disabled

noahdietz:
Don't attempt to unzip the metadata srcjar if metadata generation was disabled as the archive will be empty and produce an error.


vam-google (approved):
LGTM

tritone (approved):

PR #552, opened 2021-03-03
gapic: generate mix-in API methods on client libraries

noahdietz:
This adds support for client library mixin method generation. The (draft) AIP is at https://github.com/aip-dev/google.aip.dev/pull/694 and has the high level context. Ping me for the internal doc(s) if desired.

The supported mixin services are:
- `google.longrunning.Operations`
- `google.iam.v1.IAMPolicy`
- `google.cloud.location.Locations`

The proto descriptors for these services are loaded from their compiled Go types, fed into `pbinfo` for lookup availability, and included in generation of each product service client. The mixin API gRPC stubs (i.e. `IAMPolicyClient`) are added to the gapic client struct as an unexported field and initialized with the connection established with the product service endpoint (`i.e. pubsub.googleapis.com`). The mixin methods get all of the GAPIC features - auto-pagination, client metadata, default deadline & retry settings (if configured), request header params, client option support, etc.

Preview of Showcase [client regeneration](https://github.com/googleapis/gapic-showcase/pull/649) with these changes (as of first review request).


internal/gengapic/mixins.go
```diff
@@ -0,0 +1,118 @@
...
+func getLocationsDescriptor() *descriptor.FileDescriptorProto {
+	return mixinFiles["google.cloud.location.Locations"][0]
+}
+
+func getLocationsMethods() []*descriptor.MethodDescriptorProto {
+	return getLocationsDescriptor().GetService()[0].GetMethod()
+}
+
+func getIAMPolicyDescriptors() []*descriptor.FileDescriptorProto {
+	return mixinFiles["google.iam.v1.IAMPolicy"]
```
codyoss, line 105: nit: add `[0]` for consistency with other getters

codyoss (approved):
LGTM

internal/gengapic/client_init.go
```diff
@@ -138,10 +142,10 @@ func (g *generator) clientOptions(serv *descriptor.ServiceDescriptorProto, servN
 func (g *generator) clientInit(serv *descriptor.ServiceDescriptorProto, servName string) error {
 	p := g.printf
 
-	var hasLRO bool
+	var hasLRORpc bool
```
tbpg, line 145: It's not great, but this should be `hasLRORPC`. Perhaps `hasRPCForLRO`?

https://github.com/golang/go/wiki/CodeReviewComments#initialisms

internal/gengapic/client_init_test.go
```diff
@@ -147,13 +151,34 @@ func TestClientOpt(t *testing.T) {
 		t.Fatal(err)
 	}
 
+	servIamOverride := &descriptor.ServiceDescriptorProto{
```
tbpg, line 154: IAM

internal/gengapic/mixins.go
noahdietz: Ah that's the thing, this one returns multiple proto files, while the others are just one proto file.

tbpg (approved):
A couple super minor comments.

internal/gengapic/client_init.go
noahdietz: sgtm! Done.

internal/gengapic/client_init_test.go
noahdietz: Done.

internal/gengapic/mixins.go
codyoss: Is this function used outside this file? What I was trying to say I guess is the other methods all just do return a single item. Noticed that this is the only one that returned a list. The get method calls in the rest just operate on the descriptor directly([0] was already called when getting descriptor)

noahdietz: It's not used outside this file, no. I see what you are saying, yeah it's a little weird. I'll change it up.

noahdietz: Done! Thanks for the explanation, Cody.

noahdietz:
I confirmed locally that regeneration of google-cloud-go had no unexpected changes - the config files haven't been updated to enable this feature yet.

PR #564, opened 2021-03-15
gapic: fix link in ctx docs

### LINKED ISSUE #563: gapic: generated Use of Context pkg docs should refer to pkg.go.dev
The generated `Use of Context` documentation should use `pkg.go.dev` instead of `godoc.org` ([here](https://github.com/googleapis/gapic-generator-go/blob/93d841c55eb48819d78cd65c78814074fe6dc430/internal/gengapic/doc_file.go#L74)) and the URL should have the proper scheme in order to be hyperlinked by godoc.

noahdietz:
The godoc link was fixed in #412 but it was still missing `https://` which means godoc doesn't render it as a hyperlink.

Fixes #563


tbpg (approved):
Thanks!

PR #565, opened 2021-03-15
gapic: parse markdown in pkg doc string

### LINKED ISSUE #562: gapic: client pkg docs should be run through the markdown parser
The `doc.go` package documentation pulled from the API Service config ([here](https://github.com/googleapis/gapic-generator-go/blob/93d841c55eb48819d78cd65c78814074fe6dc430/internal/gengapic/doc_file.go#L43-L52)) should be run through the markdown parser.

[Some APIs](https://pkg.go.dev/cloud.google.com/go@v0.79.0/monitoring/apiv3#pkg-overview) include markdown links in such documentation that should be handled.

noahdietz:
Some APIs include markdown in their API Service config documentation summary section - the field that is used as the Go package documentation. It should be parsed for markdown just like the proto RPC comments.

Fixes #562

### DIFF
--- a/internal/gengapic/doc_file.go
+++ b/internal/gengapic/doc_file.go
@@ -41,7 +41,9 @@ func (g *generator) genDocFile(year int, scopes []string) {
 	}
 
 	if g.serviceConfig != nil && g.serviceConfig.GetDocumentation() != nil {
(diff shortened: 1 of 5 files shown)


tbpg (approved):
Thanks!

PR #577, opened 2021-03-31
feat: add interface abstraction primitives

software-dov:
Nomenclature within this commit:
* The current generated client is referred to as a 'transport'.
  Similarly, the WIP rest+json client will also be a 'transport'.
* The new internal-only interface is the 'transport interface'.
* The new client type whose only responsibility is to embed a
transport interface is a 'new-style client'.

Two new types are defined: an interface intended to abstract out the
mechanism of network communication and serialization (the transport interface),
and a wrapper struct that embeds the interface.

The transport interface cannot be user visible because methods will be
added to it if the API defines new methods. This is a non-breaking
change in the API but is technically a breaking change to the user; if
they defined a type that previously satisfied the contract of the
interface, the type would no longer satisfy the interface.

The workaround is to make the transport interface not user visible and
to embed the interface in a struct. The transports and transport
interface are generated concurrently, so there is no risk of breaking
the interface contract. Adding methods to the new-style client is a
non-breaking change becasuse it is a struct.
This is similar to the PIMPL idiom in C++.

TODO:
* Add factory function(s) for the new-style client to take an injected
transport + auth.
* Tweak documentation, samples, and integration tests to prefer
new-style clients instead of directly using transports.
* Write out rest+json transport.

Includes moderately substantial refactors as implementation details.


noahdietz:
Is this truly ready for review? Should we make this a draft PR for the time being?

software-dov:
Yes, this is complete module the test failures. The TODOs are caveat notes about subsequent PRs.

internal/gengapic/gengapic.go
```diff
@@ -43,7 +43,106 @@ const (
...
+// * release-level (one of 'alpha', 'beta', or empty)
+// * transport ('+' separated list of transport backends to generate)
+// The only required option is 'go-gapic-package'.
+//
+// Valid parameter example:
+// go-gapic-package=path/to/out;pkg,module=path,transport=rest+grpc,gapic-service-config=gapic_cfg.json,release-level=alpha
+//
+// It returns a pointer to a populated options if no errors were encountered while parsing.
+// If errors were encountered, it returns a nil pointer and the first error.
+func ParseOptions(parameter *string) (*options, error) {
```
noahdietz, line 64: There is already  parseOptions method in options.go. Why was this added?

internal/gengapic/gengapic_test.go
```diff
@@ -523,6 +524,91 @@ func TestIsLRO(t *testing.T) {
 	}
 }
 
+func Test_optionsParse(t *testing.T) {
```
noahdietz, line 527: same with this, there is a test for this already in options_test.go, no?

internal/gengapic/client_init.go
```diff
@@ -135,155 +135,244 @@ func (g *generator) clientOptions(serv *descriptor.ServiceDescriptorProto, servN
...
+	p := g.printf
 
 	// client struct
-	{
-		p("// %sClient is a client for interacting with %s.", servName, g.apiName)
-		p("//")
-		p("// Methods, except Close, may be called concurrently. However, fields must not be modified concurrently with method calls.")
-		p("type %sClient struct {", servName)
+	p("// %sClient is a client for interacting with %s over gRPC transport.", servName, g.apiName)
+	p("// It satisfies the %sAbstractClient interface.", servName)
```
noahdietz, line 215: Probably don't want to document the private interface (and also there is no AbstractClient interface now?).

noahdietz:
> Yes, this is complete module the test failures. The TODOs are caveat notes about subsequent PRs.

Ok! You can also download the linting report from the failing GitHub Actions. It includes govet, golint and gofmt.

internal/gengapic/testdata/empty_client_init.want
```diff
@@ -1,4 +1,16 @@
-// Client is a client for interacting with Awesome Foo API.
+// internalClient is an interface that defines the methods availaible from Awesome Foo API.
+type internalClient interface {
+	Zip(context.Context, *mypackagepb.Bar, ...gax.CallOption) (*mypackagepb.Foo, error)
+}
+
+//  is a thin wrapper that holds an internalClient
+// It is agnostic as to the underlying transport, i.e. json+http, gRPC, or other.
+type  struct {
```
noahdietz, line 8: Need to handle case where `servName` is empty (the resulting client struct name is just "Client")

noahdietz: Can you explain again why we want this struct?

software-dov: We want to provide some means of polymorphic abstraction over grpc/rest+json. In the last review, Cody pointed out that adding methods to a public interface is a breaking change, even though the interface and the two implementing structs are generated concurrently and therefore guaranteed to be in sync.
The workaround is to make the interface private and make the public type a struct. Adding methods to a struct is not breaking, and the struct just embeds the interface.

To use the monikers from Python, the existing client is a 'transport', for which there exists an ~abstract base class~ interface. The new wrapper struct is the Python 'client' equivalent.

software-dov:
> You can also download the linting report from the failing GitHub Actions

That is incredibly useful. Wow, I wish I'd known this when doing the python m-gen CI conversion. Thanks!

internal/gengapic/gengapic.go
software-dov: Transcription/merge error on my part that I'm still not sure how it happened. Good eye.

internal/gengapic/client_init.go
software-dov: Is not documenting internal structs/interfaces in the generated surface the norm? This is (mostly) to aid my own eyeball grepping when debugging generated clients.

internal/gengapic/testdata/empty_client_init.want
software-dov: Changed the naming convention away from `p("%s", serv.name)` to `p("%sClientHandle", serv.name)`

noahdietz: Ok let me see if I understand this. A new factory method like `NewFooClientHandle () FooClientHandle` where `FooClientHandle` is 
```go
type FooClientHandle struct {
  internalFooInterface
}
```

`NewFooClientHandle` would return a `FooClientHandle` backed by either the `FooRestClient` or `FooClient` (grpc). Is that right?

_Ignore naming and stuff for now, that is easily rectified._

software-dov: Exactly. The factory function signature(s) and implementation (s) are the subject of a subsequent PR.

noahdietz:
Would you mind opening a draft PR on gapic-showcase with the client regenerated using your changes? Like  https://github.com/googleapis/gapic-showcase/pull/649. This will be useful for reviewing the changes in a "real" client.

In gapic-generator-go, run `make install`. Then in gapic-showcase, run `go run ./util/cmd/compile_protos`, only check in the changes to `client/` and open it as a Draft PR please.

software-dov:
Done. showcase draft PR is here: https://github.com/googleapis/gapic-showcase/pull/652/files?file-filters%5B%5D=.go

internal/gengapic/client_init.go
noahdietz: Well, for one, there is no `*AbstractClient` type defined anywhere, so I wouldn't be documenting that :)

showcase/go.mod
```diff
@@ -11,3 +11,5 @@ require (
 	google.golang.org/genproto v0.0.0-20201201144952-b05cb90ed32e
 	google.golang.org/grpc v1.33.2
 )
+
+replace github.com/googleapis/gapic-showcase => ./gen/github.com/googleapis/gapic-showcase
```
noahdietz, line 15: Please revert this. Running `make clean` is a good thing to do before pushing changes.

software-dov: The `go.mod` and `go.sum` changes sneak in under the radar. Would it be correct to add them to `.gitignore`?

noahdietz: I'd prefer that we don't add them to the `.gitignore` because this is a submodule and we really do want changes to this file to be captured...except for this one made by the testing script :)

You could add some command to the `showcase.bash` to revert this change once the tests are done though, so that `make clean` isn't necessary.

internal/gengapic/client_init.go
```diff
@@ -139,193 +139,405 @@ func (g *generator) clientOptions(serv *descriptor.ServiceDescriptorProto, servN
...
+	p := g.printf
 
-			g.imports[pbinfo.ImportSpec{Name: "longrunningpb", Path: "google.golang.org/genproto/googleapis/longrunning"}] = true
+	// Close
+	p("func (c *%sClient) Close () error {", servName)
+	p("    return c.internalClient.Close()")
+	p("}")
+	p("")
+
+	// Need a thin wrapper around
```
noahdietz, line 250: Nit: finish this comment please.

internal/gengapic/client_init.go
```diff
@@ -139,193 +139,405 @@ func (g *generator) clientOptions(serv *descriptor.ServiceDescriptorProto, servN
...
+	for _, m := range serv.Method {
```
noahdietz, line 251: Nit: please use the accessor method for proto fields - `GetMethod()`

internal/gengapic/client_init.go
```diff
@@ -139,193 +139,405 @@ func (g *generator) clientOptions(serv *descriptor.ServiceDescriptorProto, servN
...
+	clientName := camelToSnake(serv.GetName())
+	clientName = strings.Replace(clientName, "_", " ", -1)
+	lowcaseServName := lowerFirst(servName)
+
+	// Factory function
+	p("// New%sClient creates a new %s client based on gRPC.", servName, clientName)
+	p("//")
+	g.comment(g.comments[serv])
+	p("func New%[1]sClient(ctx context.Context, opts ...option.ClientOption) (*%[1]sClient, error) {", servName)
+	p("  clientOpts := default%[1]sGrpcCallOptions()", servName)
```
noahdietz, line 420: I think we are conflating two different sets of options. One is the set of options provided to the factory function for use during client creation (specifically https://pkg.go.dev/google.golang.org/api/option#ClientOption) and call options used at call invocation (https://pkg.go.dev/github.com/googleapis/gax-go/v2#CallOption).

I think this should be `defaultClientOptions` ([this method](https://github.com/googleapis/google-cloud-go/blob/master/asset/apiv1/asset_client.go#L58-L68) used [here](https://github.com/googleapis/google-cloud-go/blob/master/asset/apiv1/asset_client.go#L191-L206)).

internal/gengapic/client_init.go
```diff
@@ -139,193 +139,405 @@ func (g *generator) clientOptions(serv *descriptor.ServiceDescriptorProto, servN
...
-			p("    // we would close a connection that's still in use.")
-			p("    // TODO: investigate error conditions.")
-			p("    return nil, err")
-			p("  }")
-		}
+	if g.hasLocationMixin() {
+		p("  c.locationsClient = locationpb.NewLocationsClient(connPool)")
+		p("")
+	}
+	p("  return &%sClient{internal%[1]sClient: c, CallOptions: clientOpts}, nil", servName)
```
noahdietz, line 474: `CallOptions` should be supplied the `defaultCallOptions` not the `ClientOption`s. They are two different things, as mentioned above.

internal/gengapic/client_init.go
noahdietz: If we want to separate the REST & gRPC default client options, another method may be necessary for that i.e. `defaultRestClientOptions` and `defaultGrpcClientOptions`. Although this is getting pretty duplicative and I wonder since `ClientOption` and `CallOption` are both transport agnostic interfaces, maybe we should just stuff all default options in one set and use for both 🤷🏻 . I'm ok with duplicating everything though if its safer because its just generated code.

internal/gengapic/client_init.go
```diff
@@ -139,193 +139,405 @@ func (g *generator) clientOptions(serv *descriptor.ServiceDescriptorProto, servN
...
+
+	for _, v := range g.opts.transports {
+		switch v {
+		case grpc:
+			g.grpcClientInit(serv, servName, imp, hasLRO)
+		case rest:
+			// TODO(dovs): add rest client struct initialization
+			continue
+		default:
+			return fmt.Errorf("unexpected transport variant (supported variants are '%s', '%s'): %d",
```
noahdietz, line 538:
```suggestion
			return fmt.Errorf("unexpected transport variant (supported variants are %q, %q): %d",
```

The `%q` format specifier prints strings wrapped in double quotes.

internal/gengapic/options.go
```diff
@@ -149,3 +150,16 @@ func parseOptions(parameter *string) (*options, error) {
 
 	return &opts, nil
 }
+
+// Utility function for stringifying the Transport enum
+func (t transport) String() string {
```
noahdietz, line 155: Nit: can we colocate methods with the type they belong to for ease of readability? Consider either moving the type def for `transport` down here or adding a new file `transport.go` (and a corresponding test file).

internal/gengapic/client_init.go
```diff
@@ -139,193 +139,405 @@ func (g *generator) clientOptions(serv *descriptor.ServiceDescriptorProto, servN
...
+			if err != nil {
+				return err
+			}
+			p("func (c *%sClient) %s(ctx context.Context, req *%s.%s, opts ...gax.CallOption) *%s {",
+				servName,
+				m.GetName(),
+				inSpec.Name,
+				inType.GetName(),
+				iter.iterTypeName)
+			p("    var opts gax.CallOption")
```
noahdietz, line 290: This is an unnecessary variable declaration, there is already a `opts gax.CallOption` defined as a parameter.

internal/gengapic/client_init.go
```diff
@@ -139,193 +139,405 @@ func (g *generator) clientOptions(serv *descriptor.ServiceDescriptorProto, servN
...
+		}
 
-			g.imports[pbinfo.ImportSpec{Name: "iampb", Path: "google.golang.org/genproto/googleapis/iam/v1"}] = true
+		switch {
+		case g.isLRO(m):
+			// Unary call where the return type is a wrapper of
+			// longrunning.Operation and more precise types
+			p("func (c *%sClient) %s(ctx context.Context, req *%s.%s, opts ...gax.CallOption) (*%s, error) {",
+				servName, m.GetName(), inSpec.Name, inType.GetName(), lroTypeName(m.GetName()))
+			p("    var opts gax.CallOption")
```
noahdietz, line 304: Same thing, unnecessary.

internal/gengapic/client_init.go
```diff
@@ -139,193 +139,405 @@ func (g *generator) clientOptions(serv *descriptor.ServiceDescriptorProto, servN
...
+			g.appendCallOpts(m)
+			p("    return c.internalClient.%s(ctx, req, opts)", m.GetName())
+			p("}")
+			p("")
+		case m.GetClientStreaming():
+			// Handles both client-streaming and bidi-streaming
+			p("func (c *%sClient) %s(ctx context.Context, opts ...gax.CallOption) (%s.%s_%sGrpcClient, error) {",
+				servName, m.GetName(), inSpec.Name, serv.GetName(), m.GetName())
+			p("    var opts gax.CallOption")
```
noahdietz, line 313: Same thing, unnecessary.

internal/gengapic/client_init.go
```diff
@@ -139,193 +139,405 @@ func (g *generator) clientOptions(serv *descriptor.ServiceDescriptorProto, servN
...
+			g.appendCallOpts(m)
+			p("    return c.internalClient.%s(ctx, opts)", m.GetName())
+			p("}")
+			p("")
+		case m.GetServerStreaming():
+			// Handles _just_ server streaming
+			p("func (c *%sClient) %s(ctx context.Context, req *%s.%s, opts ...gax.CallOption) (%s.%s_%sGrpcClient, error) {",
+				servName, m.GetName(), inSpec.Name, inType.GetName(), inSpec.Name, serv.GetName(), m.GetName())
+			p("    var opts gax.CallOption")
```
noahdietz, line 322: Same thing, unnecessary.

internal/gengapic/client_init.go
```diff
@@ -139,193 +139,405 @@ func (g *generator) clientOptions(serv *descriptor.ServiceDescriptorProto, servN
...
+			g.appendCallOpts(m)
+			p("    return c.internalClient.%s(ctx, req, opts)", m.GetName())
+			p("}")
+		default:
+			// Regular, unary call
+			p("func (c *%sClient) %s(ctx context.Context, req *%s.%s, opts ...gax.CallOption) (*%s.%s, error){",
+				servName, m.GetName(), inSpec.Name, inType.GetName(), outSpec.Name, outType.GetName())
+			p("    var opts gax.CallOption")
```
noahdietz, line 330: Same thing, unnecessary.

noahdietz:
Looking pretty good, @software-dov. You made some good changes.

internal/gengapic/client_init.go
software-dov: TIL

internal/gengapic/client_init.go
software-dov: Good catch, you're absolutely right. I think it's better to keep the the `ClientOption` structs separate, which means two methods.

software-dov: I'm leaning more in favor of using distinct structures/methods with distinct names to disambiguate. It's just easier to keep track of and say "I'm writing REST surface stuff now, I need to use names that contain the 'Rest' moniker".

internal/gengapic/client_init.go
```diff
@@ -139,193 +139,400 @@ func (g *generator) clientOptions(serv *descriptor.ServiceDescriptorProto, servN
...
 
-		if g.hasIAMPolicyMixin() {
+		if m.GetOutputType() == emptyType {
+			p("func (c *%sClient) %s(ctx context.Context, req *%s.%s, opts ...gax.CallOption) error {",
+				servName,
+				m.GetName(),
+				inSpec.Name,
+				inType.GetName())
+			g.appendCallOpts(m)
+			p("    c.internalClient.%s(ctx, req, opts)", m.GetName())
```
noahdietz, line 271: missing return statement

noahdietz (commented):
One other verification you might consider is regenerating google-cloud-go locally with your generator changes and then running the apidiff script across the repo to catch breaking changes on the _public_ surface.

From gapic-generator-go: `make install`

From google-cloud-go:
```
$ go run cloud.google.com/go/internal/gapicgen/cmd/genbot    -local    -only-gapics    -gocloud-dir=`pwd`
$ go run ./internal/apidiff -repo-metadata=internal/.repo-metadata-full.json -verbose
```

These will both take some time and probably set your computer on fire, but its a good check!

vchudnov-g:
I'm late to this party due to my unexpected OOO. I do have one general question: what's the reason for duplicating the methods in the outer client and propagating them to the `internalClient` rather than anonymously embedding `internalClient` and letting method calls fall through?

software-dov:
To answer @vchudnov-g: the `CallOptions` field must be present in the wrapper struct because the field is currently exported, forming a de facto API promise. The actual grpc invocation needs the call options; in this case, passing them in works better than duplicating the field during construction, because they're mutable. This means that unfortunately we have to have explicit boilerplate wrapping each method call and passing it to the corresponding method for the `internalClient`. On the flip side, it means that user code referencing the current Client struct would not be deprecated.

software-dov:
Having difficulty running the api diff tools. The `genbot` command is failing to compile the protos due to missing proto file dependencies, currently working on tweaking that. The other command seems to be fine.

vchudnov-g:
Ah, OK. I see the reasoning. I'm OK with it, but it seems to me there are also other options that would reduce the boilerplate:

* the outer `CallOptions` could be a pointer to the `internalClient.CallOptions`, assigned at the same time as the `internalClient`
* or instead and even better, I think, is having `EchoClient.internalClient.CallOptions` be a double deref `**EchoCallOptions` and be assigned at `internalClient` selection time via code to the effect of `EchoClient.internalClient.CallOptions = &EchoClient.CallOptions`. That way, any end-developer mutations to `EchoClient.CallOptions`, including assigning a pointer to different struct, will be automatically picked up.

WDYT?

software-dov:
@vchudnov-g 
Option 1 was something I was specifically concerned about that led to the current design. I don't think it's a tenable design for the reasons you mention in Option 2: if the user changes the target of `CallOptions` in their client struct, the internal client won't be updated, which will lead to unexpected behavior. I don't think this is a common or encouraged use case, but we have to be defensive.

I like Option 2. I wasn't thinking about double pointers, but you're right, it solves this particular problem exactly.

vchudnov-g:
Thanks for updating the Showcase demo of this. I think this may not be quite right; see [my comment](https://github.com/googleapis/gapic-showcase/pull/652#discussion_r615175688) on that PR

internal/gengapic/testdata/foo_example.want
```diff
@@ -8,7 +8,7 @@ func ExampleNewFooClient() {
 	_ = c
 }
 
-func ExampleFooClient_GetEmptyThing() {
+func ExampleFooGrpcClient_GetEmptyThing() {
```
noahdietz, line 11: I don't think we should be changing these. These are actually documentation examples, and the naming is important. Since the "public interface" is `FooClient`, the examples should correlate to `FooClient_GetEmptyThing`.

I think you can revert the example changes altogether for now since there is no transport-specific different, and because the gRPC client is internal.

software-dov: I was torn about this change in particular; eventually, the examples and documentation should clearly and unambiguously reflect the underlying transport being used. Having a convention of "the user visible default is gGRPC, anything with Rest in the name is Rest" is less than ideal.

internal/gengapic/testdata/lro_client_init.want
```diff
@@ -48,15 +70,18 @@ func NewFooClient(ctx context.Context, opts ...option.ClientOption) (*FooClient,
...
-		CallOptions: defaultFooCallOptions(),
-
 		fooClient: mypackagepb.NewFooClient(connPool),
+		CallOptions: &client.CallOptions,
 	}
 	c.setGoogleClientInfo()
 
+	client.internalFooClient = c
+
 	c.LROClient, err = lroauto.NewOperationsClient(ctx, gtransport.WithConnPool(connPool))
```
noahdietz, line 85: I mentioned this on the gapic-showcase draft PR, but I think the `LROClient` needs to be handled the same way as the `CallOptions` with the double pointer situation - right? If we expose LROClient on the wrapper, it should be initialized and set there (on the wrapper) and _referenced_ in the transport-client

internal/gengapic/testdata/empty_client_init.want
```diff
@@ -1,31 +1,48 @@
...
+	// The call options for this service.
+	CallOptions *CallOptions
+
+}
+
+// GrpcClient is a client for interacting with Awesome Foo API over gRPC transport.
+// It satisfies the internalClient interface.
+//
+// Methods, except Close, may be called concurrently. However, fields must not be modified concurrently with method calls.
+type GrpcClient struct {
```
tbpg, line 22: Nit: s/Grpc/GRPC/

internal/gengapic/testdata/empty_client_init.want
```diff
@@ -41,37 +58,40 @@ func NewClient(ctx context.Context, opts ...option.ClientOption) (*Client, error
 	if err != nil {
 		return nil, err
 	}
-	c := &Client{
+	client := Client{CallOptions: defaultCallOptions()}
```
tbpg, line 61: Seems like we could move this down right before the `return` to avoid mixing struct initialization here with assignment below.

tbpg, line 61: This appears to have type `Client`, but the return type is `*Client`. Should this line be `&Client`?

internal/gengapic/testdata/empty_client_init.want
```diff
@@ -1,31 +1,48 @@
...
+	// The internal transport-dependent client.
+	internalClient
+
+	// The call options for this service.
+	CallOptions *CallOptions
+
+}
+
+// GrpcClient is a client for interacting with Awesome Foo API over gRPC transport.
+// It satisfies the internalClient interface.
```
tbpg, line 19: No need for this comment. Users shouldn't have to know about `internalClient`.

internal/gengapic/testdata/empty_client_init.want
```diff
@@ -41,37 +58,40 @@ func NewClient(ctx context.Context, opts ...option.ClientOption) (*Client, error
...
-	return c, nil
+	client.internalClient = c
+
+	return client, nil
 }
 
 // Connection returns a connection to the API service.
 //
 // Deprecated.
-func (c *Client) Connection() *grpc.ClientConn {
```
tbpg, line 59: This appears to be a breaking change.

internal/gengapic/testdata/empty_client_init.want
```diff
@@ -41,37 +58,40 @@ func NewClient(ctx context.Context, opts ...option.ClientOption) (*Client, error
...
+func (c *GrpcClient) Connection() *grpc.ClientConn {
 	return c.connPool.Conn()
 }
 
-// Close closes the connection to the API service. The user should invoke this when
-// the client is no longer required.
-func (c *Client) Close() error {
```
tbpg, line 65: I guess this is not breaking because `internalClient` is embedded, so `Client` gets the `Close` method from there?

internal/gengapic/testdata/foo_client_init.want
```diff
@@ -1,35 +1,52 @@
...
+	// The call options for this service.
+	CallOptions *FooCallOptions
+
+}
+
+// fooGrpcClient is a client for interacting with Awesome Foo API over gRPC transport.
+// It satisfies the internalFooClient interface.
+//
+// Methods, except Close, may be called concurrently. However, fields must not be modified concurrently with method calls.
+type fooGrpcClient struct {
```
tbpg, line 22: I'm surprised this type isn't exported, but the one in `empty_client_init.want` is?

internal/gengapic/testdata/empty_client_init.want
software-dov: Yes, fixed

internal/gengapic/testdata/empty_client_init.want
software-dov: It's handled by the embedded interface; this is lost in the golden tests.

internal/gengapic/testdata/empty_client_init.want
software-dov: Exactly

internal/gengapic/testdata/foo_client_init.want
software-dov: It's an artifact of the empty client test. I can fix it if need be, but it's purely a testing issue.

internal/gengapic/testdata/lro_client_init.want
software-dov: Done

internal/gengapic/testdata/empty_client_init.want
software-dov: We need to leave it where it is so that the embedded client can get a pointer to the call options and the LROClient, if one exists.

internal/gengapic/gengapic.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2018 Google LLC
+// Copyright 2018 Google L/LC
```
noahdietz, line 1: Typo

internal/gengapic/testdata/lro_client_init.want
software-dov: Done

internal/gengapic/client_init.go
```diff
@@ -29,7 +29,7 @@ import (
 func (g *generator) clientHook(servName string) {
 	p := g.printf
 
-	p("var new%sClientHook clientHook", servName)
+	p("var new%sGRPCClientHook clientHook", servName)
```
noahdietz, line 32: This will actually break Spanner, who define a hook for the clients today. Can we leave this one named the same and maybe put a comment on it to the affect of `// newClientHook is a client initialization hook for gRPC client implementations`.

When Rest comes along, we can generate a separate hook like `newRestClientHook`. WDYT?

software-dov: I'm in favor

noahdietz (commented):
Once the `Operation` helpers are added to the generated interface, this should be backwards compat and lgtm. I want @tbpg or @codyoss to sign off. If that happens 4/23, i will be OOO and @vchudnov-g is my proxy :)

internal/gengapic/testdata/empty_client_init.want
tbpg: Ah, gotcha. Sounds good.

internal/gengapic/testdata/empty_client_init.want
tbpg: The interface embedding only brings the methods of the `internalClient` interface, not all of the methods of the `GRPCClient` type. See https://play.golang.org/p/WGSV7m8VDWN.

tbpg (changes requested):

internal/gengapic/testdata/empty_client_init.want
software-dov: Connection has now been added

codyoss (commented):
This LGTM, thanks @software-dov for all of your hard work on this! I will let someone from your team have final signoff. I think there is still on unresolved comment on the draft PR.

software-dov:
@noahdietz has pointed out that changing the name of the default client options thunk will break Spanner: https://github.com/googleapis/google-cloud-go/blob/master/spanner/apiv1/spanner_client_options.go#L24
This is a manageable change, though; when we regenerate Spanner, the build will fail, and we will just fix that when it happens. This comment is just a reminder of that eventuality.

@codyoss

noahdietz (approved):
> > changing the name of the default client options thunk will break Spanner: https://github.com/googleapis/google-cloud-go/blob/master/spanner/apiv1/spanner_client_options.go#L24
> This is a manageable change, though; when we regenerate Spanner, the build will fail, and we will just fix that when it happens. This comment is just a reminder of that eventuality.

Thanks @software-dov I think this is reasonable especially because I agree that the default client option helper(s) should probably be transport-specific.

tbpg (approved):
LGTM. Would be good to run apidiff on regenerated APIs before we release new versions to be extra careful.

PR #602, opened 2021-04-30
gapic: gen mixins based on http rules

### LINKED ISSUE #598: gapic: mixin http annotations should be pulled from the service config
The `google.api.http` annotations for mixin APIs should be extracted from the Service config and injected into the `MethodOptions` on the mixin RPCs. This ensure that the (re)gapic implementations are referencing the proper HTTP annotations for request header injection and URI creation.

noahdietz:
The main change in this PR is the use of the `http.rules` section to determine which mixin RPCs should be generated. Only those with such a binding will be generated. This is primarily because not every service implements all of the RPCs for a given mixin. The `http.rules` are the best indicator of this. Furthermore, those methods to generate will be overwritten with the `google.api.http` binding defined in the corresponding `http.rules` item. This way, the hosting service's http bindings are used for REGAPIC and header injection. 

This change also references the Service config's `documentation.rules` for method-level docs for the mixins. Absent those, a generic comment is added. 

Some refactoring of the mixin code is also done, in order to keep mixin logic inside of `mixins.go` and simplify call-sites elsewhere in the generator.

Fixes #598

### DIFF
--- a/internal/gengapic/mixins.go
+++ b/internal/gengapic/mixins.go
@@ -214,6 +214,7 @@ func (g *generator) checkIAMPolicyOverrides(servs []*descriptor.ServiceDescripto
 		for _, iamMethod := range iam {
 			if hasMethod(s, iamMethod.GetName()) {
 				g.hasIAMPolicyOverrides = true
+				return
 			}
 		}
 	}
--- a/internal/gengapic/mixins.go
+++ b/internal/gengapic/mixins.go
@@ -15,7 +15,13 @@
 package gengapic
 
 import (
+	"fmt"
+	"log"
+
+	"github.com/golang/protobuf/proto"
 	"github.com/golang/protobuf/protoc-gen-go/descriptor"
+	"github.com/googleapis/gapic-generator-go/internal/pbinfo"
+	"google.golang.org/genproto/googleapis/api/annotations"
 	"google.golang.org/genproto/googleapis/cloud/location"
 	iam "google.golang.org/genproto/googleapis/iam/v1"
 	"google.golang.org/genproto/googleapis/longrunning"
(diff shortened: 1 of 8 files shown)


internal/gengapic/mixins.go
```diff
@@ -56,65 +119,102 @@ func (g *generator) getMixinFiles() []*descriptor.FileDescriptorProto {
...
+func (g *generator) checkIAMPolicyOverrides(servs []*descriptor.ServiceDescriptorProto) {
+	iam, hasMixin := g.mixins["google.iam.v1.IAMPolicy"]
+	if !hasMixin {
+		return
+	}
+
+	for _, s := range servs {
+		for _, iamMethod := range iam {
+			if hasMethod(s, iamMethod.GetName()) {
+				g.hasIAMPolicyOverrides = true
```
software-dov, line 216: Nit: we could return after after setting the value.

internal/gengapic/mixins.go
```diff
@@ -56,65 +119,102 @@ func (g *generator) getMixinFiles() []*descriptor.FileDescriptorProto {
...
+		methods = append(methods, g.mixins["google.cloud.location.Locations"]...)
 	}
 	if g.hasIAMPolicyMixin() {
-		methods = append(methods, getIAMPolicyMethods()...)
+		methods = append(methods, g.mixins["google.iam.v1.IAMPolicy"]...)
 	}
 	if g.hasLROMixin() {
-		methods = append(methods, getOperationsMethods()...)
+		methods = append(methods, g.mixins["google.longrunning.Operations"]...)
 	}
```
software-dov, line 134: IIUC, this can be cleaned up somewhat.
```
for _, mixin_methods := range g.mixins {
    methods = append(methods, mixin_methods)
}
```

internal/gengapic/mixins.go
```diff
@@ -40,14 +46,71 @@ func init() {
...
+				// Set a default comment in case the Service does not have a DocumentationRule for it.
+				// Exclude the leading method name because methodDoc adds automatically.
+				g.comments[method] = fmt.Sprintf("is a utility method from %s.%s.", file.GetPackage(), service.GetName())
+			}
+		}
+	}
+
+	// Overwrite the google.api.http annotations with bindings from the Service config.
+	for _, rule := range g.serviceConfig.GetHttp().GetRules() {
+		m, match := methods[rule.GetSelector()]
```
software-dov, line 88: Are we guaranteed to have each mixin mentioned at most once in the http rules?

software-dov (commented):
Looks good, just some questions and a few nits.

internal/gengapic/mixins.go
noahdietz: Well, it is referenced in multiple places, and I don't want to have to execute this method multiple times (mainly because its best to do it with the full list of Services-to-generate, which isn't available in the scope of other call sites) so I saved the state in the generator its called with. Seems simpler that way.

internal/gengapic/mixins.go
noahdietz: Almost. Some of the `has` methods have logic in there to prevent the mixin from being generated in the event that 1. the API already overrides some mixin methods 2. the Service-to-generate is the mixin service itself (e.g. generating a client for Operations). 

I could look at pushing that logic into the initialization section though. I will take a look.

internal/gengapic/mixins.go
noahdietz: > Are we guaranteed to have each mixin mentioned at most once in the http rules?

Not sure I follow. Each mixin _method_? Yes, because it would have `additional_bindings` if a single method were going to have multiple `google.api.http` bindings. They would be collated into one rule by api_publish.

internal/gengapic/mixins.go
software-dov: Good to know about the `has` methods. Don't worry about it if it's a hassle; I just saw a potentially easy transform and thought "why not?"

internal/gengapic/mixins.go
software-dov: Yes, I agree; I'm saying that we can short circuit the loop and end if we set `hasIAMPolicyOverrides` to true because it's a single, sticky value.

noahdietz: Oh I totally misread your comment. yes, good catch! Thanks

internal/gengapic/mixins.go
noahdietz: It's a good idea. I'm going to punt on this for now, but I have an idea on how it could be improved. Just requires some plumbing and I'd like to separate it out.

internal/gengapic/mixins.go
software-dov: Okay, that answers my question.

software-dov (approved):

codyoss (approved):

PR #603, opened 2021-05-03
cleanup: move gRPC centric code generation to a gRPC specific file

software-dov:
This is a followup to an earlier review where it was stated that
moving gRPC centric code to its own file was a worthwhile maintenance
task.

The diff in this commit is large but is almost entirely moving around
existing code blocks with no structural or behavioral changes.
Any new code is just boilerplate intended to check which transports
the generator should emit code for and dispatch accordingly, or
renaming existing methods to indicate their gRPC centric nature.


internal/gengapic/gengrpc.go
```diff
@@ -0,0 +1,406 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package gengapic
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/golang/protobuf/proto"
```
codyoss, line 21: Could migrate to `google.golang.org/protobuf/proto`. But that could/should be its own PR.

codyoss (commented):
LGTM, but I'll defer to your team that knows this code better.

internal/gengapic/gengrpc.go
noahdietz: We should definitely start using the v2 module for all helper methods where possible, but migrating the plugin implementation to v2 will be a tall task so we may not be able to do everything. 

+1 to a separate PR though

software-dov: Agreed, that belongs in another PR.

tbpg (approved):

PR #605, opened 2021-05-05
gapic: generate service/method deprecation comments

### LINKED ISSUE #590: gapic: generate deprecation comment on proto elements with deprecated = true
When a proto element, like `service` or `rpc`, has the `deprecated` [option](https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/descriptor.proto#L682) set to `true`, `gapic-generator-go` should ensure that an appropriate comment (e.g. `// Deprecated: This may be removed in a future version.`) exists in the generated code.

If the proto element has an appropriate comment already (i.e. a comment line that starts with `Deprecated`), `gapic-generator-go` *should not* add anything - this proto comment will show up in the generated code and is sufficient for Go documentation. If it _does not_ have such a comment, then `gapic-generator-go` should *append* a *standalone* (meaning separated by a blank new line) comment line that reads `Deprecated: This may be removed in a future version.` For example:

```go
// GetFoo retrieves a Foo by name.
//
// Deprecated: This may be removed in a future version.
func (c *FooClient) GetFoo(ctx context.Context, req *foopb.GetFooRequest) (*foopb.Foo, error)
```

The [AIP (draft)](https://github.com/aip-dev/google.aip.dev/pull/736) describes generally what generators & API owners should do WRT consuming/using the `deprecated` option and supports the cases described above.

Useful links:
- [`MethodDescriptorProto.GetOptions`](https://pkg.go.dev/google.golang.org/protobuf@v1.26.0/types/descriptorpb#MethodDescriptorProto.GetOptions)
- [`MethodOptions.GetDeprecated`](https://pkg.go.dev/google.golang.org/protobuf@v1.26.0/types/descriptorpb#MethodOptions.GetDeprecated)
- [`internal/gengapic/gengapic.go methodDoc`](https://github.com/googleapis/gapic-generator-go/blob/f568699b3c0b289630a5df67ce3bd0c0b4c9c8b0/internal/gengapic/gengapic.go#L426-L436)
  - [use of `methodDoc()`](https://github.com/googleapis/gapic-generator-go/blob/f568699b3c0b289630a5df67ce3bd0c0b4c9c8b0/internal/gengapic/gengapic.go#L140) to generate `rpc` documentation
- [`internal/gengapic/gengapic.go comment`](https://github.com/googleapis/gapic-generator-go/blob/f568699b3c0b289630a5df67ce3bd0c0b4c9c8b0/internal/gengapic/gengapic.go#L438-L454)
  - [use of `comment()`](https://github.com/googleapis/gapic-generator-go/blob/f568699b3c0b289630a5df67ce3bd0c0b4c9c8b0/internal/gengapic/client_init.go#L230) to generate `service` documentation

alicejli:
When a method or service is deprecated, we will generate a default deprecation comment if the proto comments don't already contain one. 

Proto service documentation is now also populated on the client type as well as the factory method.

Note to @software-dov: Change was made to commented out genrest factory method. 

Also cleaned up some linter warnings on the client tests.

Fixes #590


internal/gengapic/gengapic_test.go
```diff
@@ -378,6 +349,59 @@ methods:
...
+			deprecated: true,
+		},
+		{
+			in:         "Deprecated: this is a proper deprecation notice.",
+			want:       "// Deprecated: this is a proper deprecation notice.\n",
+			deprecated: true,
+		},
+		{
+			in:         "This is a comment that includes Deprecated but not at the beginning.",
+			want:       "// MyMethod this is a comment that includes Deprecated but not at the beginning.\n",
```
codyoss, line 384: I think we still might want a deprecated line here. We should double check to see what the common Go linters check for when reading deprecation notices.

codyoss: https://github.com/golang/go/wiki/Deprecated

I think we need the "// Deprecated:"

noahdietz: So a proto comment (which in the generator doesn't have `//`) would need to have some line that begins with `"Deprecated:"` in order for us to _not_ generate our own generic message?

codyoss: I believe so.

internal/gengapic/client_init.go
```diff
@@ -144,12 +144,36 @@ func (g *generator) internalClientIntfInit(serv *descriptor.ServiceDescriptorPro
 	return nil
 }
 
+// This is a helper function similar to methodDoc that includes a deprecation notice for deprecated services.
```
tbpg, line 147: Nit:
```suggestion
// serviceDoc is a helper function similar to methodDoc that includes a deprecation notice for deprecated services.
```

See https://github.com/golang/go/wiki/CodeReviewComments#comment-sentences.

internal/gengapic/gengapic_test.go
```diff
@@ -378,6 +349,59 @@ methods:
 	}
 }
 
+func TestMethodDoc(t *testing.T) {
+	m := &descriptor.MethodDescriptorProto{
+		Name: proto.String("MyMethod"),
+	}
+
+	var g generator
+	g.comments = make(map[proto.Message]string)
```
tbpg, line 358:
```suggestion
	g := generator{
		comments: make(map[proto.Message]string),
	}
```

internal/gengapic/gengapic_test.go
```diff
@@ -378,6 +349,59 @@ methods:
...
+			want: "// MyMethod does stuff.\n// It also does other stuffs.\n",
+		},
+		{
+			in:         "Does not have a proper comment.",
+			want:       "// MyMethod does not have a proper comment.\n//\n// Deprecated: This may be removed in a future version.\n",
+			deprecated: true,
+		},
+		{
+			in:         "Deprecated: this is a proper deprecation notice.",
+			want:       "// Deprecated: this is a proper deprecation notice.\n",
```
tbpg, line 379: I think we still want the comment to start with the name of the thing being described (https://github.com/golang/go/wiki/CodeReviewComments#comment-sentences)? So, it could be something like:

```
// Name is deprecated.
//
// Deprecated: this is a proper deprecation notice.
```

I could be wrong, but following the comment sentence convention every time seems good.

tbpg (commented):
A few minor comments. Looks good! Thank you!

internal/gengapic/testdata/deprecated_client_init.want
```diff
@@ -0,0 +1,121 @@
...
+	setGoogleClientInfo(...string)
+	Connection() *grpc.ClientConn
+	Zip(context.Context, *mypackagepb.Bar, ...gax.CallOption) (*mypackagepb.Foo, error)
+}
+
+// Client is a client for interacting with Awesome Foo API.
+// Methods, except Close, may be called concurrently. However, fields must not be modified concurrently with method calls.
+//
+// Foo service does stuff.
+// Foo is deprecated.
```
noahdietz, line 13: Per my previous comments, this looks a little odd having the added "Foo is deprecated" statement, I think the Deprecation tag is sufficient.

internal/gengapic/gengapic.go
```diff
@@ -306,25 +306,40 @@ func (g *generator) appendCallOpts(m *descriptor.MethodDescriptorProto) {
 	g.printf("opts = append(%[1]s[0:len(%[1]s):len(%[1]s)], opts...)", "(*c.CallOptions)."+*m.Name)
 }
 
+// This is a helper function that checks whether a description contains a Deprecated header.
+func containsDeprecated(com string) bool {
+	// split the string based on line breaks
+	s := strings.Split(com, "\n")
+	// check each string for the 'Deprecated: ' prefix
+	for i := 0; i < len(s); i++ {
+		if strings.HasPrefix(s[i], "Deprecated:") {
```
tbpg, line 315:
```suggestion
	for _, l := range s {
		if strings.HasPrefix(l, "Deprecated:") {
```

internal/gengapic/client_init.go
```diff
@@ -144,16 +144,18 @@ func (g *generator) internalClientIntfInit(serv *descriptor.ServiceDescriptorPro
...
-	// If the service is marked as deprecated and there is not a comment explaining it, then add default deprecation comment.
-	// If the service includes a deprecation notice, then use that.
+	// If the service is marked as deprecated and there is no comment explaining it, then add default deprecation comment.
+	// If the service includes a deprecation notice, then prepend a comment with the service name stating it is deprecated and use the included deprecation notice.
 
 	if serv.GetOptions().GetDeprecated() {
-		if com == "" || !strings.Contains(com, "Deprecated") {
-			com += "\n\nDeprecated: This may be removed in a future version."
+		if !containsDeprecated(com) {
+			com += "\n" + serv.GetName() + " is deprecated.\n\nDeprecated: This may be removed in a future version."
```
tbpg, line 156: Sorry, one other place we can be clear with the name. Can we use the name instead of "This"? So, "Foo may be removed ..."

internal/gengapic/gengapic_test.go
```diff
@@ -378,6 +349,89 @@ methods:
...
+			deprecated: true,
+		},
+		{
+			in:         "Deprecated: this is a proper deprecation notice.",
+			want:       "// MyMethod is deprecated.\n//\n// Deprecated: this is a proper deprecation notice.\n",
+			deprecated: true,
+		},
+		{
+			in:         "",
+			want:       "// MyMethod is deprecated.\n//\n// Deprecated: This may be removed in a future version.\n",
```
noahdietz, line 419: This is the exact case I think we should include the "MyMethod is deprecated" statement.

tbpg (approved):
A couple small comments.

internal/gengapic/gengapic_test.go
```diff
@@ -378,6 +349,89 @@ methods:
...
+			in:   "",
+			want: "",
+		},
+		{
+			in:   "Does stuff.\n It also does other stuffs.",
+			want: "// MyMethod does stuff.\n// It also does other stuffs.\n",
+		},
+		{
+			in:         "This is deprecated.\n It does not have a proper comment.",
+			want:       "// This is deprecated.\n// It does not have a proper comment.\n// MyMethod is deprecated.\n//\n// Deprecated: This may be removed in a future version.\n",
```
noahdietz, line 409: We need to include the method name on this one - there was a proto comment to generate, so we need to make sure we start the whole comment with the method name.

noahdietz (approved):

internal/gengapic/gengapic_test.go
alicejli: Thanks for catching. Done.

internal/gengapic/gengapic_test.go
alicejli: Done.

internal/gengapic/client_init.go
alicejli: Makes sense - changed to reference the name instead of "This."

internal/gengapic/gengapic.go
```diff
@@ -306,16 +306,44 @@ func (g *generator) appendCallOpts(m *descriptor.MethodDescriptorProto) {
 	g.printf("opts = append(%[1]s[0:len(%[1]s):len(%[1]s)], opts...)", "(*c.CallOptions)."+*m.Name)
 }
 
+// This is a helper function that checks whether a description contains a Deprecated header.
+func containsDeprecated(com string) bool {
+	// split the string based on line breaks
+	s := strings.Split(com, "\n")
+	// check each string for the 'Deprecated: ' prefix
+	for i := 0; i < len(s); i++ {
```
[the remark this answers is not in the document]
alicejli: Thanks! Updated

internal/gengapic/gengapic_test.go
alicejli: Thanks! Updated to include the method name in the beginning.

internal/gengapic/client_init.go
```diff
@@ -144,12 +144,38 @@ func (g *generator) internalClientIntfInit(serv *descriptor.ServiceDescriptorPro
...
+// serviceDoc is a helper function similar to methodDoc that includes a deprecation notice for deprecated services.
+func (g *generator) serviceDoc(serv *descriptor.ServiceDescriptorProto) {
+	com := g.comments[serv]
+
+	// If the service is marked as deprecated and there is no comment explaining it, then add default deprecation comment.
+	// If the service includes a deprecation notice, then prepend a comment with the service name stating it is deprecated and use the included deprecation notice.
+
+	if serv.GetOptions().GetDeprecated() {
+		if !containsDeprecated(com) {
+			com += "\n" + serv.GetName() + " is deprecated.\n\nDeprecated: This may be removed in a future version."
```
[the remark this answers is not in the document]
alicejli: Updated to only include `Foo is deprecated` statement for cases where there is no comment or when the comment is prefixed with the `Deprecation` tag per our offline convo.

software-dov (approved):

internal/gengapic/client_init.go
```diff
@@ -148,22 +148,25 @@ func (g *generator) internalClientIntfInit(serv *descriptor.ServiceDescriptorPro
...
+	// If the service is marked as deprecated and there is no comment, then add default deprecation comment.
+	// If the service has a comment but it does not include a deprecation notice, then append a default deprecation notice.
+	// If the service includes a deprecation notice at the beginning of the comment, prepend a comment stating the service is deprecated and use the included deprecation notice.
 	if serv.GetOptions().GetDeprecated() {
-		if !containsDeprecated(com) {
-			com += "\n" + serv.GetName() + " is deprecated.\n\nDeprecated: This may be removed in a future version."
-		} else {
-			com = "\n" + serv.GetName() + " is deprecated.\n\n" + com
+		if com == "" {
+			com = fmt.Sprintf("\n%s is deprecated.\n\nDeprecated: %s may be removed in a future version.", serv.GetName(), serv.GetName())
```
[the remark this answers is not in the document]
alicejli: Thanks - fixed

PR #607, opened 2021-05-06
feat: generate rest client method implementation

software-dov:
This change generates code for unary and empty unary methods for REST
based clients.

TODO:
* Set http headers
* Handle query parameters
* Expose REST client factory function


internal/gengapic/genrest.go
```diff
@@ -112,7 +159,59 @@ func (g *generator) genRESTMethod(servName string, serv *descriptor.ServiceDescr
...
+	// TODO(dovs): handle deadlines?
+	// TODO(dovs): handle call options
+	p("// The default (false) for the other options are fine.")
+	p("m := protojson.MarshalOptions{AllowPartial: true, Multiline: true, EmitUnpopulated: true}")
+	p("if jsonReq, err := m.Marshal(req); err != nil {")
+	p("  return err")
+	p("}")
+	p("")
+	// TODO(dovs): handle path parameters
+	p("url := fmt.Sprintf(\"https://%%s%s\", c.host)", info.url)
```
noahdietz, line 190: You can use backtick instead of `"` so that you don't have to escape the double quotes.

internal/gengapic/genrest.go
```diff
@@ -112,7 +159,59 @@ func (g *generator) genRESTMethod(servName string, serv *descriptor.ServiceDescr
...
+	p("httpReq, err := http.NewRequestWithContext(ctx, \"%s\", url, bytes.NewReader(jsonReq))", strings.ToUpper(info.verb))
```
noahdietz, line 191: Here too with the backticks

internal/gengapic/genrest.go
```diff
@@ -112,7 +159,59 @@ func (g *generator) genRESTMethod(servName string, serv *descriptor.ServiceDescr
...
+	p("if err != nil {")
+	p("    return err")
+	p("}")
+	p("")
+	p("httpRsp, err := client.Do(httpReq)")
+	p("if err != nil{")
+	p(" return err")
+	p("} else if httpRsp.StatusCode >= 400 {")
+	p("  return errors.E(nil, httpRsp.Status)")
```
noahdietz, line 200: `errors.E` isn't a standard golang API, its an internal one for the generator. There is https://golang.org/pkg/errors though.

internal/gengapic/genrest.go
```diff
@@ -112,7 +159,59 @@ func (g *generator) genRESTMethod(servName string, serv *descriptor.ServiceDescr
...
+	p("")
+	p("return nil")
+	p("}")
+
+	return nil
+}
+
 func (g *generator) unaryRESTCall(servName string, m *descriptor.MethodDescriptorProto) error {
+	info, err := getHTTPInfo(m)
+	if err != nil || info == nil {
```
noahdietz, line 211: Same comment WRT `err == nil && info == nil` case.

internal/gengapic/genrest.go
```diff
@@ -131,24 +230,39 @@ func (g *generator) unaryRESTCall(servName string, m *descriptor.MethodDescripto
...
+	// TODO(dovs): handle deadlines?
+	// TODO(dovs): handle call options
+	// TODO(dovs): handle path parameters
+	p("// The default (false) for the other options are fine.")
+	p("marshaler := protojson.MarshalOptions{AllowPartial: true, Multiline: true, EmitUnpopulated: true}")
+	p("if jsonReq, err := marshaler.Marshal(req); err != nil {")
+	p("  return nil, err")
+	p("}")
+	p("")
+	p("url := fmt.Sprintf(\"https://%%s%s\", c.host)", info.url)
```
noahdietz, line 243: Same comment WRT use of backticks instead of `""` with the printer

internal/gengapic/genrest.go
```diff
@@ -112,7 +159,59 @@ func (g *generator) genRESTMethod(servName string, serv *descriptor.ServiceDescr
...
+	p("}")
+	p("")
+	// TODO(dovs): handle path parameters
+	p("url := fmt.Sprintf(\"https://%%s%s\", c.host)", info.url)
+	p("httpReq, err := http.NewRequestWithContext(ctx, \"%s\", url, bytes.NewReader(jsonReq))", strings.ToUpper(info.verb))
+	p("if err != nil {")
+	p("    return err")
+	p("}")
+	p("")
+	p("httpRsp, err := client.Do(httpReq)")
```
noahdietz, line 196: Do we need to do a `httpRsp.Body.Close()` somewhere? Maybe `defer` it? Even though we won't return the `emptypb.Empty`, the backed is still sending back the `google.protobuf.Empty` message in the response body.

internal/gengapic/genrest.go
```diff
@@ -131,24 +230,39 @@ func (g *generator) unaryRESTCall(servName string, m *descriptor.MethodDescripto
...
-
-	p("return nil, nil")
 	p("")
+	p("buf := new(bytes.Buffer)")
+	p("err = buf.ReadFrom(httpRsp.Body)")
+	p("if err != nil {")
+	p("  return nil, err")
+	p("}")
+	p("")
+	p("unmarshaler := protojson.UnmarshalOptions{AllowPartial: true, DiscardUnknown: true}")
```
noahdietz, line 262: Hmm do we want to discard unknown? If we _didn't_ discard them, would unmarshaling return an error? If it doesn't error, my thinking is that the unknown fields are added to the Unknown Fields section of the message (https://developers.google.com/protocol-buffers/docs/proto3#unknowns) and we should allow that.

internal/gengapic/genrest.go
```diff
@@ -131,24 +230,39 @@ func (g *generator) unaryRESTCall(servName string, m *descriptor.MethodDescripto
...
-	// * JSONify the request
-	// * determine the url
-	// ** Handle query parameters
-	// * send the request, receive the response, maybe return an error
-	// * deserialize the result and return
-
-	p("return nil, nil")
 	p("")
+	p("buf := new(bytes.Buffer)")
+	p("err = buf.ReadFrom(httpRsp.Body)")
```
noahdietz, line 257: Does `buf` auto-close the Body? If not, do we need to `defer` that?

internal/gengapic/genrest.go
```diff
@@ -131,24 +230,39 @@ func (g *generator) unaryRESTCall(servName string, m *descriptor.MethodDescripto
...
-
-	// * JSONify the request
-	// * determine the url
-	// ** Handle query parameters
-	// * send the request, receive the response, maybe return an error
-	// * deserialize the result and return
-
-	p("return nil, nil")
 	p("")
+	p("buf := new(bytes.Buffer)")
```
noahdietz, line 256: I think we will need to include the import for `bytes` in the generated code.

internal/gengapic/genrest.go
```diff
@@ -112,7 +159,59 @@ func (g *generator) genRESTMethod(servName string, serv *descriptor.ServiceDescr
 	}
 }
 
+func (g *generator) emptyUnaryRESTCall(servName string, m *descriptor.MethodDescriptorProto) error {
+	info, err := getHttpInfo(m)
+	if err != nil || info == nil {
+		return err
```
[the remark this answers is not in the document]
software-dov: Good catch. This is a fat finger on my part.

internal/gengapic/genrest.go
```diff
@@ -112,7 +159,59 @@ func (g *generator) genRESTMethod(servName string, serv *descriptor.ServiceDescr
...
+	lowcaseServName := lowcaseRestClientName(servName)
+	p("func (c *%s) %s(ctx context.Context, req *%s.%s, opts ...gax.CallOption) error {",
+		lowcaseServName, m.GetName(), inSpec.Name, inType.GetName())
+
+	// TODO(dovs): handle cancellation, metadata, osv.
+	// TODO(dovs): handle http headers
+	// TODO(dovs): handle deadlines?
+	// TODO(dovs): handle call options
+	p("// The default (false) for the other options are fine.")
+	p("m := protojson.MarshalOptions{AllowPartial: true, Multiline: true, EmitUnpopulated: true}")
```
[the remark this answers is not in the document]
software-dov: I was leanig towards multiline because it's easier to eyeball parse, but I have no strong feelings on the matter. If you think the extra bytes aren't worth it I'll change this.

internal/gengapic/genrest.go
software-dov: Good to know! I'm not sure how the format string will interact with the backtick, though. I'll try it out and see how it works.

noahdietz: To be clear, i mean you can do 
```go
p(`url := fmt.Sprintf("https://%%s%s", c.host)`, info.url)
```

Realized that my comment wasn't super clear.

internal/gengapic/genrest.go
software-dov: Yes, you're right. Taking a look at the `Response` definition,
```
It is the caller's responsitibility to close Body.
```

internal/gengapic/genrest.go
software-dov: Taking a look at the code, it looks like seeing an unknown field with `DiscardUnknown = false` **does** generate an error. 
```
		if fd == nil {
			// Field is unknown.
			if d.opts.DiscardUnknown {
				if err := d.skipJSONValue(); err != nil {
					return err
				}
				continue
			}
			return d.newError(tok.Pos(), "unknown field %v", tok.RawString())
		}
```

internal/gengapic/genrest.go
software-dov: It is something we need to do. Good catch.

internal/gengapic/genrest.go
```diff
@@ -78,6 +87,44 @@ func (g *generator) restClientUtilities(serv *descriptor.ServiceDescriptorProto,
...
+		return nil, nil
+	}
+
+	eHTTP, err := proto.GetExtension(m.GetOptions(), annotations.E_Http)
+	if err == proto.ErrMissingExtension {
+		return nil, nil
+	} else if err != nil {
+		return nil, err
+	}
+	info := httpInfo{verb: "", url: "", body: ""}
```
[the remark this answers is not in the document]
software-dov: My bad, I thought string was inherently a reference type.

internal/gengapic/genrest.go
noahdietz: TIL! Thanks, let's keep it.

internal/gengapic/genrest.go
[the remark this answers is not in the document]
software-dov: Removing multiline.

software-dov:
Open review bump @codyoss @noahdietz @tbpg @vchudnov-g

internal/gengapic/testdata/empty_client_init.want
```diff
@@ -100,14 +100,38 @@ func (c *gRPCClient) Close() error {
...
+	}
+
+	url := fmt.Sprintf("https://%s/zip", c.host)
+	httpReq, err := http.NewRequestWithContext(ctx, "GET", url, bytes.NewReader(jsonReq))
+	if err != nil {
+		return nil, err
+	}
 
+	httpRsp, err := client.Do(httpReq)
+	defer httpRsp.Body.Close()
```
codyoss, line 120: The error should be checked before the close of the body is defered.

internal/gengapic/testdata/empty_client_init.want
```diff
@@ -100,14 +100,38 @@ func (c *gRPCClient) Close() error {
...
+	if err != nil{
+		return nil, err
+	} else if httpRsp.StatusCode >= 400 {
+		return nil, errors.E(nil, httpRsp.Status)
```
codyoss, line 124: I am not sure what errors.E is, but I don't think we want this dependency? We should process errors similarily if not exactly how it is done in the apiary libraries. There is some special logic to parse known proto struct errors there. Example: https://github.com/googleapis/google-api-go-client/blob/v0.46.0/googleapi/googleapi.go#L130

internal/gengapic/testdata/empty_client_init.want
```diff
@@ -100,14 +100,38 @@ func (c *gRPCClient) Close() error {
...
 	}
-	return nil, nil
 
+	buf := new(bytes.Buffer)
```
codyoss, line 127: If we need to read everything into memory maybe just https://pkg.go.dev/io/ioutil#ReadAll?

internal/gengapic/testdata/empty_client_init.want
```diff
@@ -100,14 +100,38 @@ func (c *gRPCClient) Close() error {
...
 type restClient struct {
 	host string
+	httpClient http.Client
 }
 
 func (c *restClient) Zip(ctx context.Context, req *mypackagepb.Bar, opts ...gax.CallOption) (*mypackagepb.Foo, error) {
-	m := jsonpb.Marshaler{}
-	if jsonReq, err := m.MarshalToString(req); err != nil {
+	// The default (false) for the other options are fine.
+	marshaler := protojson.MarshalOptions{AllowPartial: true, Multiline: true, EmitUnpopulated: true}
```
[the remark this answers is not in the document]
software-dov: I believe `EmitUnpopulated` is necessary for correct field presence for optional fields.

internal/gengapic/testdata/empty_client_init.want
```diff
@@ -100,14 +100,38 @@ func (c *gRPCClient) Close() error {
...
+	if jsonReq, err := marshaler.Marshal(req); err != nil {
 		return nil, err
+	}
+
+	url := fmt.Sprintf("https://%s/zip", c.host)
```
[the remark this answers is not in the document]
software-dov: s/host/endpoint and moved the responsibility of "https://" into endpoint.

internal/gengapic/testdata/empty_client_init.want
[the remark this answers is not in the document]
codyoss: Ack

internal/gengapic/testdata/deprecated_client_init.want
```diff
@@ -131,15 +131,39 @@ func (c *gRPCClient) Close() error {
...
-	host string
+	endpoint string
+	httpClient http.Client
 }
 
 func (c *restClient) Zip(ctx context.Context, req *mypackagepb.Bar, opts ...gax.CallOption) (*mypackagepb.Foo, error) {
-	m := jsonpb.Marshaler{}
-	if jsonReq, err := m.MarshalToString(req); err != nil {
+	// The default (false) for the other options are fine.
+	marshaler := protojson.MarshalOptions{AllowPartial: true, EmitUnpopulated: true}
```
codyoss, line 140: Should we AllowPartial? This will just fail when it hits the backend then, right?

internal/gengapic/testdata/deprecated_client_init.want
```diff
@@ -131,15 +131,39 @@ func (c *gRPCClient) Close() error {
...
+	if jsonReq, err := marshaler.Marshal(req); err != nil {
 		return nil, err
+	}
+
+	url := fmt.Sprintf("https://%s/zip", c.host)
```
codyoss: This line needs to be updated.

internal/gengapic/testdata/deprecated_client_init.want
```diff
@@ -131,15 +131,39 @@ func (c *gRPCClient) Close() error {
...
+		return nil, err
+	}
+
+	httpRsp, err := client.Do(httpReq)
+	if err != nil{
+		return nil, err
+	}
+	defer httpRsp.Body.Close()
+	if httpRsp.StatusCode >= 400 {
+		return nil, fmt.Errorf(httpRsp.Status)
```
codyoss, line 157: I think we will need to do more with this error. It should be able to be cast and parsed. I think I left a previous comment showing how we do something similar in the apiaries. It is fine to leave this as a TODO if this is out-of-scope right now but this should be done before any rest client is generated.

internal/gengapic/testdata/deprecated_client_init.want
```diff
@@ -131,15 +131,39 @@ func (c *gRPCClient) Close() error {
...
+	buf, err := ioutil.ReadAll(httpRsp.Body)
+	if err != nil {
+		return nil, err
 	}
-	return nil, nil
 
+	unmarshaler := protojson.UnmarshalOptions{AllowPartial: true, DiscardUnknown: true}
+	rsp := &mypackagepb.Foo{}
+
+	return rsp, unmarshaler.Unmarshal(buf.Bytes(), rsp)
```
codyoss, line 168: buf is a []byte now.

internal/gengapic/testdata/deprecated_client_init.want
software-dov: It's already the backend's responsibility to check, is it not?

noahdietz: This may be a `proto2` hold-over. `proto3` doesn't have `required`, and the `AllowPartial` docs say this `allows messages that have missing required fields to marshal without returning an error`. There'd be no use for this in `proto3` syntax since there is no `required`

noahdietz (approved):
LGTM please allow codyoss@ to Approve

internal/gengapic/genrest.go
```diff
@@ -111,7 +161,67 @@ func (g *generator) genRESTMethod(servName string, serv *descriptor.ServiceDescr
...
+	// TODO(dovs): handle http headers
+	// TODO(dovs): handle deadlines
+	// TODO(dovs): handle call options
+	p("// The default (false) for the other options are fine.")
+	p("m := protojson.MarshalOptions{AllowPartial: true, EmitUnpopulated: true}")
+	p("if jsonReq, err := m.Marshal(req); err != nil {")
+	p("  return err")
+	p("}")
+	p("")
+	// TODO(dovs): handle path parameters
```
vchudnov-g, line 194: nit: Consider putting this TODO in the generated code, so we see it in both places (and people looking at the generated code see that's not handled yet)

vchudnov-g (commented):
lgtm, some very minor comments

internal/gengapic/testdata/foo_rest_client_init.want
```diff
@@ -72,51 +72,167 @@ func (c *FooClient) TestIamPermissions(ctx context.Context, req *iampb.TestIamPe
...
 func (c *fooRESTClient) GetLocation(ctx context.Context, req *locationpb.GetLocationRequest, opts ...gax.CallOption) (*locationpb.Location, error) {
-	m := jsonpb.Marshaler{}
-	if jsonReq, err := m.MarshalToString(req); err != nil {
+	// The default (false) for the other options are fine.
+	marshaler := protojson.MarshalOptions{AllowPartial: true, EmitUnpopulated: true}
+	if jsonReq, err := marshaler.Marshal(req); err != nil {
 		return nil, err
+	}
 
+	url := fmt.Sprintf("%s/v1/{name=locations/*}", c.endpoint)
```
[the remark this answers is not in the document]
software-dov: Correct

internal/gengapic/genrest.go
```diff
@@ -199,10 +198,11 @@ func (g *generator) emptyUnaryRESTCall(servName string, m *descriptor.MethodDesc
 	p("}")
 	p("")
 	p("httpRsp, err := client.Do(httpReq)")
-	p("defer httpRsp.Body.Close()")
 	p("if err != nil{")
 	p(" return err")
-	p("} else if httpRsp.StatusCode >= 400 {")
+	p("}")
+	p("defer httpRsp.Body.Close()")
+	p("if httpRsp.StatusCode != 200 {")
```
vchudnov-g, line 205: nit: I would use `http.StatusOK` because it's more descriptive
https://golang.org/pkg/net/http/#pkg-constants

internal/gengapic/testdata/deprecated_client_init.want
codyoss: > It's already the backend's responsibility to check, is it not?

Backend would fail, but I think it is generally better to fail fast and save sending an RPC to figure this out.

> This may actually only matter for proto2,

If it is I don't think we need to set it then.

internal/gengapic/testdata/foo_rest_client_init.want
```diff
@@ -72,51 +72,7 @@ func (c *FooClient) TestIamPermissions(ctx context.Context, req *iampb.TestIamPe
 
 // Methods, except Close, may be called concurrently. However, fields must not be modified concurrently with method calls.
 type fooRESTClient struct {
-	host string
+	endpoint string
```
vchudnov-g, line 75: So where are the generated methods like `Zip()` now?

software-dov: They were being generated incorrectly in the test. It's a longish story:
* Turns out showcase generation didn't turn on rest client surface
* Turning on rest transport for showcase revealed that method generation was erroneously happening twice
* Resolving the issue in favor of the grpc-like implementation means that the golden files no longer contain concrete transport methods
* Generating rest transport surface for showcase revealed other bugs

vchudnov-g:
So now that #623 is in, this PR can be dropped?

noahdietz:
Closing this as it is obsolete.

PR #611, opened 2021-05-07
gapic: add docs on NewClient about closing the client

codyoss:
One issue we have seen pop up multiple times is around users of our
libraries not closing their clients after they are done being used.
This can lead to memory leaks as the underlying HTTP connections
are never closed. Adding some docs on the function that returns the
Client seems like a good first step to take.

### DIFF
--- a/internal/gengapic/gengrpc.go
+++ b/internal/gengapic/gengrpc.go
@@ -318,7 +318,7 @@ func (g *generator) grpcClientUtilities(serv *descriptor.ServiceDescriptorProto,
 
 	// Factory function
 	p("// New%sClient creates a new %s client based on gRPC.", servName, clientName)
-	p("// The returned client must be closed when it is done being used to free the underlying connections that have been made.")
+	p("// The returned client must be Closed when it is done being used to clean up its underlying connections.")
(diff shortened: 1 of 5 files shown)


internal/gengapic/gengrpc.go
```diff
@@ -318,6 +318,7 @@ func (g *generator) grpcClientUtilities(serv *descriptor.ServiceDescriptorProto,
 
 	// Factory function
 	p("// New%sClient creates a new %s client based on gRPC.", servName, clientName)
+	p("// The returned client must be closed when it is done being used to free the underlying connections that have been made.")
```
tbpg, line 321:
```suggestion
	p("// The returned client must be Closed when it is done being used to clean up its underlying connections.")
```

tbpg (approved):
Minor wording suggestion. Feel free to word smith more.

noahdietz:
@codyoss can you run `make golden` and push the updated baselines? I think there was a change but the baselines weren't updated to match hence the test failures.

noahdietz (approved):

vchudnov-g (approved):

PR #619, opened 2021-05-11
gapic: fix missing LRO type wrapper

noahdietz:
This adds the missing LRO type wrappers and it also fixes the the client_init tests to produce accurate generated code (e.g. `ListLocations` should return  a `LocationIterator` nor a `ListLocationsResponse`.

### DIFF
--- a/internal/gengapic/testdata/foo_rest_client_init.want
+++ b/internal/gengapic/testdata/foo_rest_client_init.want
@@ -4,7 +4,7 @@ type internalFooClient interface {
 	setGoogleClientInfo(...string)
 	Connection() *grpc.ClientConn
 	Zip(context.Context, *mypackagepb.Bar, ...gax.CallOption) (*mypackagepb.Foo, error)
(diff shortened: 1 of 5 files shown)


internal/gengapic/testdata/foo_rest_client_init.want
```diff
@@ -83,15 +83,6 @@ func (c *fooRESTClient) Zip(ctx context.Context, req *mypackagepb.Bar, opts ...g
 	}
 	return nil, nil
 
-}
```
codyoss, line 86: Should there be a method to replace this?

noahdietz: No, b.c the REST transport hasn't implemented Paginated methods yet. It was generated b.c it was incorrectly being viewed as a standard Unary RPC (which REST transport recently implemented)

software-dov (approved):

codyoss (approved):

PR #622, opened 2021-05-12
gapic: gen RPC docs on wrapper methods

noahdietz:
Since the wrapper methods are the "public facing" methods of a `Client`, we should generate the RPC documentation on them instead of the internal, transport-based, interface implementation.

### DIFF
--- a/internal/gengapic/gengrpc.go
+++ b/internal/gengapic/gengrpc.go
@@ -32,7 +32,6 @@ func (g *generator) genGRPCMethods(serv *descriptor.ServiceDescriptorProto, serv
 
 	methods := append(serv.GetMethod(), g.getMixinMethods()...)
 	for _, m := range methods {
-		g.methodDoc(m)
 		if err := g.genGRPCMethod(servName, serv, m); err != nil {
 			return errors.E(err, "method: %s", m.GetName())
 		}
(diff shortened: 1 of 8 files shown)


internal/gengapic/gengrpc.go
```diff
@@ -32,7 +32,6 @@ func (g *generator) genGRPCMethods(serv *descriptor.ServiceDescriptorProto, serv
 
 	methods := append(serv.GetMethod(), g.getMixinMethods()...)
 	for _, m := range methods {
-		g.methodDoc(m)
```
codyoss, line 35: Should this also be removed from `genGRPCMethod `?

noahdietz: We can keep it for sure, but the `gRPCClient` is internal and would be a duplicate of what is on the wrapper. WDYT? Happy to go either way.

software-dov (approved):

internal/gengapic/gengrpc.go
vchudnov-g: I would keep the method comments on both the wrapper and the internals just to help whoever peruses the code (mostly us?). Since they're autogen'd, they'll always stay in sync.

Not a biggie either way.

software-dov: +1 for removing it from gRPC client as well.

noahdietz: @codyoss what say ye? I'm leaning towards removing it (as I've done).

noahdietz: Cody +1 my previous comment, so we will omit generated docs on the internal methods for now. Resolving.

PR #626, opened 2021-05-14
feat: stub out currently unsupported method types for REST clients

software-dov:
The initial scope for DIREGAPIC is unary method calls, but clients
must still expose _all_ API methods in order to fit the internal interface.
Just return helfpul errors for now.

### DIFF
--- a/internal/gengapic/genrest.go
+++ b/internal/gengapic/genrest.go
@@ -162,10 +162,9 @@ func (g *generator) genRESTMethod(servName string, serv *descriptor.ServiceDescr
 }
 
 func (g *generator) serverStreamRESTCall(servName string, s *descriptor.ServiceDescriptorProto, m *descriptor.MethodDescriptorProto) error {
-	// Streaming calls will most likely NEVER be supported for REST clients,
+	// Streaming calls are not currently supported for REST clients,
 	// but the interface signature must be preserved.
-	// Making sure not to call streaming methods on a REST client, or checking for
-	// errors in a situation with mixed gRPC and REST clients, is left to user code.
+	// Unimplemented REST methods will always error.
 
 	inType := g.descInfo.Type[m.GetInputType()]
 
@@ -193,10 +192,9 @@ func (g *generator) serverStreamRESTCall(servName string, s *descriptor.ServiceD
 }
 
 func (g *generator) noRequestStreamRESTCall(servName string, s *descriptor.ServiceDescriptorProto, m *descriptor.MethodDescriptorProto) error {
-	// Streaming calls will most likely NEVER be supported for REST clients,
+	// Streaming calls are not currently supported for REST clients,
 	// but the interface signature must be preserved.
-	// Making sure not to call streaming methods on a REST client, or checking for
-	// errors in a situation with mixed gRPC and REST clients, is left to user code.
+	// Unimplemented REST methods will always error.
 
 	p := g.printf
 
--- a/internal/gengapic/genrest.go
+++ b/internal/gengapic/genrest.go
@@ -132,9 +132,8 @@ func getHTTPInfo(m *descriptor.MethodDescriptorProto) (*httpInfo, error) {
 // If the generated method requires an auxiliary type, it is added to aux.
 func (g *generator) genRESTMethod(servName string, serv *descriptor.ServiceDescriptorProto, m *descriptor.MethodDescriptorProto) error {
 	if g.isLRO(m) {
-		// TODO(dovs
 		g.aux.lros[m] = true
-		return nil
+		return g.lroRESTCall(servName, m)
 	}
 
 	if m.GetOutputType() == emptyType {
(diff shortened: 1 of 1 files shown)


internal/gengapic/genrest.go
```diff
@@ -144,23 +143,152 @@ func (g *generator) genRESTMethod(servName string, serv *descriptor.ServiceDescr
...
+		return err
+	}
+	g.imports[servSpec] = true
+
+	p := g.printf
+	lowcaseServName := lowcaseRestClientName(servName)
+	p("func (c *%s) %s(ctx context.Context, req *%s.%s, opts ...gax.CallOption) (%s.%s_%sClient, error) {",
+		lowcaseServName, m.GetName(), inSpec.Name, inType.GetName(), servSpec.Name, s.GetName(), m.GetName())
+	p("  var resp %s.%s_%sClient", servSpec.Name, s.GetName(), m.GetName())
+	p("  return resp, fmt.Errorf(\"%s not yet supported for REST clients\")", m.GetName())
```
noahdietz, line 189: Why not just `return nil, fmt.Errorf...)`?

noahdietz, line 189: Use backticks instead of double quotes for the `p()` argument so that you can use double quotes in the generated code without escaping them.

internal/gengapic/genrest.go
```diff
@@ -144,23 +143,152 @@ func (g *generator) genRESTMethod(servName string, serv *descriptor.ServiceDescr
...
+		return err
+	}
+	g.imports[servSpec] = true
+
+	lowcaseServName := lowcaseRestClientName(servName)
+
+	p("func (c *%s) %s(ctx context.Context, opts ...gax.CallOption) (%s.%s_%sClient, error) {",
+		lowcaseServName, m.GetName(), servSpec.Name, s.GetName(), m.GetName())
+	p("  var resp %s.%s_%sClient", servSpec.Name, s.GetName(), m.GetName())
+	p("  return resp, fmt.Errorf(\"%s not yet supported for REST clients\")", m.GetName())
```
noahdietz, line 215: Same things here.

noahdietz (commented):
Everywhere there are escaped double quotes, please use back ticks for the entire format string so that you don't have to escape the double quotes.

internal/gengapic/genrest.go
software-dov: The client is a value type, isn't it?

noahdietz: Nope, its an interface, example: https://github.com/googleapis/go-genproto/blob/bf773b8c83843d36cf305c12f292d88251b5e690/googleapis/cloud/bigquery/storage/v1/storage.pb.go#L1103-L1106

internal/gengapic/genrest.go
```diff
@@ -144,23 +143,152 @@ func (g *generator) genRESTMethod(servName string, serv *descriptor.ServiceDescr
...
 	default:
-		// TODO(dovs)
 		return g.unaryRESTCall(servName, m)
 	}
 }
 
+func (g *generator) serverStreamRESTCall(servName string, s *descriptor.ServiceDescriptorProto, m *descriptor.MethodDescriptorProto) error {
+	// Streaming calls will most likely NEVER be supported for REST clients,
+	// but the interface signature must be preserved.
+	// Making sure not to call streaming methods on a REST client, or checking for
```
vchudnov-g, line 167: BiDi streaming will definitely never be supported over REST.
Unidirectional streaming may be. 
I would change the wording to "streaming calls are not currently supported..."

internal/gengapic/genrest.go
software-dov: My bad, it's an interface.

internal/gengapic/genrest.go
```diff
@@ -144,23 +143,152 @@ func (g *generator) genRESTMethod(servName string, serv *descriptor.ServiceDescr
...
+	// errors in a situation with mixed gRPC and REST clients, is left to user code.
```
vchudnov-g, line 168: I think we should error eagerly when calling an unimplemented REST method and minimize the burden on the developer. Sounds to me like that's what you're doing below, so your comment is slightly confusing. The user is already checking for errors. Maybe simplify to ~"Users should not call unimplemented REST methods, which will error"?

noahdietz (commented):
lgtm but I will let @vchudnov-g Approve

vchudnov-g (approved):
The code LGTM, but please do address my previous wording comments.

BTW, the paging implementation here is just for standard paging for now (not including the new GCE patterns), right?

internal/gengapic/genrest.go
software-dov: Changed wording to say "Unimplemented REST methods will always error."

software-dov:
> BTW, the paging implementation here is just for standard paging for now (not including the new GCE patterns), right?

Correct. This is just copy/pasted to set up the iterator for now.

PR #627, opened 2021-05-14
gapic: include import hint in all examples

noahdietz:
IDK why the original implementation of example import hints only generated the hint for methods with non-Empty response types...but that doesn't seem right because a Delete still has to provide a request message, which always comes from go-genproto. Fixing that oddity and some weird conditional organization.

### DIFF
--- a/internal/gengapic/testdata/empty_example.want
+++ b/internal/gengapic/testdata/empty_example.want
@@ -11,6 +11,8 @@ func ExampleNewClient() {
 }
 
 func ExampleClient_GetEmptyThing() {
+	// import mypackagepb "mypackage"
+
 	ctx := context.Background()
 	c, err := Foo.NewClient(ctx)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


internal/gengapic/testdata/empty_example.want
```diff
@@ -351,6 +353,8 @@ func ExampleClient_GetOperation() {
 }
 
 func ExampleClient_DeleteOperation() {
+	// import longrunningpb "google.golang.org/genproto/googleapis/longrunning"
```
codyoss, line 356: Should the `// import mypackagepb "mypackage"` also be included in addition?

noahdietz: In this example, no, because this is a "mix in" method - an RPC form the Operations service we are surfacing on the client b.c the API exposes it.

codyoss (approved):

PR #635, opened 2021-05-17
chore: don't include import comments in examples

tbpg:
Imports are automatically detected and included on pkg.go.dev. We don't
need to duplicate a subset of the imports as comments in the example
code.

For example, see
https://pkg.go.dev/cloud.google.com/go@v0.81.0/channel/apiv1#example-CloudChannelClient.ChangeRenewalSettings.


codyoss (approved):
LGTM

noahdietz (approved):

PR #636, opened 2021-05-17
chore: fix label sync config file name

tbpg:
See
https://github.com/googleapis/repo-automation-bots/blob/0008d5d3d5b532225881c2f78baeb6d5f6388756/packages/label-sync/src/label-sync.ts#L29.

Also see https://github.com/googleapis/repo-automation-bots/blob/0008d5d3d5b532225881c2f78baeb6d5f6388756/packages/auto-label/src/auto-label.ts#L215, to confirm the Auto Label config file is `auto-label.yaml`.

cc @tmatsuo who was looking into a general solution for this.


codyoss (approved):
LGTM, good catch!

PR #647, opened 2021-05-26
feat: headers, utils, and REST client construction

software-dov:
GAPICs that build REST clients now include factory functions that
create REST clients, as well as commensurate utilities.

### DIFF
--- a/internal/gengapic/genrest.go
+++ b/internal/gengapic/genrest.go
@@ -58,6 +58,7 @@ func (g *generator) restClientInit(serv *descriptor.ServiceDescriptorProto, serv
 	g.imports[pbinfo.ImportSpec{Path: "bytes"}] = true
 	g.imports[pbinfo.ImportSpec{Path: "google.golang.org/grpc/metadata"}] = true
 	g.imports[pbinfo.ImportSpec{Name: "httptransport", Path: "google.golang.org/api/transport/http"}] = true
+	g.imports[pbinfo.ImportSpec{Path: "google.golang.org/api/option/internaloption"}] = true
 }
 
 func (g *generator) genRESTMethods(serv *descriptor.ServiceDescriptorProto, servName string) error {
@@ -80,12 +81,13 @@ func (g *generator) restClientOptions(serv *descriptor.ServiceDescriptorProto, s
 	p := g.printf
 
 	var host string
-	if eHost, err := proto.GetExtension(serv.Options, annotations.E_DefaultHost); err == nil {
-		host = *eHost.(*string)
-	} else {
+	eHost, err := proto.GetExtension(serv.Options, annotations.E_DefaultHost)
+	if err != nil {
 		fqn := g.descInfo.ParentFile[serv].GetPackage() + "." + serv.GetName()
 		return fmt.Errorf("service %q is missing option google.api.default_host", fqn)
 	}
+
+	host = *eHost.(*string)
 
 	p("func default%sRESTClientOptions() []option.ClientOption {", servName)
 	p("  return []option.ClientOption{")
@@ -140,11 +142,18 @@ func (g *generator) restClientUtilities(serv *descriptor.ServiceDescriptorProto,
 	p("}")
 	p("")
 
+	// Close method
+	p("// Close closes the connection to the API service. The user should invoke this when")
+	p("// the client is no longer required.")
 	p("func (c *%s) Close() error {", lowcaseServName)
+	p("    c.httpClient.CloseIdleConnections()")
 	p("    return nil")
 	p("}")
 	p("")
 
+	p("// Connection returns a connection to the API service.")
+	p("//")
+	p("// Deprecated.")
 	p("func (c *%s) Connection() *grpc.ClientConn {", lowcaseServName)
 	p("    return nil")
 	p("}")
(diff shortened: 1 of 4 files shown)


internal/gengapic/genrest.go
```diff
@@ -67,25 +76,78 @@ func (g *generator) genRESTMethods(serv *descriptor.ServiceDescriptorProto, serv
...
+	if eHost, err := proto.GetExtension(serv.Options, annotations.E_DefaultHost); err == nil {
+		host = *eHost.(*string)
+	} else {
+		fqn := g.descInfo.ParentFile[serv].GetPackage() + "." + serv.GetName()
+		return fmt.Errorf("service %q is missing option google.api.default_host", fqn)
+	}
+
+	p("func default%sRESTClientOptions() []option.ClientOption {", servName)
+	p("  return []option.ClientOption{")
+	p("    internaloption.WithDefaultEndpoint(%q),", host)
```
vchudnov-g, line 92: Do we need an import for `internaloption`? I see one in gengrpc.go, but not in genrest.go.
(Also, I seem to be missing this in the `*.want` golden files.)

vchudnov-g, line 92: Do library users have the ability to manually set the protocol to be HTTP rather than HTTPS at client instantiation time? If not, until we do have such an ability, would this be the right place to automatically set it to be HTTP if the endpoint starts with `localhost`? (This would be a stop-gap measure until we can have users do it at client creation time; I can point you to the internal doc on the topic. Or see the Python PR....)

This would be needed to hook up the Showcase integration tests; we can make those two changes at the same time.

internal/gengapic/genrest.go
```diff
@@ -67,25 +76,78 @@ func (g *generator) genRESTMethods(serv *descriptor.ServiceDescriptorProto, serv
...
+
+	g.restClientOptions(serv, servName)
+
+	// setGoogleClientInfo method
+	p("// setGoogleClientInfo sets the name and version of the application in")
+	p("// the `x-goog-api-client` header passed on each request. Intended for")
+	p("// use by Google-written clients.")
+	p("func (c *%s) setGoogleClientInfo(keyval ...string) {", lowcaseServName)
+	p(`  kv := append([]string{"gl-go", versionGo()}, keyval...)`)
+	p(`  kv = append(kv, "gapic", versionClient, "gax", gax.Version, "rest", "UNKNOWN")`)
```
vchudnov-g, line 138: Maybe use `0.0` rather than `UNKNOWN` to make it clear this is an explicit choice (that way our data pipelines can fill in UNKNOWN for missing values, say)

internal/gengapic/genrest.go
```diff
@@ -324,6 +386,16 @@ func (g *generator) emptyUnaryRESTCall(servName string, m *descriptor.MethodDesc
...
 	p("")
+	p("// Set the headers")
+	p("for k, v := range c.xGoogMetadata {")
+	p("  httpReq.Header[k] = v")
+	p("}")
+	// The content type is separate from x-goog metadata.
+	// It's a little more verbose to set here explicitly, but
+	// it's conceptually cleaner.
+	// TODO(dovs) add field headers.
+	p(`httpReq.Header["content-type"] = []string{"application/json",}`)
```
vchudnov-g, line 397: nit: Suggest making the field name "canonical" as per the http package doc: Capitalized-Kebab-Case. Below as well.

vchudnov-g (commented):
lgtm, but have @noahdietz or @codyoss TAL as well.

internal/gengapic/genrest.go
codyoss: Lets keep this as UNKNOWN for now. We have other automation that post processes this value into something more meaningful

internal/gengapic/genrest.go
codyoss: At least in apiary land the default endpoints contain the scheme as well. So if a user wants non-https they just specify when they override it.

Example : 
- `internaloption.WithDefaultEndpoint(https://secretmanager.googleapis.com/)`
- ``internaloption.WithDefaultMTLSEndpoint(https://secretmanager.mtls.googleapis.com/)

internal/gengapic/testdata/empty_client_init.want
```diff
@@ -128,7 +128,48 @@ func (c *gRPCClient) Close() error {
...
+	kv := append([]string{"gl-go", versionGo()}, keyval...)
+	kv = append(kv, "gapic", versionClient, "gax", gax.Version, "rest", "UNKNOWN")
+	c.xGoogMetadata = metadata.Pairs("x-goog-api-client", gax.XGoogHeader(kv...))
+}
+
+func (c *restClient) Close() error {
+	return nil
+}
+
+func (c *restClient) Connection() *grpc.ClientConn {
```
codyoss, line 173: nit: docs

internal/gengapic/genrest.go
vchudnov-g: So this is something we'll be able to do immediately for REST clients at instantiation time, once this PR is merged? Perfect.

internal/gengapic/genrest.go
vchudnov-g: This post-processing happens before publication? OK. Like I say, once the library is in use I don't want to pollute the space of values we provide.

codyoss: Yup, it happens before the PR is raised to land the changes: https://github.com/googleapis/google-cloud-go/blob/master/internal/gapicgen/generator/gapics.go#L85-L97

codyoss (commented):
LGTM, I defer to your team for sign-off.

vchudnov-g (approved):

PR #648, opened 2021-06-02
chore: refactor protobuf Go use in testing module

noahdietz:
Simply switches use of protobuf-Go v1 with the v2 module in the `showcase` integration testing submodule.


codyoss (approved):
LGTM

Issue #435, opened 2020-08-27
gapic: add dependency inclusion tests


noahdietz:
We should add some tests for generator methods that need to introduce imports i.e. the code it generates needs a stdlib `os` import, but does that method include import in its own code? It would help catch things like #434

PR #651, opened 2021-06-04
gapic: ensure API pb import is included

noahdietz:
This ensures that even if there are no API-specific request/response messages used in a Service's RPCs, that the generated protobuf Go pkg is still imported to fulfill the grpc client stub import.

Adds a test for the imports the generator must have after client init generation, an example of #435


miraleung (approved):
Thanks!

codyoss (approved):

PR #655, opened 2021-06-08
gapic: fix enum header conversion

noahdietz:
Go protobuf code for enums makes them typed `int32` so in order to access the string representation of the enum, which is mapped with `int32` to `string`, we need to convert the enum value to an `int32` in the map key. 

This wasn't caught before, because the showcase proto with an enum http param wasn't being generated in the integration tests. It is added here.


codyoss (approved):
LGTM

PR #659, opened 2021-06-14
feat: generate complete URLs in REST client calls

software-dov:
feat: generate complete URLs in REST client calls
    
Urls are generated using path parameters and query parameters present
in the body of the request message, if a body is defined, or the
request message itself if not.

Includes tweaks for query param and path param construction.
Diffs in generated output are visible here: https://github.com/googleapis/gapic-showcase/pull/751


internal/gengapic/genrest.go
```diff
@@ -56,62 +80,515 @@ func (g *generator) genRESTMethods(serv *descriptor.ServiceDescriptorProto, serv
...
+	p("func (c *%s) %s(ctx context.Context, req *%s.%s, opts ...gax.CallOption) error {",
+		lowcaseServName, m.GetName(), inSpec.Name, inType.GetName())
+
+	// TODO(dovs): handle cancellation, metadata, osv.
+	// TODO(dovs): handle http headers
+	// TODO(dovs): handle deadlines
+	// TODO(dovs): handle call options
+	p("// The default (false) for the other options are fine.")
+	p("// Field names should be lowerCamel, not snake.")
+	p("m := protojson.MarshalOptions{AllowPartial: true, EmitUnpopulated: true, UseProtoNames: false}")
```
noahdietz, line 540: `UseProtoNames` should default to `false` - do we need to explicitly include it here?

internal/gengapic/genrest.go
```diff
@@ -56,62 +80,515 @@ func (g *generator) genRESTMethods(serv *descriptor.ServiceDescriptorProto, serv
...
+	p("")
+	p("// Set the headers")
+	p("for k, v := range c.xGoogMetadata {")
+	p("  httpReq.Header[k] = v")
+	p("}")
+	// The content type is separate from x-goog metadata.
+	// It's a little more verbose to set here explicitly, but
+	// it's conceptually cleaner.
+	// TODO(dovs) add field headers.
+	p(`httpReq.Header["Content-Type"] = []string{"application/json",}`)
```
noahdietz, line 565: I don't think we need that trailing `,` after `json"`

noahdietz (commented):
I've done a first pass!

showcase/compliance_test.go
```diff
@@ -0,0 +1,137 @@
+// Copyright 2021 Google LLC
```
vchudnov-g, line 1: Do you know why this file is showing up as "new"? It was introduced in https://github.com/googleapis/gapic-generator-go/pull/625....

vchudnov-g (commented):
First pass. My biggest confusion is the comment in genrest.go:237

.github/workflows/ci.yaml
```diff
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - master
+      - diregapic-dev-branch
```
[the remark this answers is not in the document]
software-dov: Yes...? I think this was deliberate and required to make CI work on the dev branch.

internal/gengapic/genrest.go
```diff
@@ -33,11 +41,27 @@ func (g *generator) restClientInit(serv *descriptor.ServiceDescriptorProto, serv
...
 	p("type %s struct {", lowcaseServName)
-	p("  host string")
+	p("  // The http endpoint to connect to.")
+	p("  endpoint string")
+	p("")
+	p("  // The http client.")
+	p("  httpClient *http.Client")
+	p("")
+	p("	// The x-goog-* metadata to be sent with each request.")
+	p("	 xGoogMetadata metadata.MD")
```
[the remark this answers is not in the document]
software-dov: Good eyesight

internal/gengapic/genrest.go
```diff
@@ -56,62 +80,515 @@ func (g *generator) genRESTMethods(serv *descriptor.ServiceDescriptorProto, serv
...
+	p := g.printf
+	lowcaseServName := lowcaseRestClientName(servName)
+	clientName := camelToSnake(serv.GetName())
+	clientName = strings.Replace(clientName, "_", " ", -1)
+
+	p("// New%sRESTClient creates a new %s rest client.", servName, clientName)
+	g.serviceDoc(serv)
+	p("func New%[1]sRESTClient(ctx context.Context, opts ...option.ClientOption) (*%[1]sClient, error) {", servName)
+	p("    clientOpts := default%sRESTClientOptions()", servName)
+	p("    httpClient, endpoint, err := httptransport.NewClient(ctx, append(clientOpts, opts...)...)")
```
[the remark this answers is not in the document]
software-dov: Good call.

internal/gengapic/genrest.go
```diff
@@ -56,62 +80,515 @@ func (g *generator) genRESTMethods(serv *descriptor.ServiceDescriptorProto, serv
...
+	if info.body != "" && info.body != "*" {
+		bodyStr = "body"
+		p("body := req.Get%s()", snakeToCamel(info.body))
+	}
+	p("queryParamStrs := []string{}")
+	// TODO(dovs): make this work for nested fields
+	for _, f := range fields {
+		field := queryParams[f]
+		f = snakeToCamel(f)
+		// Note: something very strange has happened
```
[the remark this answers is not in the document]
software-dov: It got truncated somehow during refactoring.

.github/workflows/ci.yaml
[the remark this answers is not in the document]
noahdietz: Ok cool!

showcase/compliance_test.go
noahdietz: This PR is against `master`

internal/gengapic/genrest.go
```diff
@@ -56,62 +80,515 @@ func (g *generator) genRESTMethods(serv *descriptor.ServiceDescriptorProto, serv
...
+			case descriptor.FieldDescriptorProto_TYPE_BYTES:
+				p("if %s.Get%s() != nil {", bodyStr, f)
+			case descriptor.FieldDescriptorProto_TYPE_STRING:
+				p(`if %s.Get%s() != "" {`, bodyStr, f)
+			case descriptor.FieldDescriptorProto_TYPE_BOOL:
+				p(`if %s.Get%s() {`, bodyStr, f)
+			default: // Handles all numeric types
+				p(`if %s.Get%s() != 0 {`, bodyStr, f)
+			}
+		}
```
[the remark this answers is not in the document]
software-dov: I think I like multiple cases in case best. While definitely more idiomatic in go, I'm generally hesitant to use `break`, `continue`, or `fallthrough`. Trying to do something clever to make the switch handle both field type and label backfired, so I'll keep them separate.

internal/gengapic/genrest.go
software-dov: I wanted to make it clear that this was a deliberate design choice. If you think it's cleaner without I'll remove it.

internal/gengapic/genrest.go
software-dov: Force of habit.

showcase/compliance_test.go
software-dov: What? No, that's totally wrong.

internal/gengapic/genrest.go
vchudnov-g: I don't mind being explicit here, but that's just my opinion.

internal/gengapic/genrest.go
```diff
@@ -56,62 +80,515 @@ func (g *generator) genRESTMethods(serv *descriptor.ServiceDescriptorProto, serv
...
+	for p, _ := range queryParams {
+		fields = append(fields, p)
+	}
+	sort.Strings(fields)
+	info, _ := getHTTPInfo(m)
+
+	bodyStr := "req"
+	if info.body != "" && info.body != "*" {
+		bodyStr = "body"
+		p("body := req.Get%s()", snakeToCamel(info.body))
```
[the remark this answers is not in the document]
software-dov: Okay, so to clarify, using pseudopython set notation:
pathParams = set(fields appearing in the url path)
bodyParams = set(

The companion showcase PR is here: https://github.com/googleapis/gapic-showcase/pull/751

internal/gengapic/genrest.go
```diff
@@ -56,62 +80,515 @@ func (g *generator) genRESTMethods(serv *descriptor.ServiceDescriptorProto, serv
...
+	// We want to iterate over fields in a deterministic order
+	// to prevent spurious deltas when regenerating gapics.
+	fields := make([]string, 0, len(queryParams))
+	for p, _ := range queryParams {
+		fields = append(fields, p)
+	}
+	sort.Strings(fields)
+	info, _ := getHTTPInfo(m)
+
+	bodyStr := "req"
```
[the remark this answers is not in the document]
software-dov: s/bodyStr/requestObject/

showcase/compliance_test.go
software-dov: Changed the base to diregapic-dev-branch. I think that clears this up.

internal/gengapic/genrest.go
```diff
@@ -56,62 +80,515 @@ func (g *generator) genRESTMethods(serv *descriptor.ServiceDescriptorProto, serv
...
+func (g *generator) queryParams(m *descriptor.MethodDescriptorProto) map[string]*descriptor.FieldDescriptorProto {
+	queryParams := map[string]*descriptor.FieldDescriptorProto{}
+	info, err := getHTTPInfo(m)
+	if info == nil || err != nil {
+		return queryParams
+	}
+	pathParams := g.pathParams(m)
+	// Minor hack: we want to make sure that the body parameter, if one exists,
+	// is NOT a query parameter.
+	pathParams[info.body] = nil
```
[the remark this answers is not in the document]
software-dov: Went with an explicit dummy field.

internal/gengapic/genrest.go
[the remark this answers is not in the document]
vchudnov-g: Correct, with the following minor caveats:
- bodyParams may overlap pathParams (eg in case of "*"). The server will prioritize the path param values
- query params: the set of fields you specified above, but each one subject to the restriction that it is encoded iff 
   ```
  ( (field.Value()!=field.DefaultValue())    ||      (field.isOptional() && field.isSet())   )
  ```

See https://cloud.google.com/endpoints/docs/grpc-service-config/reference/rpc/google.api#rules-for-http-mapping and my [comment](https://github.com/googleapis/gapic-showcase/pull/751/files#r652210323) on the Showcase PR

internal/gengapic/genrest.go
```diff
@@ -15,15 +15,23 @@
 package gengapic
 
 import (
+	"fmt"
+	"regexp"
+	"sort"
+	"strings"
+
+	"github.com/golang/protobuf/proto"
```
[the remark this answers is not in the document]
software-dov: Done.

internal/gengapic/genrest.go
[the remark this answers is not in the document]
vchudnov-g: But what's trippingme up is that your `queryParams()` above is implementing this logic, so it's just this `generateQueryString()` that is confusing.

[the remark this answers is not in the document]
vchudnov-g: BTW: I _think_ (please verify) that the condition I quote above for query params can be simplified in Go by seeing whether the proto message field is null or not (ie not using accessor methods, which try to hide that implementation detail). But it might arguably be more elegant to use accessors and have the explicit check.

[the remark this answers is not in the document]
software-dov: Okay, made the fix. Thanks for pointing this out, I don't know how I misread the spec so badly.

internal/gengapic/client_init_test.go
```diff
@@ -382,13 +382,13 @@ func TestClientInit(t *testing.T) {
...
-				{Path: "context"}:                                                                  true,
-				{Path: "google.golang.org/grpc"}:                                                   true,
-				{Path: "google.golang.org/grpc/metadata"}:                                          true,
 				{Name: "gtransport", Path: "google.golang.org/api/transport/grpc"}:                 true,
 				{Name: "longrunningpb", Path: "google.golang.org/genproto/googleapis/longrunning"}: true,
 				{Name: "lroauto", Path: "cloud.google.com/go/longrunning/autogen"}:                 true,
 				{Name: "mypackagepb", Path: "github.com/googleapis/mypackage"}:                     true,
+				{Path: "context"}:                         true,
+				{Path: "google.golang.org/grpc"}:          true,
+				{Path: "google.golang.org/grpc/metadata"}: true,
```
noahdietz, line 391: can you run `gofmt -w .` I think these should be aligned with the other values...not sure why they got moved down here from up there in the first place? Maybe revert this bit

internal/gengapic/client_init_test.go
```diff
@@ -397,18 +397,18 @@ func TestClientInit(t *testing.T) {
...
 			parameter: proto.String("go-gapic-package=path;mypackage,transport=grpc+rest"),
 			imports: map[pbinfo.ImportSpec]bool{
-				{Path: "bytes"}:                           true,
-				{Path: "context"}:                         true,
-				{Path: "google.golang.org/grpc"}:          true,
-				{Path: "google.golang.org/grpc/metadata"}: true,
-				{Name: "gtransport", Path: "google.golang.org/api/transport/grpc"}: true,
-				{Name: "mypackagepb", Path: "github.com/googleapis/mypackage"}:     true,
-				{Path: "fmt"}: true,
+				{Name: "gtransport", Path: "google.golang.org/api/transport/grpc"}:    true,
```
noahdietz, line 400: I think gofmt will fix alignment here too

internal/gengapic/gengapic.go
```diff
@@ -297,6 +297,13 @@ func buildAccessor(field string) string {
 func (g *generator) lookupField(msgName, field string) *descriptor.FieldDescriptorProto {
 	var desc *descriptor.FieldDescriptorProto
 	msg := g.descInfo.Type[msgName]
+
+	// It could be that "msgName" is a field that is not a message.
+	// Fail cleanly in that case.
+	if msg == nil {
+		return desc
+	}
```
noahdietz, line 305: I don't understand this case, can you share an example? The `msgName` is supposed to be the fully-qualified name of the Message to look up and then find the field in.

noahdietz:
Looking good to me

internal/gengapic/gengapic.go
software-dov: This is a minor workaround for a specific case I ran into during my fixes to `queryParams`.
In `queryParams`, I want to get the field pointed to by the `body` parameter and remove all of its fields from the set of query params.
If the `body` parameter points to a field in the request that is a primitive type, it's cleaner to just call `lookupField` and let it handle the lookup failure instead of changing the logic in `queryParams`, and it adds a little bit of error checking.

If you would prefer this function not be augmented so I'll add a comment about the preconditions and change the logic in `queryParams`. I'd like to keep it like this, though, if possible.

internal/gengapic/client_init_test.go
software-dov: I'm not sure either. `gofmt` does not want to change the indentation, and it reverts the change when I attempt to fix it up manually.
I don't think I can revert this bit; I needed to remove the dependency on `strings` for these cases, and apparently that triggered the gofmt failure.

internal/gengapic/client_init_test.go
software-dov: Same as above; `gofmt` is fighting me here.

internal/gengapic/client_init_test.go
noahdietz: huh ok I wonder if the lines are too long and it doesn't try to align anymore. All good!

internal/gengapic/gengapic.go
noahdietz: Nah that's fine, but maybe change the comment to be more generic like "if the msg doesn't exist ..."

noahdietz (approved):

PR #661, opened 2021-06-15
gcli: fix paginated methods with primitive result types

odsod:
Fixes an issue where the early continue caused the subcommand file to
not be written, and the remaining required imports to be omitted.


noahdietz (approved):

noahdietz:
Released in v0.20.4

PR #662, opened 2021-06-16
fix(bazel): Remove monolith Bazel rule deps [go]

miraleung:
This is needed for the preliminary removal of the monolith rules from googleapis' `switched_rules_by_language`.


noahdietz:
Would you mind testing this with `local_repository` in `googleapis`? We don't have a test for the `go_gapic_library` Bazel target itself in this repo yet...😬

noahdietz (approved):

miraleung:
Tested by building `third_party/googleapis/google/cloud/asset/v1` .

PR #664, opened 2021-06-16
gapic: gen protobuf v2 import for iterator usage

### LINKED ISSUE #663: gapic: switch generated paging code to using protobuf v2 API
The generated code for paging calls utilizes `proto.Clone`, `proto.String`, and `proto.Int32` ([here](https://github.com/googleapis/gapic-generator-go/blob/master/internal/gengapic/paging.go#L184-L204)) from `github.com/golang/protobuf/proto` and includes the import [here](https://github.com/googleapis/gapic-generator-go/blob/master/internal/gengapic/paging.go#L243).

This should be switched to include `google.golang.org/protobuf/proto`, the v2 module. The generated usage of `proto` doesn't need to change, just the import.

See https://github.com/googleapis/google-cloud-go/issues/4273.

noahdietz:
Fixes #663

### DIFF
--- a/internal/gengapic/paging.go
+++ b/internal/gengapic/paging.go
@@ -240,7 +240,7 @@ func (g *generator) pagingCall(servName string, m *descriptor.MethodDescriptorPr
 	p("}")
 	p("")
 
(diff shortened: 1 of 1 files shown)


codyoss (approved):
LGTM

PR #669, opened 2021-06-18
fix: don't cause an error on http.StatusOK

software-dov:


noahdietz (approved):

showcase/main_test.go
```diff
@@ -70,8 +70,8 @@ func TestMain(m *testing.M) {
 
 	// TODO: Change to use REST client once query params are cleaned up.
 	// The custom endpoint bypasses https.
-	// complianceClient, err = showcase.NewComplianceRESTClient(ctx, option.WithEndpoint("http://localhost:7469"), option.WithoutAuthentication())
```
noahdietz, line 73: If we are switching to REST can we remove the comments/TODOs?

vchudnov-g (approved):

showcase/main_test.go
software-dov: Done

PR #675, opened 2021-06-23
gapic: fix rest-only client generation

noahdietz:
REST-only generation had some compilation errors (missing imports, references to unknown identifiers, Operation wrapper factory method collisions, etc.), so this fixes that. 

Adds support for specifying the `transport` via `go_gapic_library`, defaulting to `["grpc"]`.

Adds some Go dependencies to `go_gapic_library` necessary for REST clients.


codyoss (approved):

showcase/go.mod
```diff
@@ -3,6 +3,7 @@ module showcase
 go 1.13
 
 require (
+	github.com/golang/protobuf v1.5.2
```
noahdietz, line 6: Not sure why that was added, but its probably needed if it was...

noahdietz: Found it, fixing!

internal/gengapic/client_init_test.go
```diff
@@ -343,13 +344,16 @@ func TestClientInit(t *testing.T) {
...
-				{Path: "fmt"}:   true,
-				{Path: "google.golang.org/api/option/internaloption"}:   true,
-				{Path: "google.golang.org/grpc/metadata"}:               true,
-				{Path: "google.golang.org/protobuf/encoding/protojson"}: true,
-				{Path: "io/ioutil"}: true,
-				{Path: "net/http"}:  true,
+				{Path: "bytes"}:                        true,
+				{Path: "context"}:                      true,
+				{Path: "fmt"}:                          true,
+				{Path: "google.golang.org/api/option"}: true,
```
vchudnov-g, line 350: nit: right column alignment weird

rules_go_gapic/go_gapic.bzl
```diff
@@ -110,8 +111,13 @@ def go_gapic_library(
     for path in samples:
         file_args[path] = "sample"
 
+  t = transport[0]
+  if len(transport) > 1:
```
vchudnov-g, line 115: do you need this line and the line above? can we simply have `"+".join(transport)` work for 0- and 1-element lists?

test.sh
```diff
@@ -36,11 +36,11 @@ generate() {
 	protoc --go_gapic_out "$OUT" -I "$GOOGLEAPIS" $*
 }
 
-echo "Generating Cloud KMS v1"
-generate --go_gapic_opt 'go-gapic-package=cloud.google.com/go/kms/apiv1;kms,transport=grpc+rest' $GOOGLEAPIS/google/cloud/kms/v1/*.proto
+echo "Generating Cloud KMS v1 - gRPC"
+generate --go_gapic_opt 'go-gapic-package=cloud.google.com/go/kms/apiv1;kms,transport=grpc' $GOOGLEAPIS/google/cloud/kms/v1/*.proto
```
vchudnov-g, line 40: why the switch to only one transport in this file?

vchudnov-g (commented):
lgtm. One non-trivial(?) question in test.sh

I suggest getting @software-dov 's eyes on this so we're all on the same page.

test.sh
noahdietz: Because the problems i noticed arose when only rest was generated - with grpc+rest, the grpc code paths were including some imports that rest also needed, but when grpc was removed, the imports were not included. That is just one example. Thus, we want to make sure we do both in isolation. The following test in this file does both.

internal/gengapic/client_init_test.go
software-dov: I ran into this exact thing myself. I think that `gofmt` may do indentation by tabs for map literals

software-dov (approved):

internal/gengapic/client_init_test.go
noahdietz: yeah it's a `gofmt` decision, we are just obeying ;)

vchudnov-g (approved):

rules_go_gapic/go_gapic.bzl
noahdietz: I'm not a python pro, and when I was testing in an online interpreter, a single element list resulted in the single string being split by character and joined by `+` haha so `["rest"]` became `"r+e+s+t"` oddly enough.

vchudnov-g: I think it depends on whether `transport` is a list of strings or just a string (a sequence of chars). The following give different results:
```
transport="rest"
print("+".join(transport))
print("+".join([transport]))
```
=>
```
❯ python testsplit.py
r+e+s+t
rest
```

PR #676, opened 2021-06-24
feat: properly detect, reference, and encode leaf query parameters

software-dov:
Generate a proper percent-encoded HTTP query string built from all leaf fields of the request object,
except for fields matching any of these criteria:
- fields mentioned in the matched path template and in the body field
- proto3 `optional` fields that have not been explicitly set
- non proto3 `optional` fields that have default value (according to
  the proto3 spec), unless they are required fields (even if they have default values).


internal/gengapic/genrest.go
```diff
@@ -210,33 +210,94 @@ func (g *generator) queryParams(m *descriptor.MethodDescriptorProto) map[string]
...
+	OUTER:
+		for _, field := range m.GetField() {
+			if field.GetType() == descriptor.FieldDescriptorProto_TYPE_MESSAGE {
+				if field.GetLabel() == descriptor.FieldDescriptorProto_LABEL_REPEATED {
+					// Repeated message fields must not be mapped because no
+					// client library can support such complicated mappings.
+					// https://cloud.google.com/endpoints/docs/grpc-service-config/reference/rpc/google.api#grpc-transcoding
+					continue OUTER
+				}
+				for _, f := range excludedFields {
```
noahdietz, line 264: Nit: maybe rename `f` to `excluded` so that the comparison is `if field == excluded` - reads a bit better.

internal/gengapic/genrest.go
```diff
@@ -210,33 +210,94 @@ func (g *generator) queryParams(m *descriptor.MethodDescriptorProto) map[string]
...
+//             int32 mass_kg = 1;
+//         }
+//         Mantle mantle = 1;
+//     }
+//     Squid squid = 1;
+// }
+//
+// The one entry would be
+// "squid.mantle.mass_kg": *descriptor.FieldDescriptorProto...
+func (g *generator) getLeafs(msg *descriptor.DescriptorProto, excludedFields ...*descriptor.FieldDescriptorProto) map[string]*descriptor.FieldDescriptorProto {
```
noahdietz, line 245: Why is `excludedFields` variadic when the only call site provides one item? Seems like it doesn't need to be variadic.

internal/gengapic/genrest.go
```diff
@@ -210,33 +210,94 @@ func (g *generator) queryParams(m *descriptor.MethodDescriptorProto) map[string]
...
+				for _, p := range stack {
+					if p == field {
+						continue OUTER
+					}
+				}
+
+				subMsg := g.descInfo.Type[field.GetTypeName()].(*descriptor.DescriptorProto)
+				recurse(append(stack, field), subMsg)
+			} else {
+				elts := []string{}
```
noahdietz, line 279: what does `elts` mean?

internal/gengapic/genrest.go
```diff
@@ -210,33 +210,94 @@ func (g *generator) queryParams(m *descriptor.MethodDescriptorProto) map[string]
...
+	pathsToLeafs := map[string]*descriptor.FieldDescriptorProto{}
+
+	var recurse func([]*descriptor.FieldDescriptorProto, *descriptor.DescriptorProto)
+
+	recurse = func(
+		stack []*descriptor.FieldDescriptorProto,
+		m *descriptor.DescriptorProto,
+	) {
+
+	OUTER:
```
noahdietz, line 255: Maybe it's just me, but this code is kind of hard to interpret. Is there a simpler approach that doesn't necessitate labels? What are we trying to achieve?

internal/gengapic/genrest.go
```diff
@@ -245,36 +306,40 @@ func (g *generator) generateQueryString(m *descriptor.MethodDescriptorProto) {
...
 			case descriptor.FieldDescriptorProto_TYPE_BOOL:
-				p(`if req.Get%s() {`, f)
-			default: // Handles all numeric types
-				p(`if req.Get%s() != 0 {`, f)
+				p(`if %s%s {`, requestObject, accessor)
+			default: // Handles all numeric types including enums
+				p(`if %s%s != 0 {`, requestObject, accessor)
 			}
+		} else {
+			// This is a required field, so we must include the field
```
noahdietz, line 334: If the field _isn't_ labeled with proto3_optional, that doesn't mean its _required_ does it? I guess our best indicator is `option (google.api.field_behavior) = REQUIRED`?

internal/gengapic/genrest.go
```diff
@@ -295,14 +360,15 @@ func (g *generator) generateURLString(m *descriptor.MethodDescriptorProto) error
...
 	// TODO(dovs): handle more complex path urls involving = and *,
 	// e.g. v1beta1/repeat/{info.f_string=first/*}/{info.f_child.f_string=second/**}:pathtrailingresource
 	re := regexp.MustCompile(`{([a-zA-Z0-9_.]+?)}`)
 	fmtStr = re.ReplaceAllStringFunc(fmtStr, func(s string) string { return "%v" })
 
-	p(`url := fmt.Sprintf("%s",`, fmtStr)
-	p("c.endpoint,")
+	// TODO(dovs): handle error
+	p("baseUrl, _ := url.Parse(c.endpoint)")
+	p(`baseUrl.Path += fmt.Sprintf("%s",`, fmtStr)
```
noahdietz, line 371: Are we guaranteed to have such params? if the regex matching comes up with nothing, this becomes uncompilable code.

internal/gengapic/genrest.go
software-dov: It's the shorthand I use for `elements`. In this situation, it's just a one to one correspondence to field names. I felt that there were already too many names using "field", so I went with "elts".

internal/gengapic/genrest.go
software-dov: There's a few things we're trying to do with the labeled `continue`.
Skip explicitly excluded fields.
Skip over recursively nested messages.
Skip over repeated message fields.

The first two are an annoying workaround because there isn't a slice `Find` method. I could do a little something to clean that up.

internal/gengapic/genrest.go
software-dov: You're right, I misread the enum and thought that `OPTIONAL` was the default enum value, when `0` is explicitly reserved for errors.

showcase/compliance_test.go
```diff
@@ -58,10 +58,12 @@ func TestComplianceSuite(t *testing.T) {
 	// restRPCs is a map of the RPC names as they appear in the compliance suite file to the
 	// actual Go methods that implement them.
 	restRPCs := map[string]method{
-		"Compliance.RepeatDataBody":                 complianceClient.RepeatDataBody,
-		"Compliance.RepeatDataBodyInfo":             complianceClient.RepeatDataBodyInfo,
-		"Compliance.RepeatDataQuery":                wipMethod, // TODO: replace with complianceClient.RepeatDataPathResource,
-		"Compliance.RepeatDataSimplePath":           wipMethod, // TODO: replace with complianceClient.RepeatDataPathResource,
+		"Compliance.RepeatDataBody":     complianceClient.RepeatDataBody,
+		"Compliance.RepeatDataBodyInfo": complianceClient.RepeatDataBodyInfo,
+		// TODO(dovs): Broken subtests are "Extreme values", "zero values for non-string fields", "Zero values for all fields"
```
vchudnov-g, line 63: Thanks for listing the broken/failing testcases. Are these unimplemented in the generator or Showcase failures?

vchudnov-g (commented):
Do you have the generated client code somewhere we can see it? That will help the review.

showcase/compliance_test.go
software-dov: These are showcase failures.

internal/gengapic/genrest.go
software-dov: I don't believe that's the case. If the regex comes up with nothing, then the `ReplaceAllStringFunc` invocation won't add any format string markers, and the result of `FindAllStringSubmatch` will be an empty slice.

internal/gengapic/genrest.go
```diff
@@ -210,33 +210,94 @@ func (g *generator) queryParams(m *descriptor.MethodDescriptorProto) map[string]
...
+	) {
+
+	OUTER:
+		for _, field := range m.GetField() {
+			if field.GetType() == descriptor.FieldDescriptorProto_TYPE_MESSAGE {
+				if field.GetLabel() == descriptor.FieldDescriptorProto_LABEL_REPEATED {
+					// Repeated message fields must not be mapped because no
+					// client library can support such complicated mappings.
+					// https://cloud.google.com/endpoints/docs/grpc-service-config/reference/rpc/google.api#grpc-transcoding
+					continue OUTER
```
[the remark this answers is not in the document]
software-dov: In this case I think regular `continue` works, but I want to make it clear and uniform the cases when we're skipping over a field and why. I don't think this is an infinite loop, but I can whip up an additional test to verify.

internal/gengapic/genrest.go
software-dov: Good call. I think this check and the next one will be lifted into helper funcs anyway.

internal/gengapic/genrest.go
software-dov: This is a combination of laziness and future proofing on my part. I want to be able to exclude the `body` field when calling `getLeafs` from `queryParams`, but I didn't want to rewrite the tests to support the parameter. Making `excludedFields` variadic is a cheater way of making it optional.

internal/gengapic/genrest.go
```diff
@@ -210,33 +210,94 @@ func (g *generator) queryParams(m *descriptor.MethodDescriptorProto) map[string]
...
+//     }
+//     Squid squid = 1;
+// }
+//
+// The one entry would be
+// "squid.mantle.mass_kg": *descriptor.FieldDescriptorProto...
+func (g *generator) getLeafs(msg *descriptor.DescriptorProto, excludedFields ...*descriptor.FieldDescriptorProto) map[string]*descriptor.FieldDescriptorProto {
+	pathsToLeafs := map[string]*descriptor.FieldDescriptorProto{}
+
+	var recurse func([]*descriptor.FieldDescriptorProto, *descriptor.DescriptorProto)
```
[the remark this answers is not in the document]
software-dov: Done

internal/gengapic/genrest.go
noahdietz: I meant that the generated code becomes uncompilable, it would look like this: https://play.golang.org/p/1Q8RTVsgSg_U

internal/gengapic/genrest.go
```diff
@@ -245,32 +245,37 @@ func (g *generator) queryParams(m *descriptor.MethodDescriptorProto) map[string]
 func (g *generator) getLeafs(msg *descriptor.DescriptorProto, excludedFields ...*descriptor.FieldDescriptorProto) map[string]*descriptor.FieldDescriptorProto {
 	pathsToLeafs := map[string]*descriptor.FieldDescriptorProto{}
 
+	isInSliceP := func(fields []*descriptor.FieldDescriptorProto, field *descriptor.FieldDescriptorProto) bool {
```
noahdietz, line 248: Nit: s/`isInSliceP`/`contains`?

internal/gengapic/genrest.go
```diff
@@ -309,15 +314,18 @@ func (g *generator) generateQueryString(m *descriptor.MethodDescriptorProto) {
...
-	// TODO(dovs): make this work for nested fields
 	for _, path := range fields {
 		field := queryParams[path]
 		accessor := buildAccessor(path)
 		closer := "}"
 		if field.GetLabel() == descriptor.FieldDescriptorProto_LABEL_REPEATED {
 			// It's a slice, so check for nil
 			p("if %s%s != nil {", requestObject, accessor)
-		} else if field.GetLabel() == descriptor.FieldDescriptorProto_LABEL_OPTIONAL {
+		} else if field.GetLabel() == descriptor.FieldDescriptorProto_LABEL_REQUIRED {
```
noahdietz, line 324: Ah so `google.api.field_behavior` is an annotation. The label in proto3 is different. We shouldn't be using this one. If you want to see if its annotated as required, check for the annotation. But if you want to check for `proto3_optional`, use the `GetProto3Optional()` API.

internal/gengapic/genrest.go
software-dov: Yes, that's a much better name :P

internal/gengapic/genrest.go
software-dov: Damn, I always forget this

internal/gengapic/genrest.go
software-dov: Took me a little while to understand, but found a workaround.

software-dov:
Companion PR is https://github.com/googleapis/gapic-showcase/pull/770

vchudnov-g:
Left some comments on the companion PR.

software-dov:
Made some changes based on companion PR review (thanks VIctor!) that caused the compliance Query tests to all pass. PTAL

internal/gengapic/genrest.go
```diff
@@ -210,33 +210,99 @@ func (g *generator) queryParams(m *descriptor.MethodDescriptorProto) map[string]
...
+				recurse(append(stack, field), subMsg)
+			} else {
+				elts := []string{}
+				for _, f := range stack {
+					elts = append(elts, f.GetName())
+				}
+				elts = append(elts, field.GetName())
+				key := strings.Join(elts, ".")
+				pathsToLeafs[key] = field
+			}
```
noahdietz: This is total nit, feel free to ignore: I (as does Go style) prefer to reduce indentation where possible. `if-else` chains are a good opportunity for that. As a matter of preference, I'd suggest inverting the `if-else` to be `if!` + fall through. Personally I think its more readable but 🤷🏻‍♂️ 

```go
if field.GetType() != descriptor.FieldDescriptorProto_TYPE_MESSAGE {
  elts := []string{}
  // ...

  // this continue is critical to avoid the fall through into the Message type logic.
  continue
}

if field.GetLabel() == REPEATED { continue }

if contains(excludedFields, field) { continue }

if contains(stack, field) { continue }

subMsg := ...
recurse...
```

internal/gengapic/genrest.go
```diff
@@ -245,36 +311,39 @@ func (g *generator) generateQueryString(m *descriptor.MethodDescriptorProto) {
...
+		accessor := buildAccessor(path, false)
 		if field.GetLabel() == descriptor.FieldDescriptorProto_LABEL_REPEATED {
 			// It's a slice, so check for nil
-			p("if req.Get%s() != nil {", f)
+			p("if %s%s != nil {", requestObject, accessor)
+		} else if field.GetProto3Optional() {
+			// Split right before the raw access
+			elts := strings.Split(path, ".")
+			elts = elts[:len(elts)-1]
+			p("if %[1]s%[2]s != nil && %[1]s%[3]s != nil {", requestObject, buildAccessor(strings.Join(elts, "."), false), buildAccessor(path, true))
```
noahdietz, line 327: Nit: consider wrapping `buildAccessor(str, false)` and `buildAccesor(str, true)` in methods that convey what the boolean toggles i.e.

```go
func getter(s string) string {
  return buildAccessor(s, true)
}

func directAccess(s string) string {
  return buildAccessor(s, false)
}
```

Honestly might be an over reaction, but might read a little better 🤷🏻‍♂️

noahdietz, line 327: Could you put the results of the `buildAccessor` methods into local variables? This line is rather complex to grok at first glance what with the format specifier indices and nested function calls.

internal/gengapic/genrest.go
```diff
@@ -693,5 +762,7 @@ func (g *generator) unaryRESTCall(servName string, m *descriptor.MethodDescripto
 	p("return rsp, unm.Unmarshal(buf, rsp)")
 	p("}")
 
+	g.imports[inSpec] = true
```
noahdietz, line 765: Was this already handled by #678 ?

software-dov: Yes, don't know how this snuck in here.

internal/gengapic/genrest.go
software-dov: In this one particular case I think I'd like to keep the base case and the recursive case more firmly distinct. IMO it makes for a better match to a recursive traversal pattern. I'll try rearranging with less indentation to see how it looks, or maybe refactor the two cases into their own standalone functions.

noahdietz: Eh leave it the way it is in that case. I think that's a fair reason. Any more refactoring and it might end up getting too complicated. :)

internal/gengapic/genrest.go
software-dov: That's really helpful, actually. One thing I particularly miss from python/lisp is keyword params, which I think could have made calls to `buildAccessor` a little more clear.
Something like
```go
// Strawman syntax
func buildAccessor(s string, rawFinal bool = false) { ... }

buildAccessor(path)
buildAccessor(path, rawFinal=true)
```

software-dov: Sure.

internal/gengapic/genrest.go
software-dov: Okay, I don't know how idiomatic the new version is, but it definitely reduces indentation a little bit.

noahdietz (approved):
LGTM

PR #687, opened 2021-06-30
gapic: fix example use of factory methods for REST transport

noahdietz:
When `rest` only transport is targeted for generation, the generated example files do not adjust to use the `rest` client factory method instead of the gRPC one resulting in a compilation error when the _examples_ are compiled.

### DIFF
--- a/internal/gengapic/example.go
+++ b/internal/gengapic/example.go
@@ -24,18 +24,11 @@ import (
 	"google.golang.org/genproto/googleapis/longrunning"
 )
 
-func (g *generator) genExampleFile(serv *descriptor.ServiceDescriptorProto, pkgName string) error {
-	servName := pbinfo.ReduceServName(*serv.Name, pkgName)
-	p := g.printf
+func (g *generator) genExampleFile(serv *descriptor.ServiceDescriptorProto) error {
+	pkgName := g.opts.pkgName
+	servName := pbinfo.ReduceServName(serv.GetName(), pkgName)
 
-	p("func ExampleNew%sClient() {", servName)
-	g.exampleInitClient(pkgName, servName)
-	p("")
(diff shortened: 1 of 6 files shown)


codyoss (approved):
LGTM, thanks!

internal/gengapic/example.go
```diff
@@ -85,7 +98,14 @@ func (g *generator) exampleMethod(pkgName, servName string, m *descriptor.Method
 		return err
 	}
 
-	g.exampleInitClient(pkgName, servName)
+	// Pick the first transport for simplicity. We don't need examples
+	// of each method for both transports when they have the same surface.
```
vchudnov-g, line 102: Where do these examples go? If they go into `_test.go` files, then they show up in godoc alongside the init methods, right? In that case, it might be worthwhile to have one for each transport, even if the only difference is one name. WDYT?

noahdietz: They go into the `_example_test.go` files which are indeed included in the GoDoc, but there is only 1 exported client type, the transport agnostic one. Thus, there will only be docs for that type's methods. There is no difference in the usage of a method based on transport and the only difference is in the factory method used to create the client, which will have separate examples because they are separate publicly exported methods. It would be a one line difference but double the examples, and I don't think that's all that valuable.

vchudnov-g: ok, sg

vchudnov-g: One very minor idea for clients that have multiple transports: in the examples for non-factory methods, in these init lines, maybe we continue having only one transport but add a comment ` // use the Init method corresponding to your desired transport`. That way we avoid code duplication but also don't confuse people who may want to use the transport not cited in the example.

Very minor thing, we  don't need to do it now, and we can discuss.

Issue #691, opened 2021-07-01
gapic: metadata JSON diffs on whitespace and its rather annoying


noahdietz:
Example of the gapic_metadata.json diffing on whitespace: https://github.com/googleapis/google-cloud-go/pull/4360/files/a666083de1a4f4245473d79a5cb2eba701815169#r662355143. This shouldn't happen.

codyoss:
@noahdietz `protojson` on purpose provides non-deterministic json out(with whitespace) so users don't rely on its output. I am not sure it is documented but you can at least see this mentioned in the issue tracker. I only know this because I have been down this rabbit hole before 😸

noahdietz:
haha well boo! Thanks, I will try to figure something out...

PR #692, opened 2021-07-01
fix(gengapic): standardize output of metadata files

codyoss:
protojson produces non-deterministic output in the form of adding
random amounts of space after keys. This causes unneeded diffs in
google-cloud-go. This is a hack to make the output deterministic.

More context: https://go-review.googlesource.com/c/protobuf/+/151340
Fixes: #691

### DIFF
--- a/internal/gengapic/metadata.go
+++ b/internal/gengapic/metadata.go
@@ -15,16 +15,22 @@
 package gengapic
 
 import (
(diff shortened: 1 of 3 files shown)


noahdietz (approved):

noahdietz:
Thanks @codyoss ! Super helpful

PR #695, opened 2021-07-01
Revert "fix(gengapic): standardize output of metadata files"

noahdietz:
Reverts googleapis/gapic-generator-go#692. The test is not stable on CI and blocking other changes. We will come back to this.


codyoss (approved):
Whoops, will try this again.

PR #699, opened 2021-07-02
gapic: exclude body for get & delete

noahdietz:


software-dov (approved):

internal/gengapic/genrest.go
```diff
@@ -629,24 +627,30 @@ func (g *generator) emptyUnaryRESTCall(servName string, m *descriptor.MethodDesc
...
 	// TODO(dovs): handle deadlines
 	// TODO(dovs): handle call options
-	p("// The default (false) for the other options are fine.")
-	p("// Field names should be lowerCamel, not snake.")
-	p("m := protojson.MarshalOptions{AllowPartial: true, EmitUnpopulated: true, UseProtoNames: false}")
-	requestObject := "req"
-	if info.body != "" && info.body != "*" {
-		requestObject = "body"
-		p("body := req.Get%s()", snakeToCamel(info.body))
+	body := "nil"
```
vchudnov-g, line 630: nit: add comment "Marshal body for HTTP methods that take a body" Below as well.

noahdietz:
Regenerated the `compute` client locally and ran the same smoke test we debugged with and it worked as expected.

internal/gengapic/genrest.go
```diff
@@ -629,24 +627,30 @@ func (g *generator) emptyUnaryRESTCall(servName string, m *descriptor.MethodDesc
...
+	if info.verb != "get" && info.verb != "delete" {
+		p("// The default (false) for the other options are fine.")
+		p("// Field names should be lowerCamel, not snake.")
+		p("m := protojson.MarshalOptions{AllowPartial: true, EmitUnpopulated: true, UseProtoNames: false}")
+		requestObject := "req"
+		if info.body != "" && info.body != "*" {
```
vchudnov-g, line 636: I think this should be the main, outermost if statement instead of the http method checks themselves. GETs and DELETEs should not have body annotations, and RPCs using other HTTP methods may not as well.

In fact, we should have a generation error if there's a body annotation for GET or DELETE, because the query param logic will exclude those fields from the query string, so even if we silently omit a body, those fields will not get sent.

vchudnov-g (changes requested):

internal/gengapic/genrest.go
vchudnov-g: To be more precise, what I mean is
```
if info.body != "" {
  assert(method!="get" && method!="delete")
  // do everything else
  ...
}
```

noahdietz: > outermost if statement instead of the http method checks themselves

Yep, agreed.

internal/gengapic/genrest.go
noahdietz: Done.

internal/gengapic/genrest.go
noahdietz: Done.

internal/gengapic/genrest.go
```diff
@@ -714,23 +722,35 @@ func (g *generator) unaryRESTCall(servName string, m *descriptor.MethodDescripto
...
-	requestObject := "req"
-	if info.body != "" && info.body != "*" {
-		requestObject = "body"
-		p("body := req.Get%s()", snakeToCamel(info.body))
+
+	body := "nil"
+
+	// Marshal body for HTTP methods that take a body.
+	if info.body != "" {
+		if info.verb == "get" || info.verb == "detele" {
```
vchudnov-g, line 730: "delete"

internal/gengapic/genrest.go
```diff
@@ -714,23 +722,35 @@ func (g *generator) unaryRESTCall(servName string, m *descriptor.MethodDescripto
...
 	// TODO(dovs): handle calloptions
-	p("m := protojson.MarshalOptions{AllowPartial: true, EmitUnpopulated: true}")
-	requestObject := "req"
-	if info.body != "" && info.body != "*" {
-		requestObject = "body"
-		p("body := req.Get%s()", snakeToCamel(info.body))
+
+	body := "nil"
+
+	// Marshal body for HTTP methods that take a body.
```
vchudnov-g, line 728: Suggest adding a "// TODO: Have a test check that we generate requests with and without a body", since it looks like no test files are changing because of this fix.

vchudnov-g (approved):
Non-trivial typo: `"delete"` for string comparison. Otherwise LGTM.

internal/gengapic/genrest.go
```diff
@@ -629,24 +627,34 @@ func (g *generator) emptyUnaryRESTCall(servName string, m *descriptor.MethodDesc
...
+	if info.body != "" {
+		if info.verb == "get" || info.verb == "detele" {
```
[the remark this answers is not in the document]
noahdietz: Switched to forcing uppercase and comparing to the `http` constants.

internal/gengapic/genrest.go
noahdietz: Yep. Done.

PR #704, opened 2021-07-12
chore: mv integration test protos to tmp

noahdietz:
Apparently there is bugged code in an example application in googleapis: https://github.com/googleapis/gapic-generator-go/pull/703/checks?check_run_id=3049979631.

This moves the local download of googleapis to a separate directory so as to avoid being compiled via `go test ./...`


codyoss (approved):
LGTM

vchudnov-g:
IIUC, this is what we want long-term, even after we fix the buggy code? sgtm

vchudnov-g (approved):

noahdietz:
> IIUC, this is what we want long-term, even after we fix the buggy code? sgtm

Yeah I'm going to fix the bugged code. It's not ours but happened to pop up in our CI by happenstance

PR #711, opened 2021-07-16
feat: generator detects diregapic repeated/map field for paged methods

software-dov:
Certain APIs follow a different convention than AIP-4233 for paginated
methods:
* They permit either a page_size or max_results field with identical
semantics
* They allow pagination over maps.

The existing paging detection logic now explicitly checks if the
repeated field is mapped and fails if it is.
New paging detection logic checks for the updated semantics for use in
future generated code changes.


internal/gengapic/paging.go
```diff
@@ -161,6 +170,69 @@ func (g *generator) pagingField(m *descriptor.MethodDescriptorProto) (*descripto
 	return elemFields[0], nil
 }
 
+// diregapicPagingField reports the "resource field" to be iterated over by paginating method m
+// and the "num elements" field that tells the server the maximum number of elements to return per page.
+// Makes particular allowance for diregapic idioms: maps can be paginated over,
+// and either 'page_size' XOR 'max_results' are allowable fields in the request.
+func (g *generator) diregapicPagingField(m *descriptor.MethodDescriptorProto) (*descriptor.FieldDescriptorProto, *descriptor.FieldDescriptorProto, error) {
```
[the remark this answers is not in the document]
vchudnov-g: I would consider changing the name to something like `getPagingFields` because:
- plural: you're getting more than one field
- diregapic: this returns fields for either the diregapic or non-diregapic case; ie it handles both scenarios

internal/gengapic/paging_test.go
```diff
@@ -216,3 +248,244 @@ func TestIterTypeOf(t *testing.T) {
...
+	req := plugin.CodeGeneratorRequest{
+		Parameter: proto.String("go-gapic-package=path;mypackage,transport=rest"),
+		ProtoFile: []*descriptor.FileDescriptorProto{file},
+	}
+	g.init(&req)
+
+	for _, tst := range []struct {
+		mthd      *descriptor.MethodDescriptorProto
+		sizeField *descriptor.FieldDescriptorProto // A nil field means this is not a Diregapic paged method
+		iterField *descriptor.FieldDescriptorProto // A nil field means this is not a Diregapic paged method
```
vchudnov-g, line 473: nit: "DIREGAPIC" capitalization?

internal/gengapic/paging_test.go
```diff
@@ -216,3 +248,244 @@ func TestIterTypeOf(t *testing.T) {
...
+
+	req := plugin.CodeGeneratorRequest{
+		Parameter: proto.String("go-gapic-package=path;mypackage,transport=rest"),
+		ProtoFile: []*descriptor.FileDescriptorProto{file},
+	}
+	g.init(&req)
+
+	for _, tst := range []struct {
+		mthd      *descriptor.MethodDescriptorProto
+		sizeField *descriptor.FieldDescriptorProto // A nil field means this is not a Diregapic paged method
```
vchudnov-g, line 472: The `diregapicPagingField` function will return non-nils even  for correct but non-DIREGAPIC paging, IIUC?

internal/gengapic/paging.go
```diff
@@ -138,6 +138,15 @@ func (g *generator) pagingField(m *descriptor.MethodDescriptorProto) (*descripto
 			hasNextToken = true
 		}
 		if f.GetLabel() == descriptor.FieldDescriptorProto_LABEL_REPEATED {
+			if f.GetType() == descriptor.FieldDescriptorProto_TYPE_MESSAGE {
```
vchudnov-g, line 141: It seems like `diregapicPagingField()` covers a superset of functionality as `pagingField()` and similarly for tests, so there's some duplicated logic. Would it make sense to simply have one function with a flag `allowDIREGAPIC bool`?

internal/gengapic/paging.go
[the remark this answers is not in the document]
vchudnov-g: (or, in the spirit of the previous comment, making this essentially be _the_ `pagingFields()` function.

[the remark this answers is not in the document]
software-dov: Good point; named returns sound like the correct thing here.
You're right: the diregapic paging fields are a superset of the 'regular' paging fields, so it might make sense to use the same function for both. The one exception I can think of is if the response has multiple repeated fields. I'll think about that some more.

internal/gengapic/paging_test.go
software-dov: Correct. My original thought was to check for regular paging first then diregapic paging. Since it's a superset, however, it makes sense to just have the one function.

internal/gengapic/paging.go
software-dov: I think it would make sense to do that. This brings us back to the question of having an explicit diregapic flag, though. I'm in favor of not having a flag and instead treating it like diregapic pagination if it passes the predicate. The two differences are the name variant of page size field, i.e. `max_results`, and maps.

software-dov:
Integrated Victor's proposals. There is now a single `getPagingFields` method in the generator that does not differentiate between regular methods and diregapic methods.

noahdietz:
There are still a few open questions on the design for this. Can we nail those down before submitting code?

internal/gengapic/paging.go
```diff
@@ -86,82 +87,75 @@ func (g *generator) iterTypeOf(elemField *descriptor.FieldDescriptorProto) (*ite
...
+				continue
 			}
+			repeatedField = f
 		}
+
+		hasNextPageToken = hasNextPageToken || (f.GetName() == "next_page_token" && f.GetType() == descriptor.FieldDescriptorProto_TYPE_STRING)
 	}
-	return elemFields[0], nil
+
+	return repeatedField, pageSizeField, nil
```
vchudnov-g, line 155: How about:
```
if pagedField == nil || hasNextPageToken == nil {
  // not paging
  return nil, nil, nil
}
return pagedField, pageSizeField, nil
```

internal/gengapic/paging.go
```diff
@@ -86,82 +87,75 @@ func (g *generator) iterTypeOf(elemField *descriptor.FieldDescriptorProto) (*ite
...
+				if repeatedField.GetNumber() > f.GetNumber() {
+					return nil, nil, errors.E(nil, "found multiple repeated or map fields in message %q", m.GetOutputType())
+				}
+				// We want the _first_ repeated field to be one paged over.
+				continue
 			}
+			repeatedField = f
 		}
+
+		hasNextPageToken = hasNextPageToken || (f.GetName() == "next_page_token" && f.GetType() == descriptor.FieldDescriptorProto_TYPE_STRING)
```
vchudnov-g, line 152: Where do you use `hasNextPageToken`? I suggest something in my other comment below.

internal/gengapic/paging.go
```diff
@@ -86,82 +87,75 @@ func (g *generator) iterTypeOf(elemField *descriptor.FieldDescriptorProto) (*ite
...
-	if len(elemFields) > 1 {
-		// ensure that the first repeated field visited has the lowest field number
-		// according to https://aip.dev/4233, and use that one.
-		min := elemFields[0].GetNumber()
-		for _, elem := range elemFields {
-			if elem.GetNumber() < min {
-				return nil, fmt.Errorf("%s: can't pick repeated field in %s based on aip.dev/4233", *m.Name, outType.GetName())
+	hasNextPageToken := false
+	for _, f := range outMsg.GetField() {
+		if f.GetLabel() == descriptor.FieldDescriptorProto_LABEL_REPEATED {
```
vchudnov-g, line 137: You're not checking for map responses yet? I would either
(a) add a TODO and change the PR description to say you're deferring map-method detection or
(b) detect map responses here and now and add an error/TODO in the paging iterator for map responses

internal/gengapic/paging.go
vchudnov-g: Well, the function could have a parameter `allowDIREGAPIC bool`, and we can always set it to `true` at calling sites for now. If/when we introduce a flag for DIREGAPIC, that value will be taken from the flag. 

- Advantage: it marks and flags the DIREGAPIC-specific code right now, while it's fresh in our mind. I t also allows us to test the code with the parameter set to `false` right away.

- Disadvantage: until we do have a top-level flag, all the calling sites will have an extra parameter `(.., true)` essentially. (Though we should make it into a constant for comprehensibility: `(..., allowDIREGAPIC)`

Actually, here's an idea to get rid of that disadvantage: have `allowDIREGAPIC` be a (package?) var referenced *inside* the function(s) that have DIREGAPIC-specific code, and NOT a function param. That way we can still flag DIREGAPIC-specific sections of code, and because it's a test, we can set it to false and test non-DIREGAPIC flows (careful with parallel tests!). It only becomes a function param instead of a package var when we do have a DIREGAPIC flag to pass.

WDYT?

vchudnov-g:
Also, could you have a companion PR so we can see what the generated code is like? That is always very helpful. Thanks!

noahdietz (commented):
If I am understanding, this should have 0 affect on existing clients, correct? Thank you for breaking up the internal API changes and the feature work.

internal/gengapic/paging.go
noahdietz: > have allowDIREGAPIC be a (package?) var referenced inside the function(s) that have DIREGAPIC-specific code

Maybe it would be cleaner to add a new field in the `generator.options` struct that we use to carry both flags and other config at the generator-level. And if there is ever a plugin option added, it just gets wired up to that.

vchudnov-g: +1, even better! This will allow the tests to be modular without much worry, while preserving all the other benefits.

internal/gengapic/paging.go
```diff
@@ -86,82 +87,75 @@ func (g *generator) iterTypeOf(elemField *descriptor.FieldDescriptorProto) (*ite
...
-	// TODO: remove this once the next version of the Talent API is published.
-	//
-	// This is a workaround to disable auto-pagination for specifc RPCs in
-	// Talent v4beta1. The API team will make their API non-conforming in the
-	// next version.
-	//
-	// This should not be done for any other API.
-	if g.descInfo.ParentFile[m].GetPackage() == "google.cloud.talent.v4beta1" &&
-		(m.GetName() == "SearchProfiles" || m.GetName() == "SearchJobs") {
-		return nil, nil
```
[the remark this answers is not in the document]
software-dov: You are correct, it should not be. It slipped through my fingers as I pithed the method.

internal/gengapic/paging.go
```diff
@@ -86,82 +87,75 @@ func (g *generator) iterTypeOf(elemField *descriptor.FieldDescriptorProto) (*ite
...
 
 	inType := g.descInfo.Type[m.GetInputType()]
 	if inType == nil {
-		return nil, errors.E(nil, "cannot find message type %q, malformed descriptor?", m.GetInputType())
+		return nil, nil, errors.E(nil, "cannot find message type %q, malformed descriptor?", m.GetInputType())
 	}
 	inMsg, ok := inType.(*descriptor.DescriptorProto)
 	if !ok {
-		return nil, errors.E(nil, "expected %q to be message type, found %T", m.GetInputType(), inType)
+		return nil, nil, errors.E(nil, "cannot find message type %q, malformed descriptor?", m.GetInputType())
```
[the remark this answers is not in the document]
software-dov: It is changed because this method was originally separate instead of replacing `pagingField` and had different errors but it subsumed `pagingField`.

internal/gengapic/paging.go
software-dov: Before checking the options struct for diregapic mode, there's not a lot to do for detecting map responses here. They are also repeated; the main difference would be that we need to generate a custom pair type and use that for the iterator.

internal/gengapic/paging.go
software-dov: Good catch. This is a test lacuna.

internal/gengapic/paging.go
```diff
@@ -86,82 +87,75 @@ func (g *generator) iterTypeOf(elemField *descriptor.FieldDescriptorProto) (*ite
...
-		// according to https://aip.dev/4233, and use that one.
-		min := elemFields[0].GetNumber()
-		for _, elem := range elemFields {
-			if elem.GetNumber() < min {
-				return nil, fmt.Errorf("%s: can't pick repeated field in %s based on aip.dev/4233", *m.Name, outType.GetName())
+	hasNextPageToken := false
+	for _, f := range outMsg.GetField() {
+		if f.GetLabel() == descriptor.FieldDescriptorProto_LABEL_REPEATED {
+			if repeatedField != nil {
+				// Repeated fields are tacitly okay as long as the
```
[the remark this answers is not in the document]
software-dov: Yes, that's exactly what I meant :P
I am not sure if it's guaranteed that the order is preserved. The previous code certainly acted as if it was preserved, and the wording for AIP-4233 strongly implies that it is.

internal/gengapic/paging_test.go
software-dov: Just unilaterally removed the diregapic moniker from the code.

internal/gengapic/paging.go
software-dov: Not dropping the diregapic option, just holding off on this for a little bit. Will post code containing it soon.

vchudnov-g: OK. I think having the bool be always be set in the struct for now make sense, and having code here and in tests use it. And like Noah says, we can wire up to any eventual plugin option later.

internal/gengapic/paging.go
```diff
@@ -49,6 +49,15 @@ func (g *generator) iterTypeOf(elemField *descriptor.FieldDescriptorProto) (*ite
 			return &iterType{}, err
 		}
 
+		eMsg, ok := eType.(*descriptor.DescriptorProto)
+		if !ok {
+			return nil, errors.E(nil, "cannot find message type %q, malformed descriptor", eType)
+		}
+		// TODO: handle map entries correctly
+		if eMsg.GetOptions().GetMapEntry() {
+			return nil, errors.E(nil, "iterating over maps not supported yet: %q", eType)
```
noahdietz, line 58: Just to make sure, we won't really ever get to this code path today with `compute` because `getPagingFields` would say "this is not a paginated rpc", right? I'm just concerned with this breaking `compute` generation today.

vchudnov-g: I gather from the tests  that a map field is represented as a repeated set of (key,value) tuples, so in that case the old version of `pagingFields()` would have recognized this as a paging field (assuming `page_size`, `page_token`, and `next_page_token` were there) and this iterator would have done something? Am I interpreting correctly?

That said, we're in preview, so we do have the ability to make breaking changes right now.

vchudnov-g (approved):
I can't vouch for the representation of mapped fields in protos, but otherwise LGTM

internal/gengapic/paging.go
noahdietz: I'm not concerned with breaking changes to the client surface, I'm concerned with failure during generation, breaking the build.

vchudnov-g: Oh, OK.

@software-dov it would still be useful to have a companion PR with the generated code; it's easier top spot some issues there. And to Noah's point, were you able to run CI test locally to verify no breakages should occur?

noahdietz: We don't have a `map<string, Message>` in Showcase at the moment (you were working on that), so this specific case wouldn't be exercised I believe. A chicken in the egg problem. I would say instead of returning a generator error here, add code to `getPagingFields` that skips those RPCs that only have a `map` for pagination, treating them as normal Unary RPCs. This will of course cause a breaking change when we implement mapped pagination if we were to release this skipping behavior. Not sure this is a big deal though given the Compute client is in alpha for us.

noahdietz: [This is the hacky example](https://github.com/googleapis/gapic-generator-go/pull/716) of code that would skip mapped pagination.

vchudnov-g: True, yes. Probably the best thing is to skip mapped pagination for right now (that will unblock my upcoming Showcase changes) and then implement it shortly after that.

My very slight inclination is to not do a new release until we have both max_results and mapped_pagination in, but I'm fine doing a release without mapped pagination and then issuing a breaking change with it. After all, we are still alpha and have forewarned users.

noahdietz: > not do a new release until we have both max_results and mapped_pagination in

sgtm! But one reason we'd want to is that Showcase regen CI uses the latest released version of the generator. So without a released version that handles/skips mapped pagination, your changes will still fail. That's the chicken and the egg problem here.

software-dov: Fine by me. I'll add an escape hatch for mapped pagination and tweak the tests real quick.

noahdietz (approved):

PR #720, opened 2021-07-28
chore: update protoc verison in CI

noahdietz:


codyoss (approved):
LGTM, thank you!

codyoss:
We should make this the same in google-cloud-go too.

software-dov (approved):

PR #721, opened 2021-07-28
feat: add support for paginated maps

software-dov:
In certain APIs, paginated results can use maps as the repeated field,
with the following semantics:
* Keys in the map are always strings
* The keyspace is identical for all pages in the response
* Keys can map to empty values

This change makes changes to signatures, generated types, and
generated methods to support properly described paginated methods.

Iterators over maps return two element structs whose first element is
the key string and whose second element is the value elemnt.

Iterator types and container struct types are coalesced. There is
exactly one such definition for every value type, even if the value
type is used in multiple paged map fields.


software-dov:
Companion PR: https://github.com/googleapis/gapic-showcase/pull/813

internal/gengapic/paging.go
```diff
@@ -67,8 +57,33 @@ func (g *generator) iterTypeOf(elemField *descriptor.FieldDescriptorProto) (*ite
...
+				pt.mapValueTypeName = valueField.GetTypeName()
+			} else {
+				pt.mapValueTypeName = pbinfo.GoTypeForPrim[valueField.GetType()]
+			}
+			pt.elemTypeName = fmt.Sprintf("String%sPair", upperFirst(pt.mapValueTypeName))
+			pt.iterTypeName = pt.elemTypeName + "Iterator"
+		} else {
+			pt.elemTypeName = fmt.Sprintf("*%s.%s", imp.Name, typeName)
+			pt.iterTypeName = typeName + "Iterator"
+		}
```
noahdietz, line 86: Nit, style: Can you move this out of the `else` to just above the `if`? I prefer to do the base case (if it has no side effects by doing so) by default, then handle special cases with `if` following it. This keep sthe main flow of the program left-justified.

internal/gengapic/paging.go
```diff
@@ -67,8 +57,33 @@ func (g *generator) iterTypeOf(elemField *descriptor.FieldDescriptorProto) (*ite
...
+			}
+			if valueField == nil {
+				return nil, errors.E(nil, "unusual map entry message: %q", eMsg)
+			}
+
+			if valueField.GetType() == descriptor.FieldDescriptorProto_TYPE_MESSAGE {
+				pt.mapValueTypeName = valueField.GetTypeName()
+			} else {
+				pt.mapValueTypeName = pbinfo.GoTypeForPrim[valueField.GetType()]
+			}
```
noahdietz, line 80: Same comment here WRT `else` statement contents.

internal/gengapic/paging.go
```diff
@@ -278,6 +301,13 @@ func (g *generator) pagingCall(servName string, m *descriptor.MethodDescriptorPr
 
 func (g *generator) pagingIter(pt *iterType) {
 	p := g.printf
+	if pt.mapValueTypeName != "" {
+		p("// Holder type for string/%s map entries", pt.mapValueTypeName)
+		p("type %s struct {", pt.elemTypeName)
+		p("  Key string")
+		p("  Value %s", pt.mapValueTypeName)
```
noahdietz, line 308: This is not a sufficient type name. The protobuf-go types are in a different package i.e. `computepb.Instance`. You will want to store the `*descriptor.DescriptorProto` as `pt.mapValueType` and look up the `pbinfo.NameSpec`, using the `Name` it returns for the type and the `ImportSpec.Name`.

noahdietz, line 308: The value is going to be a pointer `*` too

internal/gengapic/paging.go
```diff
@@ -67,8 +57,33 @@ func (g *generator) iterTypeOf(elemField *descriptor.FieldDescriptorProto) (*ite
...
+			if valueField == nil {
+				return nil, errors.E(nil, "unusual map entry message: %q", eMsg)
+			}
+
+			if valueField.GetType() == descriptor.FieldDescriptorProto_TYPE_MESSAGE {
+				pt.mapValueTypeName = valueField.GetTypeName()
+			} else {
+				pt.mapValueTypeName = pbinfo.GoTypeForPrim[valueField.GetType()]
+			}
+			pt.elemTypeName = fmt.Sprintf("String%sPair", upperFirst(pt.mapValueTypeName))
```
noahdietz, line 81: I feel like encoding the types into the struct name is a little unnecessary. Maybe just `%Entry`? like `AcceleratorTypesScopedListEntry`. `Pair` is fine too, or `Item`. I just don't think the `String` is necessary.

internal/gengapic/genrest.go
```diff
@@ -603,8 +603,22 @@ func (g *generator) pagingRESTCall(servName string, m *descriptor.MethodDescript
 	p("")
 	p("  unm.Unmarshal(buf, resp)")
 	p("  it.Response = resp")
-	// TODO(dovs): handle map pages
-	p("  return resp.Get%s(), resp.GetNextPageToken(), nil", snakeToCamel(elemField.GetName()))
+	repeatedField, elts := fmt.Sprintf("resp.Get%s()", snakeToCamel(elemField.GetName())), ""
+	if pt.mapValueTypeName != "" {
+		elts = "elts"
```
vchudnov-g, line 608: nit: I know Go style advocates for short variable names, so this may be a losing fight, but I'd prefer just `elements` or even `elems`

internal/gengapic/genrest.go
```diff
@@ -603,8 +603,22 @@ func (g *generator) pagingRESTCall(servName string, m *descriptor.MethodDescript
...
+		p("")
+		p("    elts := make([]%s, 0, len(%s))", pt.elemTypeName, repeatedField)
```
vchudnov-g, line 610: This will work when the map's value type is a list (ie `repeated`)? That's the GCE use case: each page is a map of strings to list, and each item being iterated should be a (string, List) pair. (The code I have pending for Showcase to test mapped responses implements this. I can put out a draft PR so you can test against it.)

noahdietz: I think what this code is doing is actually building a slice of the the key-value pair structs from the map. Basically "flattening" the `map` into a list.

The value in the map will be a message that has a slice inside of it, if I'm not mistaken.

internal/gengapic/genrest.go
noahdietz: +1 `elems` would be better :)

internal/gengapic/genrest.go
software-dov: Noah is correct. This code is reconstructing the proto-implementation of having repeated key/value pairs.
If we have the following proto message:
```proto
message List {
    repeated string items = 1;
}

message Request {
    string next_page_token = 1;
    map<string, List> lists = 2;
}
```

Gives us the following helper structs:
```go
type ListPair struct {
    Key string
    Value *List
} 
```

The `elts` slice is then of type `[]ListPair`.

internal/gengapic/genrest.go
software-dov: s/elts/elems it is then.

internal/gengapic/paging.go
software-dov: I am leaning towards `Pair` more than `Entry` to make it more clear that this struct is subtly distinct from the MapEntry message generated by protoc since the types are coalesced. I'll remove `String`.

noahdietz: > the MapEntry message generated by protoc

That `MapEntry` message is an internal detail of the descriptors presented by protoc to a plugin. It doesn't make it into the generated Go code from protobuf-go, so there is nothing to differentiate it from in the generated output. But `Pair` is fine.

software-dov: >  It doesn't make it into the generated Go code from protobuf-go
Correct, and I wanted to make it clear that this is not that. I'll stay with `Pair`.

internal/gengapic/genrest.go
vchudnov-g: OK, what you describe is correct behavior. I just wanted to make sure it works when the map values are lists, but as you reminded me. `repeated`s need to be embedded in `message`s, so my concern is the same as ensuring that this works when the map values are messages instead of primitives. In your companion PR, the return proto map has primitive value type, so just wanting to make sure this works with messages....

vchudnov-g:
Here's an older fork PR version of my work with paginated maps. You can get the protos and Showcase server code from it: https://github.com/vchudnov-g/gapic-showcase/tree/regapic-pagination

internal/gengapic/genrest.go
software-dov: Noah's comment about full message type names prompted me to add a test where the value type is a message, which fails in an interesting and unexpected but not serious way. It is mostly fine.

internal/gengapic/paging.go
software-dov: This encouraged me to run a test with a message value type to make sure the pointer aspect was being preserved, and it highlighted a different problem.

software-dov: I went with manipulating the `mapValueType` during assingment.

internal/gengapic/paging.go
software-dov: Done

internal/gengapic/paging.go
software-dov: This one wound up gaining additional complexity because of message type package disambiguation. The result I think is more clear with explicit if and else instead of default and special case.

internal/gengapic/paging.go
```diff
@@ -73,16 +80,23 @@ func (g *generator) iterTypeOf(elemField *descriptor.FieldDescriptorProto) (*ite
...
+			// but check in case it's a primitive.
 			if valueField.GetType() == descriptor.FieldDescriptorProto_TYPE_MESSAGE {
-				pt.mapValueTypeName = valueField.GetTypeName()
+				vType := g.descInfo.Type[valueField.GetTypeName()]
+				imp, err := g.descInfo.ImportSpec(vType)
+				if err != nil {
+					return nil, err
+				}
+
+				elems := strings.Split(valueField.GetTypeName(), ".")
```
noahdietz, line 92: If you use `NameSpec` it will give the Go name for the type. I think that's what you're trying to do with all of the splitting?

software-dov: Correct. I'll make another pass and use `NameSpec`, since it's taking care of all the string shenanigans for me.

internal/gengapic/paging.go
```diff
@@ -246,7 +269,21 @@ func (g *generator) pagingCall(servName string, m *descriptor.MethodDescriptorPr
...
+		p("    elems := make([]%s, 0, len(%s))", pt.elemTypeName, repeatedField)
+		p("    for k, v := range %s {", repeatedField)
+		p("        elems = append(elems, %s{k, v})", pt.elemTypeName)
+		p("    }")
+		p("    sort.Slice(elems, func(i, j int) bool { return elems[i].Key < elems[j].Key } )")
+		p("")
+		g.imports[pbinfo.ImportSpec{Path: "sort"}] = true
+	} else {
+		elems = repeatedField
+	}
```
noahdietz, line 285: Nit: This seems like one that could be moved up as the default case.

Feel free to ignore for the sake of expediency if you'd prefer.

internal/gengapic/paging.go
```diff
@@ -278,6 +315,13 @@ func (g *generator) pagingCall(servName string, m *descriptor.MethodDescriptorPr
 
 func (g *generator) pagingIter(pt *iterType) {
 	p := g.printf
+	if pt.mapValueTypeName != "" {
+		p("// Holder type for string/%s map entries", pt.mapValueTypeName)
```
noahdietz, line 319: Comments for an exported element should begin with the name of that element e.g. `// FooPair is a holder type...`

internal/gengapic/genrest.go
```diff
@@ -603,8 +603,22 @@ func (g *generator) pagingRESTCall(servName string, m *descriptor.MethodDescript
...
+		p("    elems := make([]%s, 0, len(%s))", pt.elemTypeName, repeatedField)
+		p("    for k, v := range %s {", repeatedField)
+		p("        elems = append(elems, %s{k, v})", pt.elemTypeName)
+		p("    }")
+		p("    sort.Slice(elems, func(i, j int) bool { return elems[i].Key < elems[j].Key } )")
+		p("")
+		g.imports[pbinfo.ImportSpec{Path: "sort"}] = true
+	} else {
+		elems = repeatedField
+	}
```
noahdietz, line 619: Nit: This seems like one that could be moved up as the default case.

Feel free to ignore for the sake of expediency if you'd prefer.

internal/gengapic/paging.go
```diff
@@ -246,7 +269,21 @@ func (g *generator) pagingCall(servName string, m *descriptor.MethodDescriptorPr
 	p("  }")
 	p("")
 	p("  it.Response = resp")
-	p("  return resp.Get%s(), resp.GetNextPageToken(), nil", snakeToCamel(elemField.GetName()))
+	repeatedField, elems := fmt.Sprintf("resp.Get%s()", snakeToCamel(elemField.GetName())), ""
+	if pt.mapValueTypeName != "" {
```
noahdietz, line 273: Is this the exact same code as what's in `genrest` `paginRESTCall`? Why is it needed in both? Should it be a function then?

vchudnov-g: Along similar lines: paging.go is for gRPC-specific paging, or paging in general? And genrest.go is for REST in particular? It seems like some of the functionality/logic is common, so it could be factored out to prevent duplication.

software-dov:
New companion PR: https://github.com/googleapis/gapic-showcase/pull/817

noahdietz (approved):
Pending local showcase testing of mapped pagination, LGTM.

PR #722, opened 2021-07-28
gapic: wrap diregapic custom operations

noahdietz:
In `diregapic` mode, a new plugin flag, the generator will look for a type named `Operation` in the proto-package-to-generate, and wherever it is used as a return type for an RPC, it will be replaced by an `Operation` type generated in the gapic's package that wraps the proto-defined `Operation`.  This will allow us to implement long running operation helpers in the future by extending the generated `Operation` wrapper with new functionality. Without this, we'd most likely need to take a breaking change to wrap the proto-defined `Operation` and extend it with helpers.

Companion PR for `compute`: https://github.com/googleapis/google-cloud-go/pull/4511


noahdietz:
I need to add some baseline tests for the generated methods and type(s).

codyoss:
This looks good overall, I'll let your team do the formal review. What sort of helpers/extensions do you see us adding? I think that would help paint the picture of how this would be used.

noahdietz:
> What sort of helpers/extensions do you see us adding?

The working overall design doc is at go/actools-diregapic-lro, but similar APIs to what we have on normal GAPICs today - `Polling`, `Wait`, `Done`, etc. The configuration for this is still being hammered out.

noahdietz:
Update: I've added a baseline for `genRESTMethod` and this specific case. PTAL everyone.

vchudnov-g (commented):
looks good, but I want to ensure I understand why I don't see the custom operation in the companion pr.

internal/gengapic/custom_operation.go
```diff
@@ -0,0 +1,92 @@
...
+// customOpProtoName builds the fully-qualified proto name for the custom
+// operation message type.
+func (g *generator) customOpProtoName() string {
+	f := g.descInfo.ParentFile[g.aux.customOp.proto]
+	return fmt.Sprintf(".%s.%s", f.GetPackage(), g.aux.customOp.proto.GetName())
+}
+
+// customOpPointerTyp builds a string containing the Go code for a pointer to
+// the custom operation type.
+func (g *generator) customOpPointerTyp() (string, error) {
```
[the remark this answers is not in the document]
noahdietz: Done!

noahdietz:
> looks good, but I want to ensure I understand why I don't see the custom operation in the companion pr.

It is there :) I've replied to your comments.

vchudnov-g (approved):

internal/gengapic/custom_operation.go
```diff
@@ -0,0 +1,92 @@
...
+import (
+	"fmt"
+
+	"github.com/golang/protobuf/protoc-gen-go/descriptor"
+)
+
+// customOp represents a custom operation type for long running operations.
+type customOp struct {
+	proto     *descriptor.DescriptorProto
+	generated bool
```
software-dov, line 26: My one nit in the entire PR are about this field name. It may be clearer what it means to use `message`.

software-dov (approved):
LGTM, one field naming nit

internal/gengapic/custom_operation.go
noahdietz: Good call! Done.

PR #730, opened 2021-08-02
chore: add apidiff to showcase test

### LINKED ISSUE #593: chore: can we enable apidiff of the generated code as a presubmit?
While reviewing #577, verifying that there are no breaking changes to the generated surface must be done separately from the generator (install modified generator locally, clone & regenerate google-cloud-go, either manually run `apidiff` script or open a PR for CI to run it). This is cumbersome and requires that either the author does it and mentions that they have OR the reviewer must request it if they remember, adding "unforeseen" work for the PR author.

It would be really nice to be able to run `golang.org/x/exp/apidiff` against some or all of google-cloud-go in the presubmits to alleviate this concern/manual verification.

noahdietz:
Fixes #593


codyoss (approved):
LGTM

noahdietz:
Realized I need to add a break-glass option for this so we can actually make intentional breaking changes to Showcase.

PR #732, opened 2021-08-03
chore: add tests that verify codegen imports

### LINKED ISSUE #435: gapic: add dependency inclusion tests

noahdietz:
Fixes #435


codyoss (approved):
LGTM

PR #745, opened 2021-08-18
gapic: wrap errors from REST better

software-dov:
If the http client encounters an error when making a requst, it uses
the response to build a googleapi.Error instead in order to better
preserve information and aid readability.


showcase/echo_test.go
```diff
@@ -68,19 +69,25 @@ func TestEcho_error(t *testing.T) {
...
-		status, _ := status.FromError(err)
-		if status.Code() != val {
-			t.Errorf("%s Echo() errors with %d, want %d", typ, status.Code(), val)
+		if typ == "grpc" {
+			status, _ := status.FromError(err)
+			if status.Code() != val {
+				t.Errorf("%s Echo() errors with %d, want %d", typ, status.Code(), val)
+			}
+		} else {
+			const ERR_CODE int = 500 // Unavailable
```
noahdietz, line 82: Use `StatusInternalServerError` from https://pkg.go.dev/net/http instead please. also 503 is unavailable, 500 is internal server error.

internal/gengapic/genrest.go
```diff
@@ -845,8 +846,9 @@ func (g *generator) unaryRESTCall(servName string, m *descriptor.MethodDescripto
 	p("}")
 	p("defer httpRsp.Body.Close()")
 	p("")
-	p("if httpRsp.StatusCode != http.StatusOK {")
-	p("  return nil, fmt.Errorf(httpRsp.Status)")
+	p("err = googleapi.CheckResponse(httpRsp)")
+	p("if err != nil {")
```
noahdietz, line 850: Nit: Can you inline this please?

internal/gengapic/genrest.go
```diff
@@ -608,8 +608,9 @@ func (g *generator) pagingRESTCall(servName string, m *descriptor.MethodDescript
 	p("  }")
 	p("  defer httpRsp.Body.Close()")
 	p("")
-	p("  if httpRsp.StatusCode != http.StatusOK {")
-	p(`    return nil, "", fmt.Errorf(httpRsp.Status)`)
+	p("  err = googleapi.CheckResponse(httpRsp)")
+	p("  if err != nil {")
```
noahdietz, line 612: Nit: Can you inline this please?

showcase/echo_test.go
software-dov: That is exactly what I was looking for. Thanks!

internal/gengapic/genrest.go
software-dov: Done

internal/gengapic/genrest.go
software-dov: Done.

noahdietz (approved):

PR #746, opened 2021-08-19
gapic: fix fields with nums

noahdietz:
When a proto field has a number in a snake_case field name like `foo_123_bar`, protobuf-go generates the field name and resulting accessor as `Foo_123Bar` and `GetFoo_123Bar` respectively. We recently had an API add such a field, seemingly a first for us.

This change updates the `snakeToCamel` helper to account for that, and updates several places where it was being used directly to generate accessors to instead use the `fieldGetter` helper that uses `snakeToCamel` under the hood.

Fixes b/197160410.


codyoss (approved):
LGTM, thanks for this!

PR #754, opened 2021-08-29
gapic: exclude diregapic Wait RPC from op wrapping

noahdietz:
This excludes the `Wait` RPCs in DIREGAPIC APIs from having their return type be the custom operation wrapper. Instead, `Wait` RPCs will return the raw operation type. 

Note: This will be a breaking change for compute, but it is intended.

### DIFF
--- a/internal/gengapic/gengapic.go
+++ b/internal/gengapic/gengapic.go
@@ -481,9 +481,14 @@ func (g *generator) returnType(m *descriptor.MethodDescriptorProto) (string, err
 
 	// Regular return type.
 	retTyp := fmt.Sprintf("*%s.%s", outSpec.Name, outType.GetName())
+	wrap := g.opts.diregapic && // Generator in DIREGAPIC mode.
+		g.aux.customOp != nil && // API Defines a custom operation.
+		m.GetOutputType() == g.customOpProtoName() && // Method returns the custom operation.
+		info.verb != "get" && // Method is not a GET (polling methods).
+		m.GetName() != "Wait" // Method is not a Wait (uses POST).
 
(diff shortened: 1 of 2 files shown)


codyoss (approved):

internal/gengapic/gengapic.go
```diff
@@ -481,9 +481,14 @@ func (g *generator) returnType(m *descriptor.MethodDescriptorProto) (string, err
...
 	// Regular return type.
 	retTyp := fmt.Sprintf("*%s.%s", outSpec.Name, outType.GetName())
-
-	// Returning a custom operation, not a GET (which would be a polling RPC), use the wrapper type.
-	if g.opts.diregapic && g.aux.customOp != nil && m.GetOutputType() == g.customOpProtoName() && info.verb != "get" {
+	wrap := g.opts.diregapic && // Generator in DIREGAPIC mode.
+		g.aux.customOp != nil && // API Defines a custom operation.
+		m.GetOutputType() == g.customOpProtoName() && // Method returns the custom operation.
+		info.verb != "get" && // Method is not a GET (polling methods).
+		m.GetName() != "Wait" // Method is not a Wait (uses POST).
```
vchudnov-g, line 488: This looks right, but let's have @vam-google confirm. These are just heuristics that will be replaced by more robust logic once the annotations to the synthetic protos get committed, right?

noahdietz: That's a great question. @vam-google please confirm the heuristics I've documented here and Victor's question regarding their temporary nature.

vam-google (approved):
LGTM

Issue #747, opened 2021-08-22
Regression: Index out of range


valery-barysok:
#### Environment details

  - Programming language: golang
  - OS: mac os
  - Language runtime version: 1.17
  - Package version: github.com/googleapis/gapic-generator-go v0.23.0

#### Steps to reproduce

  Define a service without methods like:
  
  ```proto
  service MyService {
       option (google.api.default_host) = "my.service.com";
  }
  ```
 
The issue is located at https://github.com/googleapis/gapic-generator-go/blob/master/internal/gengapic/doc_file.go#L90
 
 ```
 panic: runtime error: index out of range [0] with length 0

goroutine 1 [running]:
github.com/googleapis/gapic-generator-go/internal/gengapic.(*generator).genDocFile(0xc000216d80, 0x1, {0x0, 0x0, 0xb}, 0xc0003745a0)
	/go/pkg/mod/github.com/googleapis/gapic-generator-go@v0.23.0/internal/gengapic/doc_file.go:90 +0x1bfd
github.com/googleapis/gapic-generator-go/internal/gengapic.Gen(0xc000265dc0)
	/go/pkg/mod/github.com/googleapis/gapic-generator-go@v0.23.0/internal/gengapic/gengapic.go:110 +0x916
main.main()
	/go/pkg/mod/github.com/googleapis/gapic-generator-go@v0.23.0/cmd/protoc-gen-go_gapic/main.go:37 +0xf7
--go_gapic_out: protoc-gen-go_gapic: Plugin failed with status code 2.
 ```

noahdietz:
@valery-barysok thanks for reporting this, please feel free to send a PR to fix this. Otherwise @alicejli can you take a look when you get a chance?

PR #756, opened 2021-08-30
fix: do not generate sample method snippet for services without methods

alicejli:
This should fix: https://github.com/googleapis/gapic-generator-go/issues/747.

For services that do not contain methods: 1) we should not generate a sample code snippet using a method for the documentation, and 2) we should not throw an error when end users are attempting generate a client library for such services.


internal/gengapic/doc_file.go
```diff
@@ -79,18 +79,25 @@ func (g *generator) genDocFile(year int, scopes []string, serv *descriptor.Servi
...
-	g.pt = printer.P{}
-	g.exampleMethodBody(g.opts.pkgName, pbinfo.ReduceServName(serv.GetName(), g.opts.pkgName), serv.GetMethod()[0])
-	snipMethod := g.pt.String()
-	g.pt = tmpMethod
-	g.codesnippet(snipMethod)
+
+	// If the service does not have any methods, then do not generate sample method snippet.
+	if serv.GetMethod() == nil || len(serv.GetMethod()[0].GetName()) == 0 {
+		p("//")
+	} else {
```
noahdietz, line 86:
```suggestion
	p("//")
	if len(serv.GetMethod()) > 0 {
```

The p("//") is a common line to both branches so move it outside of the condition, and invert the condition to only generate pkg level examples if the service has at least one RPC defined.

noahdietz:
The `compute-regen` job fails because of a recent change that introduces a breaking change in the compute client. It is OK, just disable it for this using the label (or use `breaking change allowed` label) and rerun the checks.

internal/gengapic/doc_file.go
alicejli: makes sense - I had to use `serv.GetMethod()[0].GetName()` as opposed to just `serve.GetMethod()` because if `[]*MethodDescriptorProto` existed but just had an empty Method object, then the length was still 1 as opposed to 0 (although I'm not sure if that's a valid concern actually - if a service doesn't have any methods defined, would `[]*MethodDescriptorProto` be nil?). 
I inverted the if statement; let me know how it looks.

noahdietz: > existed but just had an empty Method object, then the length was still 1 as opposed to 0 (although I'm not sure if that's a valid concern actually

It's not a concern :)

> if a service doesn't have any methods defined, would []*MethodDescriptorProto be nil?

Yep! And in Go, using `len` or `range` on a `nil` slice produces `0` and nothing respectively.

internal/gengapic/doc_file.go
```diff
@@ -79,12 +79,9 @@ func (g *generator) genDocFile(year int, scopes []string, serv *descriptor.Servi
...
 	p("// The methods of Client are safe for concurrent use by multiple goroutines.")
 	p("// The returned client must be Closed when it is done being used.")
-
+	p("//")
 	// If the service does not have any methods, then do not generate sample method snippet.
-	if serv.GetMethod() == nil || len(serv.GetMethod()[0].GetName()) == 0 {
-		p("//")
-	} else {
-		p("//")
+	if len(serv.GetMethod()[0].GetName()) > 0 {
```
noahdietz, line 84: If a service has no methods it will be a `nil` map and this produces an index out of bounds `panic`. We just need to check `len(serv.GetMethod()) > 0`. If a method is present, it will always have a `Name`.

internal/gengapic/doc_file.go
alicejli: ahhh makes sense; okay I've updated the if statement as well the test. Thanks for clarifying!

noahdietz (approved):

PR #764, opened 2021-08-31
gapic: rest GAPIC wrap unknown enum error

noahdietz:
This adds generation and use of a `maybeUnknownEnum` helper to REST clients. This helper will `Wrap` proto-JSON parsing errors caused by the server replying with an unknown enum value. The goal is to produce a more friendly/actionable error message. See http://go/actools-regapic-unknown-enum for more details. The wrapped error produces an error message like the following:

    received an unknown enum value; a later version of the library may support it: proto: (line 1:17): invalid value for enum type: "NOT_VALID_ENUM"

This is (hopefully) a temporary solution, with the long-term solution being a fix to the proto-JSON parsers. 

A companion PR for `compute` is at https://github.com/googleapis/google-cloud-go/pull/4710.

Note: This change introduces the `golang.org/x/xerrors` module as a dependency.

### DIFF
--- a/internal/gengapic/doc_file.go
+++ b/internal/gengapic/doc_file.go
@@ -33,6 +33,7 @@ import (
 // it does not use g.commit().
 func (g *generator) genDocFile(year int, scopes []string, serv *descriptor.ServiceDescriptorProto) {
 	p := g.printf
+	hasREST := containsTransport(g.opts.transports, rest)
 
 	p(license.Apache, year)
 	p("")
@@ -116,6 +117,9 @@ func (g *generator) genDocFile(year int, scopes []string, serv *descriptor.Servi
 	p("%s%q", "\t", "strings")
 	p("%s%q", "\t", "unicode")
 	p("")
+	if hasREST {
+		p("%s%q", "\t", "golang.org/x/xerrors")
+	}
 	p("%s%q", "\t", "google.golang.org/api/option")
 	p("%s%q", "\t", "google.golang.org/grpc/metadata")
 	p(")")
@@ -202,6 +206,17 @@ func (g *generator) genDocFile(year int, scopes []string, serv *descriptor.Servi
 		p("  return %q", "UNKNOWN")
 		p("}")
 		p("")
+	}
+
+	if hasREST {
+		p("// maybeUnknownEnum wraps the given proto-JSON parsing error if it is the result")
+		p("// of receiving an unknown enum value.")
+		p("func maybeUnknownEnum(err error) error {")
+		p(`  if strings.Contains(err.Error(), "invalid value for enum type") {`)
+		p(`    err = xerrors.Errorf("received an unknown enum value, a later version of the library may support it: %%w", err)`)
+		p("  }")
+		p("  return err")
+		p("}")
 	}
 }
 
--- a/internal/gengapic/doc_file.go
+++ b/internal/gengapic/doc_file.go
@@ -213,7 +213,7 @@ func (g *generator) genDocFile(year int, scopes []string, serv *descriptor.Servi
 		p("// of receiving an unknown enum value.")
 		p("func maybeUnknownEnum(err error) error {")
 		p(`  if strings.Contains(err.Error(), "invalid value for enum type") {`)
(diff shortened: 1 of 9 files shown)


noahdietz:
Worth noting:
- Go error messages should not start with a capital letter and should not end with punctuation
- I used `xerrors.ErrorF` with `%w` so that we could do idiomatic Go error wrapping, rather than replacing the error with out own

internal/gengapic/doc_file.go
```diff
@@ -203,6 +207,17 @@ func (g *generator) genDocFile(year int, scopes []string, serv *descriptor.Servi
 		p("}")
 		p("")
 	}
+
+	if hasREST {
+		p("// maybeUnknownEnum wraps the given proto-JSON parsing error if it is the result")
+		p("// of receiving an unknown enum value.")
+		p("func maybeUnknownEnum(err error) error {")
+		p(`  if strings.Contains(err.Error(), "invalid value for enum type") {`)
+		p(`    err = xerrors.Errorf("received an unknown enum value, a later version of the library may support it: %%w", err)`)
```
vchudnov-g, line 216: nit: s/`,`/`;`/

vchudnov-g: I'm not sure about reporting the line number if the JSON response is not available to the caller. Maybe we can omit it somehow? I did not know about xerrors, but it says "The returned error includes the file and line number of the caller when formatted *with additional detail enabled*" (emphasis added). It's not clear if that enablement is the `%w` specifically. I'm not quite grokking it yet.

WDYT? Regardless, probably not a blocker, though it would be nice to remove the line number.

vchudnov-g (approved):
Not sure about the line number (see comment). If we can fix that, great, but otherwise not a blocker.

internal/gengapic/doc_file.go
noahdietz: > nit: s/,/;/

Ack.

The message that comes after the `:` is the raw error message from the parser error and `%w` is the formatter that indicates the given error should be wrapped so that the original error can be `Unwrap`'d and retained. The `%w` in this case isn't adding any extra information, what is printed is just the output of `Error()` from the parser error. So I'd prefer not to mangle the original error by stripping parts of the message.

noahdietz:
@codyoss can you give this a once over? It generates new code for compute so I want you to approve it.

codyoss:
LGTM, it almost seems like a thing that should live more so in a gax layer instead of generated in the client. But since this is for rest gapic I don't have strong opinions.

internal/gengapic/doc_file.go
vchudnov-g: Ah, I see. In that case, maybe we should wrap the error but format it differently? "If the format specifier includes a %w verb with an error operand in a position other than at the end, the returned error will still implement an Unwrap method returning the operand, but the error's Format method will not return the wrapped error." (https://pkg.go.dev/golang.org/x/xerrors#Errorf). IIUC, this would give us the error message  currently in the PR description up to the first `:`. WDYT?

(But no need to rat-hole on this)

noahdietz: > IIUC, this would give us the error message currently in the PR description up to the first :. WDYT

Your understanding is correct, but then we wouldn't get the unknown enum value, which is part of the original error message. Is that an appropriate trade off? I'd rather not parse it out because we should just report the original wrapped error message (as we are now) instead.

noahdietz:
> LGTM, it almost seems like a thing that should live more so in a gax layer instead of generated in the client. But since this is for rest gapic I don't have strong opinions.

Yeah we could go that route, but since its specific to DIREGAPIC, I didn't want to get too in the weeds in gax, and hopefully this is removed if we get a fix to protojson.

internal/gengapic/doc_file.go
vchudnov-g: Oh, shoot, you're right. Displaying the unknown enum is probably more important than omitting the line number. Parsing the error is somewhat brittle, in case the text changes...and not worth doing now given the likelihood we'll adopt another solution.

OK, nm.

PR #810, opened 2021-11-10
gapic: ignore mixin proto files in input when generating mixins

noahdietz:
We need to exclude the mixin proto files when they are present in the input from client generation, so that we do not generate extra clients.


noahdietz:
Disabling the compute regen job because it is broken. Since the protos were added to googleapis, we need to fix the generation script to look there instead of googleapis-discovery.

alicejli (approved):

PR #821, opened 2021-12-01
gapic: REGAPIC fix ignored empty value required fields in query

### LINKED ISSUE #819: gapic: non-optional number fields are excluded from REST query params if value is 0
Currently, REST transport implementations check if a number field (e.g. `int32`) is `0` or not, and will only include the field's value in request params if it is **not** `0`. [Example](https://github.com/googleapis/google-cloud-go/blob/f58a9f7b88e2ce6101cb4bd3c85c267a688a1a1d/compute/apiv1/instance_group_managers_client.go#L1301).

This could be a problem since `0` is both a valid value in many cases _and_ the default empty/unset value for type `int32` (which is why it is being checked in the first place).

Specifically, in DIREGAPIC compute, the only non-optional number fields are also annotated `REQUIRED`, and in all cases `0` is a valid value. This results in BadRequest errors because a field that is required is not sent when it is `0`, which is a valid value. See https://github.com/googleapis/google-cloud-go/issues/5174.

noahdietz:
A required, singular, primitive type field will always be added as a query param when configured as one, whether the default empty value of the primitive is used or not. 

Note: Also added some constants for proto field types to improve readability.

Fixes #819


codyoss (approved):
LGTM, thanks for the fast fix!

PR #824, opened 2021-12-03
feat(gengapic): support complex URIs in REGAPIC

timburks:
This extends generated REST clients to handle a few more types of patterns in URL paths. 

Verified on the [github.com/apigee/registry](https://github.com/apigee/registry) API. Examples of newly-supported paths include 
[`"/v1/{parent=projects/*/locations/*}/apis"`](https://github.com/apigee/registry/blob/5d8e97519c22d3582e73e632531096da41624299/google/cloud/apigeeregistry/v1/registry_service.proto#L56) and [`"/v1/{api.name=projects/*/locations/*/apis/*}"`](https://github.com/apigee/registry/blob/5d8e97519c22d3582e73e632531096da41624299/google/cloud/apigeeregistry/v1/registry_service.proto#L65).

PR #829, opened 2021-12-07
Add HTTP client support for HttpBody response values

timburks:
Follows #824. This adds explicit handling of [`HttpBody`](https://github.com/googleapis/googleapis/blob/9b9da22ee21750793c110aa98eeb63034c1484ca/google/api/httpbody.proto#L71) responses to generated unary HTTP clients. The `HttpBody` message is handled specially by transcoders, which return its `Data` field as the response body and its `ContentType` in an HTTP `Content-Type` header.


timburks:
@noahdietz I'll replace this with a new PR against main when #846 is in.

timburks:
Replaced with #847

PR #847, opened 2021-12-16
feat(gengapic): Add HTTP client support for HttpBody response values

timburks:
Follows #846. This adds explicit handling of HttpBody responses to generated unary HTTP clients. The HttpBody message is handled specially by transcoders, which return its Data field as the response body and its ContentType in an HTTP Content-Type header.


internal/gengapic/genrest.go
```diff
@@ -838,12 +838,19 @@ func (g *generator) unaryRESTCall(servName string, m *descriptor.MethodDescripto
...
-	p("unm := protojson.UnmarshalOptions{AllowPartial: true, DiscardUnknown: true}")
-	p("rsp := &%s.%s{}", outSpec.Name, outType.GetName())
-	p("")
-	p("if err := unm.Unmarshal(buf, rsp); err != nil {")
-	p("  return nil, maybeUnknownEnum(err)")
-	p("}")
+	if outSpec.Name == "httpbodypb" && outType.GetName() == "HttpBody" {
+		p("rsp := &httpbodypb.HttpBody{")
+		p("  Data: buf,")
+		p("  ContentType: httpRsp.Header[\"Content-Type\"][0],")
```
timburks, line 844: This is brittle - clients will crash if there is no Content-Type header.

internal/gengapic/genrest.go
```diff
@@ -838,12 +838,19 @@ func (g *generator) unaryRESTCall(servName string, m *descriptor.MethodDescripto
 	p("  return nil, err")
 	p("}")
 	p("")
-	p("unm := protojson.UnmarshalOptions{AllowPartial: true, DiscardUnknown: true}")
-	p("rsp := &%s.%s{}", outSpec.Name, outType.GetName())
-	p("")
-	p("if err := unm.Unmarshal(buf, rsp); err != nil {")
-	p("  return nil, maybeUnknownEnum(err)")
-	p("}")
+	if outSpec.Name == "httpbodypb" && outType.GetName() == "HttpBody" {
```
noahdietz, line 841: Can we instead compare the `outType`'s fully-qualified proto name to that of the `HttpBody`? `google.api.HttpBody`

You can build the FQN of the `outType` by retrieving the parent file and combining its package with the message name. `fqn := fmt.Sprintf("%s.%s", g.descInfo.ParentFile[outType].GetPackage(), outType.GetName())`. I realize this pattern doesn't work for nested messages, but this message isn't nested so it doesn't matter.

internal/gengapic/genrest.go
noahdietz: Yeah, let's make this more robust, see my suggestion.

internal/gengapic/genrest.go
```diff
@@ -838,12 +838,19 @@ func (g *generator) unaryRESTCall(servName string, m *descriptor.MethodDescripto
...
+		p("rsp := &httpbodypb.HttpBody{")
+		p("  Data: buf,")
+		p("  ContentType: httpRsp.Header[\"Content-Type\"][0],")
+		p("}")
```
noahdietz, line 845:
```suggestion
		p("}")
		p(`if headers := httpRsp.Header; len(headers["Content-Type"]) > 0 {`)
		p(`  rsp.ContentType = headers["Content-Type"][0]`)
		p("}")
```

Notice the use of back ticks instead of double-quotes so that we can use double-quotes unescaped in the string.

noahdietz: Considering there could be more than one value for `Content-Type`, should we join on `,` and return "all values" of the header? IDK if comma-separator is how HTTP delineates multiple values, or if its `;`, but we should do that if at all.

noahdietz (changes requested):

internal/gengapic/genrest.go
timburks: I'm not sure that multiple Content-Type headers are actually possible - it might just be an artifact of the implementation. https://stackoverflow.com/questions/5809099/does-the-http-protocol-support-multiple-content-types-in-response-headers

noahdietz: Cool thanks for looking into it.

internal/gengapic/genrest.go
timburks: Done, thanks!

internal/gengapic/genrest.go
timburks: Done.

noahdietz (approved):

PR #856, opened 2021-12-21
feat(gengapic): REGAPIC requests use gax.Invoke

### LINKED ISSUE #842: gapic: REGAPIC calls must be invoked via gax
Today, REGAPIC API calls are not wrapped in `gax.Invoke`, which means they are not able to support use of [`gax.Retryer`](https://pkg.go.dev/github.com/googleapis/gax-go/v2#Retryer) and any errors returned are not wrapped in [`apierror.APIError`](https://pkg.go.dev/github.com/googleapis/gax-go/v2/apierror#APIError).

noahdietz:
This wraps all REGAPIC API calls in `gax.Invoke`. As a result all instances of `googleapi.Error` are wrapped in `apierror.APIError`, which matches gRPC-based API calls. Using `gax.Invoke` enables configuration of default and call-time `gax.Retryer`.

The `gax.APICall` supplied includes the `http.NewRequest` call. This is intentional because it can be unsafe to reuse the same `http.Request` instance. Thus each attempt creates a new `http.Request` to send.

Fixes #842


codyoss (approved):
LGTM

PR #857, opened 2021-12-21
feat(gengapic): REGAPIC headers built from metadata

### LINKED ISSUE #855: gapic: REGAPIC header construction should use a pkg-level helper
All REGAPIC calls have the following code:

```go
// Set the headers
for k, v := range c.xGoogMetadata {
	httpReq.Header[k] = v
}
httpReq.Header["Content-Type"] = []string{"application/json",}
```

This can and should be extracted into a common method like `func insertHeaders(request *http.Request, md metadata.MD)`. Furthermore, the heads by default should include `["Content-Type"] = []string{"application/json"}` so that not every method must specify this explicitly.

noahdietz:
This adds a `buildHeaders` helper to REGAPIC clients that takes `metadata.MD` from the outgoing `context.Context`, joins it with other `metadata.MD` (like `xGoogMetadata` and the `Content-Type` header), and converts it into an `http.Header` for use on each `http.Request`. This extends the existing code by pulling `metadata.MD` from the outgoing `context.Context` which allows end-users to continue to set metadata/headers the way our GAPIC/gRPC already supports.

The client method builds the `http.Header` once and reuses it for each new `http.Request` built per attempt. This saves a small bit of processing time.

Fixes #855

### DIFF
--- a/internal/gengapic/doc_file.go
+++ b/internal/gengapic/doc_file.go
@@ -110,6 +110,9 @@ func (g *generator) genDocFile(year int, scopes []string, serv *descriptor.Servi
 
 	p("import (")
 	p("%s%q", "\t", "context")
+	if hasREST {
+		p("%s%q", "\t", "net/http")
+	}
 	p("%s%q", "\t", "os")
 	p("%s%q", "\t", "runtime")
 	p("%s%q", "\t", "strconv")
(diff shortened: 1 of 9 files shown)


codyoss (approved):
LGTM

PR #865, opened 2022-01-07
chore(gengapic): add fqn helper

noahdietz:
Adds a helper that builds the fully-qualified proto element name for a given element, changes some places where it can be used, and removes an extraneous helper `getServiceName` replaced by `fqn` calls. This helper aids me in other code I'm writing.


codyoss (approved):
LGTM

PR #866, opened 2022-01-11
feat(gengapic): implement diregapic lro foundation + polling

noahdietz:
This is the foundation of Full LRO support for DIREGAPIC/Compute:
* operation service initialization
* operation service-specific operation wrappers
* Poll, Done, Name methods

Missing APIs compared to Standard LRO:
* Cancel (all Compute operation services expose a Delete rpc, but nothing exactly like Cancel)
* Wait (only 3/4 Compute operation services expose a Wait rpc, need to follow up here)
* Metadata (No plans to support)
* Response (No plans to support)

Specifically, the Operation wrapper(s) generated for DIREGAPIC APIs will create a single Operation wrapper type, that embeds an (internal) operationHandle interface consisting of `Proto` and `Poll` methods. Every operation service will get an operationHandle implementation, which will be embedded in Operation wrappers created by RPCs that create Operations. Each method knows which operation service handle they need to create. The operation service handle is instantiated with the consumer API client and used for every Operation wrapper created by that client.

Companion PR for compute is: https://github.com/googleapis/google-cloud-go/pull/5346
Specifically, the [`operations.go`](https://github.com/googleapis/google-cloud-go/pull/5346/files#diff-f4a0c12e2a5a0ec101424679218e0c5270a8d0b61e697b4cb3edaedaf835d790) file contains the new operation handles.

### DIFF
--- a/internal/gengapic/genrest.go
+++ b/internal/gengapic/genrest.go
@@ -879,7 +879,7 @@ func (g *generator) unaryRESTCall(servName string, m *descriptor.MethodDescripto
 	p("}")
 	ret := "return resp, nil"
 	if isCustomOp {
-		s := g.getCustomOpService(m)
+		s := g.customOpService(m)
 		handleName := handleName(s.GetName(), g.opts.pkgName)
 		p("op := %s", g.customOpInit(handleName, "resp"))
 		ret = "return op, nil"
--- a/internal/gengapic/genrest.go
+++ b/internal/gengapic/genrest.go
@@ -134,12 +134,14 @@ func (g *generator) restClientUtilities(serv *descriptor.ServiceDescriptorProto,
 	p("")
 	if hasCustomOp {
 		opServName := pbinfo.ReduceServName(opServ.GetName(), g.opts.pkgName)
-		p("opC, err := New%sRESTClient(ctx, opts...)", opServName)
+		p("o := append(opts, option.WithHTTPClient(httpClient))")
+		p("opC, err := New%sRESTClient(ctx, o...)", opServName)
 		p("if err != nil {")
 		p("  return nil, err")
 		p("}")
 		p("c.operationClient = opC")
 		p("")
+		g.imports[pbinfo.ImportSpec{Path: "google.golang.org/api/option"}] = true
 	}
 	// TODO(dovs): make rest default call options
 	// TODO(dovs): set the LRO client
--- a/internal/gengapic/genrest.go
+++ b/internal/gengapic/genrest.go
@@ -134,8 +134,7 @@ func (g *generator) restClientUtilities(serv *descriptor.ServiceDescriptorProto,
 	p("")
 	if hasCustomOp {
 		opServName := pbinfo.ReduceServName(opServ.GetName(), g.opts.pkgName)
-		p("o := append(opts, option.WithHTTPClient(httpClient))")
-		p("opC, err := New%sRESTClient(ctx, o...)", opServName)
+		p("opC, err := New%sRESTClient(ctx, option.WithHTTPClient(httpClient))", opServName)
 		p("if err != nil {")
 		p("  return nil, err")
 		p("}")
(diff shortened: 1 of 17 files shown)


internal/gengapic/testdata/custom_op_init.want
```diff
@@ -74,6 +77,12 @@ func NewRESTClient(ctx context.Context, opts ...option.ClientOption) (*Client, e
 	}
 	c.setGoogleClientInfo()
 
+	opC, err := NewFooOperationRESTClient(ctx, opts...)
```
codyoss, line 80: I wonder if we should be passing the httpClient to this method so they share the same base TokenSource under the hood. Thoughts?

codyoss (approved):
LGTM, overall but one question.

internal/gengapic/testdata/custom_op_init.want
noahdietz: Great idea! I will do that.

noahdietz: Done!

internal/gengapic/genrest.go
```diff
@@ -134,12 +134,14 @@ func (g *generator) restClientUtilities(serv *descriptor.ServiceDescriptorProto,
 	p("")
 	if hasCustomOp {
 		opServName := pbinfo.ReduceServName(opServ.GetName(), g.opts.pkgName)
-		p("opC, err := New%sRESTClient(ctx, opts...)", opServName)
+		p("o := append(opts, option.WithHTTPClient(httpClient))")
```
codyoss, line 137: I think what we want to do here is just explicitly pass the httpClient and endpoint and not other user options. This will avoid some validation errors that users might otherwise run into: https://github.com/googleapis/google-api-go-client/blob/33250c025eceebe6d5ed6dae1e0689344fb7ce33/internal/settings.go#L126-L134. WDTY?

codyoss (commented):
One more small comment

internal/gengapic/genrest.go
noahdietz: Ah I see what you're saying. Yeah lets do that.

noahdietz: Done.

internal/gengapic/testdata/custom_op_init.want
```diff
@@ -77,8 +77,7 @@ func NewRESTClient(ctx context.Context, opts ...option.ClientOption) (*Client, e
 	}
 	c.setGoogleClientInfo()
 
-	o := append(opts, option.WithHTTPClient(httpClient))
-	opC, err := NewFooOperationRESTClient(ctx, o...)
+	opC, err := NewFooOperationRESTClient(ctx, option.WithHTTPClient(httpClient))
```
codyoss, line 80: I forgot, do operations operate off the same endpoint as a mixin or are they global. If they are based off the endpoint the endpoint returned from creating the client should also be passed here.

noahdietz: As in via `option.WithEndpoint`?

The operation client(s) for DIREGAPIC isn't standard LRO, but rather another API on the same endpoint defined in the same proto package. Thus, those operation clients by default target the same endpoint as the non-operation clients (those that precipitate the creation of operations). 

So both the containing client and the operation client refer to `compute.googleapis.com` and should refer to the same endpoint if changed at runtime. Hence we should also provide `WithEndpoint` using the endpoint from `httpClient` creation?

noahdietz: And MTLS endpoint too?

noahdietz: maybe not...

noahdietz: Providing WithEndpoint too. Done.

codyoss (approved):
Ship it!

noahdietz:
@frodothetrue can you please give this new generated surface (and the companion PR) a look when you get a chance?

FrodoTheTrue (approved):
LGTM

PR #873, opened 2022-01-26
feat(gengapic): diregapic add operation Wait helper

noahdietz:
Adds the Operation helper `Wait` to DIREGAPIC operation wrappers. This uses `Poll` to check the status of the operation until it is done or the context is canceled. It uses exponential backoff until that backoff reaches 1 min, when it will then poll regularly every minute. This is the same for standard GAPICs using the LRO helper. 

cc: @vchudnov-g @vam-google FYI


codyoss (approved):
LGTM

internal/gengapic/custom_operation.go
```diff
@@ -130,6 +130,28 @@ func (g *generator) customOperationType() error {
 	p("}")
 	p("")
 
+	// Wait
+	p("// Wait blocks until the operation is complete, polling regularly")
+	p("// after an intial period of backing off between attempts.")
+	p("func (o *%s) Wait(ctx context.Context, opts ...gax.CallOption) error {", opName)
+	p("	 bo := gax.Backoff{")
+	p("    Initial: 1 * time.Second,")
+	p("    Max:     time.Minute,")
```
vchudnov-g, line 139: nit: why not `1 * time.Minute` here like in the line above for `Second`?

internal/gengapic/custom_operation.go
```diff
@@ -130,6 +130,28 @@ func (g *generator) customOperationType() error {
...
+	p("	 }")
+	p("	 for {")
```
vchudnov-g, line 141: nit: indentation within the quotes is wonky (I know it doesn't affect the final product, but for readability here)

internal/gengapic/custom_operation.go
```diff
@@ -130,6 +130,28 @@ func (g *generator) customOperationType() error {
 	p("}")
 	p("")
 
+	// Wait
+	p("// Wait blocks until the operation is complete, polling regularly")
+	p("// after an intial period of backing off between attempts.")
+	p("func (o *%s) Wait(ctx context.Context, opts ...gax.CallOption) error {", opName)
+	p("	 bo := gax.Backoff{")
+	p("    Initial: 1 * time.Second,")
```
vchudnov-g, line 138: Are these numbers the same as for gRPC? It would probably make sense to define them centrally as constants rather than hard-coding them as literals into the templates. (though I'm fine with the generated code having literals)

vchudnov-g (approved):
One comment about not hard coding literals in the generator, and a couple of really trivial nits. Looks good. Thanks for doing this!

internal/gengapic/custom_operation.go
noahdietz: Yes they are the same and as gRPC. I'll put them in constants for the generator.

internal/gengapic/custom_operation.go
noahdietz: Ah thanks for catching this, copy-paste resulted in some tab-characters sneaking in. Fixed.

internal/gengapic/custom_operation.go
noahdietz: Good catch. Done.

internal/gengapic/custom_operation.go
noahdietz: I switched them both to drop the `1 *`, it is extraneous.

PR #876, opened 2022-01-28
feat(gengapic): diregapic lro polling request params

noahdietz:


internal/gengapic/custom_operation.go
```diff
@@ -66,12 +68,42 @@ func (g *generator) customOpPointerType() (string, error) {
 // customOpInit builds a string containing the Go code for initializing the
 // operation wrapper type with the Go identifier for a variable that is the
 // proto-defined operation type.
-func (g *generator) customOpInit(h, p string) string {
+func (g *generator) customOpInit(p, r, o string, req *descriptor.DescriptorProto, s *descriptor.ServiceDescriptorProto) {
```
codyoss, line 71: nit: given p,r,o are all just strings it may be beneficial to give them better names for readability. Hard to know what they are used for without a stronger type name.

internal/gengapic/custom_operation.go
```diff
@@ -66,12 +68,42 @@ func (g *generator) customOpPointerType() (string, error) {
...
+	h := handleName(s.GetName(), g.opts.pkgName)
 	opName := g.aux.customOp.message.GetName()
+	pt := func(f string, v ...interface{}) {
```
codyoss, line 74: seems like other spots in the file we just use p. Also `pt := g.printf` if you want an alias.

internal/gengapic/custom_operation.go
```diff
@@ -298,6 +338,57 @@ func operationResponseField(m *descriptor.DescriptorProto, target string) *descr
...
+func (g *generator) pollingRequestParameters(m *descriptor.MethodDescriptorProto, opServ *descriptor.ServiceDescriptorProto) []string {
+	poll := operationPollingMethod(opServ)
+	if poll == nil {
+		return nil
+	}
+	pollReqName := poll.GetInputType()
+
+	inType := g.descInfo.Type[m.GetInputType()].(*descriptor.DescriptorProto)
+
+	params := []string{}
```
codyoss, line 375: nit: `var params []string` maybe useful enough as nil slice is meaningful. If this works could move it to the top and always return the slice?

codyoss (commented):
Just a couple small things.

internal/gengapic/custom_operation.go
noahdietz: SGTM! Done.

internal/gengapic/custom_operation.go
noahdietz: Yes, good call, I had something else before I started using the printer and didn't clean this up properly. Thanks.

internal/gengapic/custom_operation.go
noahdietz: Good point, thanks for keeping me from being lazy :) Done.

codyoss (approved):
LGTM

PR #893, opened 2022-02-11
feat: set version equal to module version

codyoss:
As of https://github.com/googleapis/google-cloud-go/pull/5517 we
now have a version file in the internal package for every module.
This change reads variable and uses it for the client version. This
means all autogenerated clients will now get a meaningful realease
version in thier call metadata.

### DIFF
--- a/internal/gengapic/doc_file.go
+++ b/internal/gengapic/doc_file.go
@@ -15,6 +15,7 @@
 package gengapic
 
 import (
+	"fmt"
 	"sort"
 	"strings"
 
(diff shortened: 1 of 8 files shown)


codyoss:
@noahdietz This is a little hacky but something roughly like this is what I was thinking... LMK WYT

noahdietz:
I want to make this happen because it will be useful. My main concern is that a Go gapic is no longer compilable without the addition of the `internal.Version` variable. Let's figure out how we can overcome that. Perhaps with a flag of some sort that we only use in our regen pipeline? Or does the generator also generate that file with `var Version = 0.1.0`, and our pipeline only copies the generated file on the first generation of the pkg?

PR #912, opened 2022-02-23
feat: make versionClient a var

codyoss:
This will allow us to override the value after the client has been
generated in an init function. For that code please see
https://github.com/googleapis/google-cloud-go/pull/5687

Previous "UNKNOWN" version behaviour in maintained in a new
function.

### DIFF
--- a/internal/gengapic/doc_file.go
+++ b/internal/gengapic/doc_file.go
@@ -134,7 +134,7 @@ func (g *generator) genDocFile(year int, scopes []string, serv *descriptor.Servi
 	p("var versionClient string")
 	p("")
 	p("func getVersionClient() string {")
-	p("  if versionClient == \"\" {")
+	p(`  if versionClient == "" {`)
 	p("    return %q", "UNKNOWN")
 	p("  }")
 	p("  return versionClient")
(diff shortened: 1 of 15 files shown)


internal/gengapic/doc_file.go
```diff
@@ -131,7 +131,14 @@ func (g *generator) genDocFile(year int, scopes []string, serv *descriptor.Servi
 	p("type clientHook func(context.Context, clientHookParams) ([]option.ClientOption, error)")
 	p("")
 
-	p("const versionClient = %q", "UNKNOWN")
+	p("var versionClient string")
+	p("")
+	p("func getVersionClient() string {")
+	p("  if versionClient == \"\" {")
```
noahdietz, line 137:
```suggestion
	p(`  if versionClient == "" {`)
```

internal/gengapic/doc_file.go
```diff
@@ -131,7 +131,14 @@ func (g *generator) genDocFile(year int, scopes []string, serv *descriptor.Servi
...
+	p("    return %q", "UNKNOWN")
```
noahdietz, line 138:
```suggestion
	p(`    return "UNKNOWN"`)
```

noahdietz (approved):
LGTM with two nits

PR #933, opened 2022-03-16
feat(gengapic): regapic server-streaming

noahdietz:
This adds support for REGAPIC Server Stream RPCs. Since server-stream RPCs in GAPICs must return the gRPC-defined, RPC-specific "stream client" interface, we must implement that interface around the HTTP-JSON stream (the Response Body) opened by the HTTP-JSON request. An RPC-specific implementation of that stream client interface is generated in the GAPIC that utilizes the `gax.ProtoJSONStream` helper to decode the stream contents into the proto messages. The behavior of this generated stream client is the same as documented by the gRPC-defined interface.

When the stream is done/closed, `io.EOF` is returned.

If the `context.Context` used to initialize the stream is canceled, the invocations of `Recv` fail.

When the stream is finished or in the event of an error, the `gax.ProtoJSONStream` is closed by the generated stream client.

### DIFF
--- a/internal/gengapic/testdata/rest_ServerStreamRPC.want
+++ b/internal/gengapic/testdata/rest_ServerStreamRPC.want
@@ -35,13 +35,12 @@ func (c *fooRESTClient) ServerStreamRPC(ctx context.Context, req *foopb.Foo, opt
 		}
 		return nil
 	}, opts...)
-	if e != nil {
-		return nil, e
-	}
 
-	return streamClient, nil
+	return streamClient, e
 }
 
+// serverStreamRPCRESTClient is the stream client used to consume the server stream created by
+// the REST implementation of ServerStreamRPC.
 type serverStreamRPCRESTClient struct {
 	ctx context.Context
 	md metadata.MD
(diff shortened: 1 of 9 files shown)


internal/gengapic/testdata/rest_ServerStreamRPC.want
```diff
@@ -0,0 +1,90 @@
...
+func (c *serverStreamRPCRESTClient) CloseSend() error {
+	// This is a no-op to fulfill the interface.
+	return nil
+}
+
+func (c *serverStreamRPCRESTClient) Context() context.Context {
+	return c.ctx
+}
+
+func (c *serverStreamRPCRESTClient) SendMsg(m interface{}) error {
```
codyoss, line 81: I wonder if these "noop" impls should return errors when they are not supported, thoughts?

noahdietz: Yeah, I think that would be fine! Is it worth making a single "ErrUnimplemented" error? WDYT is best?

codyoss: I am not sure we need a sentinel, I don't think it provides too much value at runtime to know it was an unimplmented error as this will only be run into at dev time. Maybe just return a grpc.Status with a code unimiplmented?

noahdietz: Done!

noahdietz:
I regenerated the Firestore client with these changes and with REST transport generation enabled. I was then able to successfully consume a RunQuery REST stream.

PR #942, opened 2022-03-24
feat(gengapic): regapic lro support

### LINKED ISSUE #845: gapic: REGAPIC support for standard LRO wrappers
The longrunning Operation wrappers generated for gRPC must work the same for REGAPIC clients, which means the REGAPIC implementation of the LRO client must be used.

noahdietz:
Adds support for LRO Wrappers to REGAPIC. This is done by supplying a `gax.WithPath` call option to the wrapper that is used on every `Poll` call. The option contains the HTTP Path override specific to the GetOperation HTTP Rule override for the API. See http://go/actools-regapic-lro-go for more details.

Some other changes:
* fix API Service Config YAML parsing
* switch from gopkg.in/yaml to ghodss/yaml
* update bazel deps accordingly

Fixes #845


go.mod
```diff
@@ -1,13 +1,13 @@
...
 
 require (
+	github.com/ghodss/yaml v1.0.0
 	github.com/golang/protobuf v1.5.2
 	github.com/google/go-cmp v0.5.7
 	github.com/jhump/protoreflect v1.12.0
 	gitlab.com/golang-commonmark/markdown v0.0.0-20211110145824-bf3e522c626a
 	google.golang.org/genproto v0.0.0-20220414192740-2d67ff6cf2b4
 	google.golang.org/protobuf v1.28.0
-	gopkg.in/yaml.v2 v2.4.0
```
codyoss, line 10: Out of curiosity, any reason for this switch? This is what I would consider the more popular lib in the ecosystem

internal/gengapic/generator.go
```diff
@@ -99,17 +101,22 @@ func (g *generator) init(req *plugin.CodeGeneratorRequest) error {
...
-		err = yaml.NewDecoder(f).Decode(g.serviceConfig)
+		j, err := yaml.YAMLToJSON(y)
 		if err != nil {
-			return errors.E(nil, "error decoding service config: %v", err)
+			return errors.E(nil, "error converting YAML to JSON: %v", err)
+		}
+
+		cfg := &serviceconfig.Service{}
+		unm := protojson.UnmarshalOptions{DiscardUnknown: true}
+		if err := unm.Unmarshal(j, cfg); err != nil {
```
codyoss, line 116: nit: this and the above line could be combined.

internal/gengapic/genrest.go
```diff
@@ -28,6 +28,8 @@ import (
 	"google.golang.org/protobuf/proto"
 )
 
+var patternRegex = regexp.MustCompile(`{([a-zA-Z0-9_.]+?)(=[^{}]+)?}`)
```
codyoss, line 31: nit: maybe use a slightly more descriptive name here. What kind of pattern is this?

codyoss (approved):
Just a couple of small things, or else looks good to me.

go.mod
noahdietz: `gopkg.in/yaml.v2` doesn't have a Yaml-to-JSON converter, which is what I need to correctly parse the contents using the protojson decoder. There isn't a protoyaml decoder, so :)

internal/gengapic/genrest.go
noahdietz: Switched to `httpPatternVarRegex`. Done.

internal/gengapic/generator.go
noahdietz: Done.

PR #947, opened 2022-03-30
fix(bazel): add repo_mapping to all go_repository deps

noahdietz:
In order to align on the `go_proto_library` targets used during generator compilation, map use of `@go_googleapis` to `@com_google_googleapis` in every `go_repository` dependency listed in `repositories.bzl`.

cc @vam-google


vam-google (approved):
LGTM

codyoss (approved):

PR #957, opened 2022-04-11
fix(gengapic): REGAPIC support gax.CallSettings.Path override

noahdietz:
Part of #845, this is necessary so that we can regenerate the common LRO GAPIC with REGAPIC support to accept a Path override.


codyoss (approved):
LGTM

PR #993, opened 2022-05-12
feat(gengapic): REGAPIC retry support

### LINKED ISSUE #843: gapic: REGAPIC calls must have default client-side retry
Blocked by #842.

The default client-side retry config used for gRPC can be translated in terms of HTTP codes and generated as an [`gax.OnErrorFunc`](https://pkg.go.dev/github.com/googleapis/gax-go/v2#OnErrorFunc) default call option.

noahdietz:
This adds generation for default retry config for REGAPIC clients. The status codes in the retry configuration used today for gRPC are converted into HTTP status codes according to a canonical mapping that all languages are using.

Fixes #843.

### DIFF
--- a/internal/gengapic/genrest.go
+++ b/internal/gengapic/genrest.go
@@ -1184,7 +1184,7 @@ func (g *generator) restCallOptions(serv *descriptor.ServiceDescriptorProto, ser
 		sFQN := g.fqn(g.descInfo.ParentElement[m])
 		mn := m.GetName()
 		p("%s: []gax.CallOption{", mn)
-		if rp, ok := c.RetryPolicy(sFQN, mn); ok && rp != nil {
+		if rp, ok := c.RetryPolicy(sFQN, mn); ok && rp != nil && len(rp.GetRetryableStatusCodes()) > 0 {
 			p("gax.WithRetry(func() gax.Retryer {")
 			p("  return gax.OnHTTPCodes(gax.Backoff{")
 			// this ignores max_attempts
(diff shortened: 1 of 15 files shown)


internal/gengapic/genrest.go
```diff
@@ -1126,3 +1155,48 @@ func (g *generator) unaryRESTCall(servName string, m *descriptor.MethodDescripto
...
+			p("gax.WithRetry(func() gax.Retryer {")
+			p("  return gax.OnHTTPCodes(gax.Backoff{")
+			// this ignores max_attempts
+			p("    Initial:    %d * time.Millisecond,", conf.ToMillis(rp.GetInitialBackoff()))
+			p("    Max:        %d * time.Millisecond,", conf.ToMillis(rp.GetMaxBackoff()))
+			p("    Multiplier: %.2f,", rp.GetBackoffMultiplier())
+			p("	 },")
+
+			rc := rp.GetRetryableStatusCodes()
+			for ndx, c := range rc {
```
codyoss, line 1184: Is it possible for rc to be nil here? Do we first need to check if there are codes before adding the retry?

codyoss (approved):
LGTM, one question

internal/gengapic/genrest.go
noahdietz: It is _possible_, but would be considered a misconfiguration. If a method should not be retried, it shouldn't have a retry policy at all. I will add a check though, no point in not :)

noahdietz:
Going to hold off on submitting this PR until gax-go v2.4.0 has landed in google-cloud-go. Don't want any accidental regens with this producing broken code or blocking integration of a fix.

noahdietz:
FYI: Rebased onto main in order to fix merge conflicts with #994

PR #994, opened 2022-05-12
fix(gengapic): REGAPIC encode enums as numbers

noahdietz:
Part of our REGAPIC effort to standardize JSON (de)serialization of enums as numbers to avoid the "unknown enum" error that comes with using the string identifiers instead.

Note: I've removed the use of `UseProtoNames: false` option because it is `false` by default so stating that explicitly is unnecessary and it was done inconsistently.


codyoss (approved):
LGTM

vchudnov-g (commented):
LGTM

PR #1000, opened 2022-05-17
Update cli generator to handle LROs with empty response types.

timburks:
I have a generated CLI for an API that just added an LRO with an empty response type:
```
  rpc DeleteInstance(DeleteInstanceRequest) returns (google.longrunning.Operation) {
    option (google.api.http) = {
      delete: "/v1/{name=projects/*/locations/*/instances/*}"
    };
    option (google.api.method_signature) = "name";
    option (google.longrunning.operation_info) = {
      response_type: "google.protobuf.Empty"
      metadata_type: "OperationMetadata"
    };
  }
```

Because the response type was empty, the generated grpc `Wait()` and `Poll()` methods returned single values (`error` only), which broke the generated CLI.

This change gets the generated CLI to compile but I have not yet verified the functionality. (I'll follow up when I do)

### DIFF
--- a/internal/gencli/cmd_file.go
+++ b/internal/gencli/cmd_file.go
@@ -407,7 +407,11 @@ var {{ $pollingCmdVar }} = &cobra.Command{
 		}
 		{{ end }}
 
-		fmt.Println(fmt.Sprintf("Operation %s not done", op.Name()))
+		if op.Done() {
+			fmt.Println(fmt.Sprintf("Operation %s is done", op.Name()))
+		} else {
+			fmt.Println(fmt.Sprintf("Operation %s not done", op.Name()))
+		}
 
 		return err
 	},
--- a/internal/gencli/cmd_file.go
+++ b/internal/gencli/cmd_file.go
@@ -352,7 +352,8 @@ var {{$methodCmdVar}} = &cobra.Command{
 			return err
 		}
 
-		result, err := resp.Wait(ctx)
+		{{ if .IsLRORespEmpty }}err = resp.Wait(ctx)
+		{{else}}result, err := resp.Wait(ctx)
 		if err != nil {
 			return err
 		}
@@ -360,7 +361,7 @@ var {{$methodCmdVar}} = &cobra.Command{
 		if Verbose {
 			fmt.Print("Output: ")
 		}
-		printMessage(result)
+		printMessage(result){{ end }}
 		{{ end }}
 		{{ end }}
 		return err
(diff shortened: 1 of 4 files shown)


internal/gencli/cmd_file.go
```diff
@@ -398,6 +410,7 @@ var {{ $pollingCmdVar }} = &cobra.Command{
 			printMessage(resp)
 			return
 		}
+		{{ end }}
 
 		fmt.Println(fmt.Sprintf("Operation %s not done", op.Name()))
```
timburks, line 415: I'm not sure that either the original code or the code that I added is determining this correctly.

From [this documentation](https://pkg.go.dev/cloud.google.com/go/longrunning#Operation.Poll), I think we should be checking `op.Done()` to determine that the operation is done.

WDYT?

noahdietz: Yeah so the previous code was (improperly) depending on the presence of a response message (i.e. it is non-`nil`) to determine if it was done. Now that we want to handle google.protobuf.Empty where there will be no response message returned by the operation, we should use `op.Done()` to check. 

You are right.

noahdietz (approved):
LGTM thanks for adding this. The `op.Done()` change could happen in a separate PR if you want, either way.

timburks:
@noahdietz I think this can be merged at your convenience. Thanks!

internal/gencli/cmd_file.go
```diff
@@ -398,8 +405,13 @@ var {{ $pollingCmdVar }} = &cobra.Command{
 			printMessage(resp)
 			return
 		}
+		{{ end }}
 
-		fmt.Println(fmt.Sprintf("Operation %s not done", op.Name()))
+		if op.Done() {
```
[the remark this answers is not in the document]
noahdietz: Looks great.

timburks:
Here's what I get now for my API:
```
$ registry rpc provisioning poll-create-instance --operation projects/1077573052867/locations/us-central1/operations/operation-1652907533047-5df4f865a9595-7ca43785-420d3f2e
Operation projects/1077573052867/locations/us-central1/operations/operation-1652907533047-5df4f865a9595-7ca43785-420d3f2e not done

$ registry rpc provisioning poll-create-instance --operation projects/1077573052867/locations/us-central1/operations/operation-1652907533047-5df4f865a9595-7ca43785-420d3f2e
name:"projects/apiglot/locations/us-central1/instances/default"  create_time:{seconds:1652907533  nanos:165898512}  update_time:{seconds:1652907533  nanos:165898512}  state:ACTIVE  state_message:"Instance is active and ready to use."  config:{location:"us-central1"  cmek_key_name:"projects/apiglot/locations/us-central1/keyRings/r/cryptoKeys/k"}
```

PR #1027, opened 2022-06-09
chore(showcase): lookup showcase version in go.mod

noahdietz:
Allows us to remove the `showcaseSemver` variable and update the gapic-showcase dependency used in our integration tests via renovate.


codyoss (approved):
LGTM

PR #1029, opened 2022-06-10
feat(genpapic): support protobuf-go go_package mapping option

noahdietz:
Per the [protobuf Go docs](https://developers.google.com/protocol-buffers/docs/reference/go-generated#package), there is an option to override the `go_package` option in specific proto files. If this is used during generation of the protobuf/grpc stubs, the GAPIC generator needs to know about it, so that the proper imports can be generated for the request/response/stub types.


codyoss (approved):
LGTM!

quartzmo (approved):

Issue #1058, opened 2022-06-29
fmt.Sprintf call has arguments but no formatting directives


codyoss:
This error has been observed when trying to generate REGAPICS. There were a handful of libraries that were manually excluded from the initial beta/alpha generation of REGAPICS due to this error and now it seems some more are starting to have it. In https://github.com/googleapis/google-cloud-go/pull/6269 It was observed in both security/privateca and gaming:

```
# cloud.google.com/go/security/privateca/apiv1beta1
privateca/apiv1beta1/certificate_authority_client.go:1968:25: fmt.Sprintf call has arguments but no formatting directives
privateca/apiv1beta1/certificate_authority_client.go:2040:25: fmt.Sprintf call has arguments but no formatting directives
privateca/apiv1beta1/certificate_authority_client.go:2103:25: fmt.Sprintf call has arguments but no formatting directives
privateca/apiv1beta1/certificate_authority_client.go:2166:25: fmt.Sprintf call has arguments but no formatting directives
privateca/apiv1beta1/certificate_authority_client.go:2434:25: fmt.Sprintf call has arguments but no formatting directives
privateca/apiv1beta1/certificate_authority_client.go:2497:25: fmt.Sprintf call has arguments but no formatting directives
privateca/apiv1beta1/certificate_authority_client.go:2571:25: fmt.Sprintf call has arguments but no formatting directives
privateca/apiv1beta1/certificate_authority_client.go:2791:25: fmt.Sprintf call has arguments but no formatting directives
privateca/apiv1beta1/certificate_authority_client.go:2961:25: fmt.Sprintf call has arguments but no formatting directives
privateca/apiv1beta1/certificate_authority_client.go:3043:25: fmt.Sprintf call has arguments but no formatting directives
privateca/apiv1beta1/certificate_authority_client.go:3125:25: fmt.Sprintf call has arguments but no formatting directives
privateca/apiv1beta1/certificate_authority_client.go:3207:25: fmt.Sprintf call has arguments but no formatting directives
privateca/apiv1beta1/certificate_authority_client.go:3289:25: fmt.Sprintf call has arguments but no formatting directives
privateca/apiv1beta1/certificate_authority_client.go:3371:25: fmt.Sprintf call has arguments but no formatting directives
privateca/apiv1beta1/certificate_authority_client.go:3453:25: fmt.Sprintf call has arguments but no formatting directives
[...]

PR #1072, opened 2022-07-15
fix(gengapic): regapic GetOperation path fallback logic

quartzmo:
fixes: #1058


internal/gengapic/genrest.go
```diff
@@ -937,7 +937,7 @@ func (g *generator) lroRESTCall(servName string, m *descriptor.MethodDescriptorP
 	p("  return nil, e")
 	p("}")
 	p("")
-	override := g.getOperationPathOverride()
+	override := g.getOperationPathOverride(g.descInfo.ParentFile[serv].GetPackage())
```
noahdietz, line 940: I just realized in reviewing this is that we can use the `m *descriptor.MethodDecriptorProto` here instead of `serv`, since every proto element is tied back to its parent file descriptor for this exact case. So...we don't need to add a `serv` param, and we can instead just use the `m` MethodDescriptor here.

quartzmo: Fixed, thanks!

noahdietz (approved):

PR #1124, opened 2022-09-01
fix: regapic support for proto wkt in query params

noahdietz:
Properly handle protobuf well known message types that have special JSON mappings e.g. `google.protobuf.FieldMask` when they appear as query parameters.

The list of well known types is defined here: https://developers.google.com/protocol-buffers/docs/proto3#json

Update gapic-showcase to v0.25.0 that exercises this code, and update the Identity service tests to run against REST as well, to exercise the FieldMask encoding.

### DIFF
--- a/internal/gengapic/genrest.go
+++ b/internal/gengapic/genrest.go
@@ -420,9 +420,8 @@ func (g *generator) generateQueryString(m *descriptor.MethodDescriptorProto) {
 		singularPrimitive := field.GetType() != fieldTypeMessage &&
 			field.GetType() != fieldTypeBytes &&
 			field.GetLabel() != fieldLabelRepeated
-		g.imports[pbinfo.ImportSpec{Path: "fmt"}] = true
-		paramAdd := fmt.Sprintf("params.Add(%q, fmt.Sprintf(%q, req%s))", lowerFirst(snakeToCamel(path)), "%v", accessor)
 
+		var paramAdd string
 		// Handle well known protobuf types with special JSON encodings.
 		if strContains(wellKnownTypes, field.GetTypeName()) {
 			b := strings.Builder{}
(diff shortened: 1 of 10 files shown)


codyoss (approved):
Thanks!

internal/gengapic/genrest.go
```diff
@@ -406,6 +423,21 @@ func (g *generator) generateQueryString(m *descriptor.MethodDescriptorProto) {
...
+			b.WriteString(fmt.Sprintf("%s, err := protojson.Marshal(req%s)\n", field.GetJsonName(), accessor))
+			b.WriteString("if err != nil {\n")
+			if m.GetOutputType() == emptyType {
+				b.WriteString(fmt.Sprintf("  return err\n"))
+			} else {
+				b.WriteString(fmt.Sprintf("  return nil, err\n"))
+			}
+			b.WriteString("}\n")
+			b.WriteString(fmt.Sprintf("params.Add(%q, string(%s))", lowerFirst(snakeToCamel(path)), field.GetJsonName()))
+			paramAdd = b.String()
```
vchudnov-g, line 438: So you're overwriting the `parmAdd` from above, line 424? What about  registering the `fmt` import in line 423? Is that still valid? I know Go style tries to avoid `else`s, but given the non-trivial processing in both branches I think it might be clearer to organize this as
```
if wellKnownType {
…
} else {
…
}
```

showcase/identity_test.go
```diff
@@ -23,134 +23,139 @@ import (
 	showcasepb "github.com/googleapis/gapic-showcase/server/genproto"
 	"google.golang.org/api/iterator"
 	"google.golang.org/protobuf/proto"
+	"google.golang.org/protobuf/types/known/fieldmaskpb"
 )
 
 var identity *showcase.IdentityClient
+var identityREST *showcase.IdentityClient
```
vchudnov-g, line 30: FWIW, I would find it helpful to include a comment that these are initialized in `main_test.go`. I was confused how these differed until I saw that.

noahdietz: Fair enough! Done (and applied to all such `var`).

internal/gengapic/genrest.go
noahdietz: Good point about the `fmt` import, I need to conditionally include that. 

I'm not sure I understand the scope of your "if/else" comment though. Do you just mean WRT when to import `fmt`? If so, yeah that is what I plan to do :)

noahdietz: PTAL :)

vchudnov-g: Yeah, that's exactly what I meant. Thanks!

vchudnov-g (approved):
Looks great. Thanks for doing this!

PR #1132, opened 2022-09-13
fix: handle wkt in paged requests

noahdietz:
For Standard List methods, and most other paginated methods, having fields in the request message outside of a resource name and the standard fields is non-compliant. However, there is at least one API that utilizes such non-standard fields, one of them being a well-known type that would end up being encoded in the query params. So that case needs to be supported.


codyoss (approved):
Thank you!

PR #1161, opened 2022-10-20
fix(gengapic): separate repeated prim qp values

### LINKED ISSUE #1160: Repeated fields in parameter are set as a single string v.s. a slice of string (incompatible with `grpc-gateway`)
I am using this client generator to generate clients that can talk to my server via  REST/GRPC in an easy approach. And I am using `grpc-gateway` for routing the HTTP/REST requests to GRPC. I thought both this and `grpc-gateway` are popular tools and they all generate code from protobuf so that they should be compatible with each other.
However I noticed that the repeated field parameter didn't work well.

Could you please help to check if this is a bug/issue, or if this is a feature that is nice to have?
I think people would be more than happy if the generated REST Clients can be directly used against `grpc-gateway`!

Thanks!

#### Environment details

  - Programming language: Golang
  - OS: MacOS
  - Language runtime version: v1.17.13
  - Package version:

#### Steps to reproduce
1. 
My proto is similar to this:
```proto
rpc List(ListRequest) returns (ListResponse) {
    option (google.api.http) = {
      get: "/v1/{parent=*}/lists"
    };
}

message ListRequest {
    enum Types {
        A=0;
        B=1;
    }
    repeated Types = 1; 
}
```
So I can send a request with Types = [0, 1].

2. 
And in the generated REST client's `List` method, the `Types` is set in this way
```
		if req.GetTypes() != nil {
			params.Add("types", fmt.Sprintf("%v", req.GetTypes()))
		}
```
This sets a single string `[0,1]` to the parameter `types`.

3. 
Whereas in the server side, `grpc-gateway` receives the request and parses the parameter using [this method: Parse](https://github.com/grpc-ecosystem/grpc-gateway/blob/v1/runtime/query.go#L37), where the input `values` is the `request.Form` which contains the parameter k-v pair.

And it will parse the above `types: "[0,1]"` to `key: types`, `value: "[0,1]"` (Note, it is a string `"[0,1]"`). And it will fail saying that `[0,1]` is not a correct enum type.

4. 
My solution is to adjust the above code in the generated REST clients:
```
		if req.GetTypes() != nil {
			for _,n := range req.GetTypes(){
				params.Add("types", fmt.Sprintf("%v", n))
			}
		}
```
to add multiple strings for the types and this will work perfectly with `grpc-gateway`.
This will set the parameter fields with a slice of strings `0`, `1` and `grpc-gateway` will parse and return correct values.

noahdietz:
For REGAPIC query params, repeated non-message fields should be included in the query params as multiple, individual key-value pairs using the same field name as the key and each item in the slice as a singular value.

Fixes #1160

### DIFF
--- a/internal/gengapic/genrest.go
+++ b/internal/gengapic/genrest.go
@@ -420,6 +420,7 @@ func (g *generator) generateQueryString(m *descriptor.MethodDescriptorProto) {
 		singularPrimitive := field.GetType() != fieldTypeMessage &&
 			field.GetType() != fieldTypeBytes &&
 			field.GetLabel() != fieldLabelRepeated
+		key := lowerFirst(snakeToCamel(path))
 
 		var paramAdd string
 		// Handle well known protobuf types with special JSON encodings.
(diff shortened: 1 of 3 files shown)


codyoss (approved):
LGTM

PR #1175, opened 2022-11-01
fix(gengapic): move top-level package doc links to top

### LINKED ISSUE #1140: gengapic: move top-level package docs link to more prominent position
The top-level Go Client docs contain a lot of information that applies to most/all of the clients it contains: https://pkg.go.dev/cloud.google.com/go.

In GAPICs, the link to this documentation is kind of buried. It would be better if it had its own header or a more prominent position so that it is easier to find. Currently, it is the last thing in the package level documentation under a more specific header.

image

adrianajg:
Fixes #1140


noahdietz:
Looks great! Can you add a short description and mention that this "fixes" the issue #1140 ? That way, once this is merged, the issue will be closed as fixed automatically.

noahdietz (approved):

PR #1199, opened 2022-12-07
fix(gengapic): only gen REST client for RESTable services

noahdietz:
This is a fix, and introduces the possibility of a breaking change, in generation of clients when `rest` is a selected transport. If a proto service being generated doesn't have any RPCs with `google.api.http` annotations i.e. they are not "REST-able", then a REST client is not generated for the proto service.

I've added a sanity check but based on where the code is placed it is really hard to test without setting up the entire generator in code. I have tested this in local generation and with the known example added to `test.sh`.

### DIFF
--- a/internal/gengapic/gengapic.go
+++ b/internal/gengapic/gengapic.go
@@ -101,6 +101,13 @@ func Gen(genReq *plugin.CodeGeneratorRequest) (*plugin.CodeGeneratorResponse, er
 		outFile = filepath.Join(g.opts.outDir, outFile)
 
 		g.reset()
+		// If the service has no REST-able RPCs, then a REGAPIC should not be
+		// generated for it, even if REST is an enabled transport.
+		transports := g.opts.transports
+		hasREST := hasRESTMethod(s)
+		if !hasREST {
+			g.opts.transports = []transport{grpc}
+		}
 		if err := g.gen(s); err != nil {
 			return &g.resp, err
 		}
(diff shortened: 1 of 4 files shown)


internal/gengapic/gengapic.go
```diff
@@ -112,6 +119,12 @@ func Gen(genReq *plugin.CodeGeneratorRequest) (*plugin.CodeGeneratorResponse, er
 		}
 		g.imports[pbinfo.ImportSpec{Name: g.opts.pkgName, Path: g.opts.pkgPath}] = true
 		g.commit(outFile+"_client_example_test.go", g.opts.pkgName+"_test")
+
+		// Replace original set of transports for the next service that may have
+		// REST-able RPCs.
+		if !hasREST {
+			g.opts.transports = transports
```
codyoss, line 126: nit: you could use defer above for this case.

codyoss (approved):
Two small comments to check or else looks good

internal/gengapic/helpers.go
```diff
@@ -122,6 +122,19 @@ func hasMethod(service *descriptor.ServiceDescriptorProto, method string) bool {
 	return false
 }
 
+// hasRESTMethod reports if there is at least one RPC on the Service that
+// has a gRPC-HTTP transcoding, or REST, annotation on it.
+func hasRESTMethod(service *descriptor.ServiceDescriptorProto) bool {
+	for _, m := range service.GetMethod() {
+		eHTTP := proto.GetExtension(m.GetOptions(), annotations.E_Http)
+		if h := eHTTP.(*annotations.HttpRule); h.GetPattern() != nil {
```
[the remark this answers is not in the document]
noahdietz: No, because `proto.GetExtension` will always return a `HttpRule`. If the annotation is unset, a default one is returned. If the annotation is present, then it will be a `HttpRule`.

internal/gengapic/gengapic.go
codyoss: Ah, did not see this was in a loop from GH

PR #1220, opened 2023-01-27
feat(gengapic): add snippets

quartzmo:
The snippet output added in this change is known to differ from legacy snippetgen output as follows:

1. In new snippet metadata, `resultType` pointer values contain the `*` prefix. (In current metadata, the `*` does not appear.)
2. In new snippet file license headers, `https` replaces current headers' usage of `http`.
3. In new snippet file "generated by" header comments, the source is changed to the `protoc` plugin.
4. In new snippet file imports, a blank line that currently appears before `iam` imports is now removed. Note that this decreases the `segments.end` value by `1` in the new snippet metadata.
   ```diff
           datacatalog "cloud.google.com/go/datacatalog/apiv1beta1"
   -
           iampb "google.golang.org/genproto/googleapis/iam/v1"
   ```
5. In new snippet metadata, `clientMethod.method` values are changed to the actual protobuf source values for mixin methods:
   ```diff
            "method": {
              "shortName": "GetLocation",
   -          "fullName": "google.cloud.kms.v1.EkmService.GetLocation",
   +          "fullName": "google.cloud.location.Locations.GetLocation",
              "service": {
   -            "shortName": "EkmService",
   -            "fullName": "google.cloud.kms.v1.EkmService"
   +            "shortName": "Locations",
   +            "fullName": "google.cloud.location.Locations"
              }
   ```


internal/snippets/snippets.go
```diff
@@ -0,0 +1,279 @@
...
+	pkgName string
+}
+
+// NewMetadata initializes the model that will collect snippet metadata, from:
+// protoPkg - dot-separated, without final type name element (e.g. "google.cloud.bigquery.migration.v2")
+// libPkg - the Go import path for the GAPIC client, per libraryPackage in gapic_metadata.json (e.g. "cloud.google.com/go/bigquery/migration/apiv2")
+// pkgName - stored in g.opts.pkgName, used as an argument to pbinfo.ReduceServName
+func NewMetadata(protoPkg, libPkg, pkgName string) (*SnippetMetadata, error) {
+	protoParts := strings.Split(protoPkg, ".")
+	apiVersion := protoParts[len(protoParts)-1]
```
noahdietz, line 71: Nit: Maybe just this:
```suggestion
	lastDot := strings.LastIndex(protoPkg, ".")
	apiVersion := protoPkg[lastDot+1:]
```

Saves an intermediate array/slice.

internal/snippets/snippets.go
```diff
@@ -0,0 +1,279 @@
...
+						Service: &metadata.Service{
+							ShortName: service.protoName,
+							FullName:  fmt.Sprintf("%s.%s", sm.protoPkg, service.protoName),
+						},
+					},
+				},
+			}
+			segment := &metadata.Snippet_Segment{
+				Start: int32(method.regionTagStart + 1),
+				End:   int32(method.regionTagEnd - 1),
```
noahdietz, line 223: Could you add a comment explaining why we must +/- 1 on these values? Just for future reference.

noahdietz (commented):
This is pretty cool, @quartzmo well done.

internal/gengapic/gengapic.go
```diff
@@ -516,10 +565,20 @@ func (g *generator) methodDoc(m *descriptor.MethodDescriptorProto) {
...
+	if !g.opts.omitSnippets && m.GetClientStreaming() == m.GetServerStreaming() {
+		g.snippetMetadata.UpdateMethodDoc(serv.GetName(), m.GetName(), com)
+	}
+
 	g.comment(com)
 }
 
-func (g *generator) comment(s string) {
+// printFunc defines a function type accepting arguments to be format-printed.
+type printFunc func(s string, a ...interface{})
```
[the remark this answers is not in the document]
quartzmo: I'm not sure why I didn't just do that in the first place. I'll try it.

internal/gengapic/gengapic.go
```diff
@@ -488,7 +537,7 @@ func containsDeprecated(com string) bool {
 	return false
 }
 
-func (g *generator) methodDoc(m *descriptor.MethodDescriptorProto) {
+func (g *generator) methodDoc(m *descriptor.MethodDescriptorProto, serv *descriptor.ServiceDescriptorProto) {
```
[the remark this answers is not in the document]
quartzmo: I tried this as we discussed, but there was an error in integration testing in which the lookup failed. Should we investigate that error, or is it a valid state for `descInfo` to not have a parent element mapping for a method?

noahdietz:
Can you add a new config file for SnippetBot to stop it from bothering us about region tags in generator code? Add the following to `.github/snippet-bot.yml`:
```yaml
# The generator prints region tags and snippet-bot wants to correct
# the printing code.
ignoreFiles:
  - "internal/**"
```

internal/gengapic/gengapic.go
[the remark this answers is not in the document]
noahdietz: It is not valid for a Method to not have a Parent mapped, so that would be an issue we should investigate. LMK if you want to pair on it.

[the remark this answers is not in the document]
quartzmo: Yes, let's do this on a new branch off of main.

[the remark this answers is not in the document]
quartzmo: My apologies for the false alarm: After reproducing it again, it looks like the lookup issue is in finding the matching service in the snippet metadata, and that I just need to re-reduce the service name.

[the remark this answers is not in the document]
quartzmo: Actually, this turns out to be interesting and a bit subtle: The lookup failure in the showcase integration test is for the `ListLocations` method, the parent of which in `descInfo` is `Locations`. In the snippets domain, the "service" containing this method is derived from the client in which it appears (it is a mixin), for example, `Echo`, as seen in the metadata below:

```json
        "method": {
          "shortName": "ListLocations",
          "fullName": "google.showcase.v1beta1.Echo.ListLocations",
          "service": {
            "shortName": "Echo",
            "fullName": "google.showcase.v1beta1.Echo"
          }
        }
```

I need to verify that the metadata above is correct. Alternatively, it may be that the `service` in the metadata (and the entire snippets domain) should be the true service for the mixin (e.g. `Locations`), not the primary service for the client.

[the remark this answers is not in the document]
quartzmo: Here is another example from [currently published snippet metadata](https://github.com/googleapis/google-cloud-go/blob/bigquery/v1.47.0/internal/generated/snippets/bigquery/datatransfer/apiv1/snippet_metadata.google.cloud.bigquery.datatransfer.v1.json#L499-L506) for google-cloud-go:

```json
        "method": {
          "shortName": "ListLocations",
          "fullName": "google.cloud.bigquery.datatransfer.v1.DataTransferService.ListLocations",
          "service": {
            "shortName": "DataTransferService",
            "fullName": "google.cloud.bigquery.datatransfer.v1.DataTransferService"
          }
        }
```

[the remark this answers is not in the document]
quartzmo: @amanda-tarafa Is the current snippetgen metadata implementation in google-cloud-go incorrect? Should the above fragment instead be:

```json
        "method": {
          "shortName": "ListLocations",
          "fullName": "google.cloud.location.Locations.ListLocations",
          "service": {
            "shortName": "Locations",
            "fullName": "google.cloud.location.Locations"
          }
        }
```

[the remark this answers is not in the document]
amanda-tarafa: My inclination would be the metadata as you are proposing and not as it is now.

You are effectively saying, my client library method actually refers to this service method.
The fact that the actual endpoint is different for each library that has the mixin is an implementation detail that we don't care to surface in samples I think, samples are about interface, not about implementation. And even if we were going to surface that information somehow, saying the the locations service is in the package of the API using the mixing seems the wrong way to do that, as it's not accurate.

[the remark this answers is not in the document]
quartzmo: @amanda-tarafa Thank you, I will update this PR to implement `clientMethod.method` in the snippet metadata as you describe, so that it references the original proto values that are provided to the generator. The currently published snippet metadata will be updated when this change is released in the generator and replaces the current implementation in google-cloud-go.

internal/snippets/snippets.go
quartzmo: Done, thanks!

internal/snippets/snippets.go
quartzmo: Done, thanks!

internal/gengapic/gengapic.go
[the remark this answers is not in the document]
quartzmo: Done, thanks!

quartzmo:
@noahdietz I don't know why the snippet-bot checks are still failing after I added the `ignoreFiles` directive, which LGTM.

noahdietz:
> @noahdietz I don't know why the snippet-bot checks are still failing after I added the `ignoreFiles` directive, which LGTM.

It might need to be checked into the repo to actually work. In which case, you could either put it in another PR to merge sooner, or just wait until this merges and ignore the Action for now :) Either way, I'm fine.

internal/gengapic/gengapic.go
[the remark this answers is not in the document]
noahdietz: That makes a lot of sense to me. Thanks folks, sgtm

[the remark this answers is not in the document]
quartzmo: See updated PR description, number 5, above, for an example of this change.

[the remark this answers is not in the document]
quartzmo: Fixed.

internal/gengapic/example_test.go
```diff
@@ -276,6 +277,92 @@ func TestExample(t *testing.T) {
...
+		Options: &descriptor.FileOptions{
+			GoPackage: proto.String("cloud.google.com/go/bigquery/migration/apiv2/migrationpb"),
+		},
+		Package: proto.String(protoPkg),
+	}
+
+	files := []*descriptor.FileDescriptorProto{}
+	g.descInfo = pbinfo.Of(files)
+	for _, typ := range []*descriptor.DescriptorProto{
+		inputType, outputType,
```
codyoss, line 315: nit: Please always provide struct keys.

internal/gengapic/gengapic.go
```diff
@@ -95,15 +97,53 @@ func Gen(genReq *plugin.CodeGeneratorRequest) (*plugin.CodeGeneratorResponse, er
...
 		// the file name is "logging_client.go".
 		// Keep the current behavior for now, but we could revisit this later.
-		outFile := pbinfo.ReduceServName(s.GetName(), "")
-		outFile = camelToSnake(outFile)
+		servName := pbinfo.ReduceServName(s.GetName(), "")
+		outFile := camelToSnake(servName)
 		outFile = filepath.Join(g.opts.outDir, outFile)
 
+		if !g.opts.omitSnippets {
+			defaultHost := proto.GetExtension(s.Options, annotations.E_DefaultHost).(string)
```
codyoss, line 119: There are a couple of type casts in this method. Are we posative of thier value or should they be checked to avoid panics?

internal/gengapic/gengapic.go
```diff
@@ -95,15 +97,53 @@ func Gen(genReq *plugin.CodeGeneratorRequest) (*plugin.CodeGeneratorResponse, er
...
 		// so even though the client for LoggingServiceV2 is just "Client"
 		// the file name is "logging_client.go".
 		// Keep the current behavior for now, but we could revisit this later.
-		outFile := pbinfo.ReduceServName(s.GetName(), "")
-		outFile = camelToSnake(outFile)
+		servName := pbinfo.ReduceServName(s.GetName(), "")
+		outFile := camelToSnake(servName)
 		outFile = filepath.Join(g.opts.outDir, outFile)
 
+		if !g.opts.omitSnippets {
```
codyoss, line 118: Not saying we should do it now. But I think if we ever refactor in the future we should avoid all of these if checks in the main generation methods. I think it would be cleaner to call methods and just have them do nothing if flags are not set. I think it would make the code more readable. I know there is presidence for this pattern rn though. But something we should consider should we do a rewrite of this package.

internal/gengapic/gengapic_test.go
```diff
@@ -485,11 +493,14 @@ methods:
 	} {
 		g.reset()
 		g.descInfo.ParentElement[tst.m] = serv
+		serv.Method = []*descriptor.MethodDescriptorProto{
+			tst.m,
+		}
```
codyoss, line 498: nit: add key

codyoss (commented):
Just a couple of small things

internal/gengapic/example_test.go
quartzmo: `inputType` and `outputType` hold pointers to `descriptor.DescriptorProto` instances. In this line they are being added to to a new slice of `[]*descriptor.DescriptorProto`. That's ok, isn't it?

quartzmo: I can change this to 4 lines of code that directly call `g.descInfo` methods for `inputType` and `outputType`. This loop pattern is well established in the tests. It makes more sense when there are more elements in the slice, for example [here](https://github.com/googleapis/gapic-generator-go/blob/d5ed011a154765d2ff8fa555ccaad35ce4fd1dcd/internal/gengapic/example_test.go#L121-L126). But if you ask me, loops almost always reduce the clarity of the tests. I personally think it's usually better to have more lines of simple setup statements. I was just sticking to established style here.

internal/gengapic/gengapic.go
quartzmo: I'm open to attempting this refactoring now to see how it looks. The new method would probably need a quite a few parameters, but it is probably worth it. I agree with this suggestion.

internal/gengapic/gengapic.go
quartzmo: I think the value is quite certain, but good call, will add. Thanks!

internal/gengapic/gengapic_test.go
quartzmo: This is the inline creation of a slice with a single element. It's very confusing, but `ServiceDescriptorProto.Method` is a collection of `[]*MethodDescriptorProto`. It should have been named `Methods`, right?

internal/gengapic/gengapic.go
quartzmo: @noahdietz In addition to this statement, there are 6 existing calls to `proto.GetExtension` that cast to a string without checking `ok`. Are they all good as-is, or do you see any reason to add the checks?

internal/gengapic/gengapic.go
codyoss: I think we should land this first regardless as this is already a large PR

internal/gengapic/gengapic.go
noahdietz: The value type of a proto extension will never change (breaking change) and if the annotation is not present, this getter will return the default value (e.g. `""` for default_host). I don't think we need to check the `ok`.

noahdietz (commented):
I am confused why changes that have already landed in `main` are showing up in the diffs here, is anyone else seeing that? Like `ci.yaml` and `deps.yaml` changing to bazel verison 6.0.0. And the CHANGELOG.md including the latest release...why are they in this PR?

internal/gengapic/gengapic.go
quartzmo: I pushed a commit moving this logic into a new file `internal/gengapic/snippets.go`, which I think cleans up the existing files quite a bit, at the cost of a bit of indirection. It may also make it possible to improve the test coverage.

quartzmo:
@noahdietz I also see those diffs in the file view for this PR in GH. But I do not see them locally. The following returns no differences:

```sh
git diff main -- .github/workflows/ci.yaml
```

I guess maybe GH might be confused by so many commits?

noahdietz (approved):
Good call on doing the conditional refactor now, it made things a lot cleaner.

internal/gengapic/testdata/snippet_BidiThings.want
```diff
@@ -0,0 +1,41 @@
+func main() {
```
noahdietz, line 1: Should there be an opening region tag?

quartzmo: That's a good point. The existing unit tests just diff against `g.pt.String()`, so I followed that pattern, but let me see if I can compare against the entire `g.resp.File` content instead.

quartzmo: I added an assertion for the header content (the start region tag). Dealing with the `g.resp.File` is messy since it isn't reset between test cases. Thanks for catching this!

noahdietz: Good idea!

codyoss (approved):

PR #1247, opened 2023-03-08
feat: update generator for new iam and longrunning locations

codyoss:


codyoss:
@noahdietz If there is a less hacky way to do what I did here I am all ear 😄

noahdietz:
> @noahdietz If there is a less hacky way to do what I did here I am all ears 😄

Not necessarily less hacky, but maybe more holistic is to put this type of override in the `pbinfo` package here: https://github.com/googleapis/gapic-generator-go/blob/main/internal/pbinfo/pbinfo.go#L167-L170. But We will remove this code within the week so I don't really care.

noahdietz (approved):

noahdietz (approved):

Issue #1206, opened 2023-01-04
gengapic: could deadlines be made configurable?


timburks:
gRPC deadlines are currently determined by `timeout` values specified in grpc service configurations ([example](https://github.com/apigee/registry/blob/268e07f266684aa5360fa79150ce5d5843349714/gapic/grpc_service_config.json#L9). 

As we've used the Go GAPICs, we've often wanted to adjust the timeouts used, e.g. [this PR](https://github.com/apigee/registry/pull/900), which for us, involves changing the value in the service config and regenerating the GAPIC. 

Reviewing the changes in the PR above, it seems that it might be easy to replace the hard-coded instances of the timeout value with a reference to a variable that is initialized with the configured value but overrideable through an API -- either by making the timeout variable public or by providing an API to get/set it.

We could patch the generated output, but it seems fairly harmless to add to the core generator. Would you be open to a change like this?

noahdietz:
_tl;dr: Sure let's look at this, but I'd like to look at introducing a GAX option instead._

What you are referring to is modifying the _default_ deadline for a method. The for an individual method call is configurable already, just set a deadline on the `context.Context` given to the RPC. What you want is to be able to change the default deadline for an RPC for the life of a Client _instance_ i.e. `client := NewClient(...); client.CallOptions.GetFoo = append(client.CallOptions.GetFoo, gax.WithTimeout(5 *time.second))`, at least that's how I envision it as a gax call option, which is my preferred means of introducing such a "setting". This `WithTimeout` option would still defer to an existing deadline on the `context.Context`.

This seems pretty reasonable, and probably the easiest way to introduce a client setting that still respects the use of `context.WithTimeout` users my set.

That said, the config should still be the source of the default timeout for an RPC on _all instances_ of the client, if unmodified, as it is today. I don't see anything wrong with your PR, it is the right thing to do, but is cumbersome if you only want to change the timeout for 1 RPC for 1 client instance in a specific application.

@codyoss @quartzmo wdyt?

noahdietz:
After discussing offline with @codyoss this seems like a good idea, and can also be pitched as code simplification/verbosity reduction if we move logic into a gax call option that is then implemented in gax.Invoke.

codyoss:
@noahdietz Would you want to file an issue for gax and move tracking to that issue?

noahdietz:
@codyoss yep, good call, done! https://github.com/googleapis/gax-go/issues/249

timburks:
[...]

PR #1267, opened 2023-03-15
feat(gengapic): use WithTimeout for default logical timeout

### LINKED ISSUE #411: gapic: clients should set a default timeout in the generated code based on the generation config (cloud spanner)

noahdietz:
This switches the implementation of default timeout application to leverage the new `gax.WithTimeout` call option as a default call option instead of having context deadline manipulation code generated into each GAPIC method.

This also removes the code that disables the default deadline logic via an environment environment variable. This was always going to be removed. Closes #411. 

Updates #1206


codyoss (commented):
Looks good overall, can we add a test that shows WithTimeout generated in?

noahdietz:
> Looks good overall, can we add a test that shows WithTimeout generated in?

Yes, definitely, I was actually surprised to see that we didn't have any already.

noahdietz:
> > Looks good overall, can we add a test that shows WithTimeout generated in?
> 
> Yes, definitely, I was actually surprised to see that we didn't have any already.

Done @codyoss

codyoss (approved):
Thanks!

PR #1288, opened 2023-04-06
fix: add wkt desc, bazel snippet copy

noahdietz:
Fixes for snippetgen in a few different ways:
- force include all of the `google.protobuf` well known types so that they are always resolvable (solves issue with `CancelOperation` returning an `Empty` without that proto being loaded via `protoc`)
- remove seemingly extraneous `chmod` in `go_gapic_assembly_pkg` target that was making the `snippets` dir unreadable, and thus un-copyable

Updates `BUILD.bazel` via new target `make gazelle` which runs gazelle, and does some replacements that we need to keep in place (which were done manually before, annoying).

Adds datastore as a sanity test for generation.

All tested in `googleapis` with `datastore`.


rules_go_gapic/go_gapic_pkg.bzl
```diff
@@ -49,10 +49,6 @@ def _go_gapic_src_pkg_impl(ctx):
             mkdir -p {package_dir_path}/${{src_str%:*}}
             cp -f ${{src_str#*:}} {package_dir_path}/${{src_str%:*}}
         fi
-        # dest_dir_path is split again, on the ';' delimiter, for the
-        # Go importpath that includes the package name following ';'
-        s=${{src_str%:*}}
-        chmod 644 {package_dir_path}/${{s%;*}}/*
```
noahdietz, line 55: I don't really know why this was needed in the first place, but it works without it, and actually removing it is necessary because for some reason the `snippets` dir becomes unreadable after this is run...

CONTRIBUTING.md
```diff
@@ -32,12 +32,18 @@ However, this project supports Bazel build environments as well. As such, the Ba
...
+_Note: If running on linux, you must comment remove the `''` value of the `-i` flag in the this
+target's `sed` command._
 
 If there are changes to the `go.mod`, an equivalent change must be made on the `repositories.bzl`
 macro that defines the Go repository dependencies. This can be done manually or by executing the
 Make target: `make update-bazel-repos`.
 
+_Note: If running on linux, you must comment remove the `''` value of the `-i` flag in the this
+target's `sed` command._
+
```
noahdietz, line 46: I know this is silly, but I can't be bothered to make a target that checks the OS and changes how `sed` is invoked.

Makefile
```diff
@@ -30,6 +30,12 @@ update-bazel-repos:
 	bazel run //:gazelle -- update-repos -from_file=showcase/go.mod -to_macro=repositories.bzl%com_googleapis_gapic_generator_go_repositories
 	sed -i ''  "s/    \"go_repository\",//g" repositories.bzl
 
+gazelle:
+	bazel run //:gazelle
+	sed -i '' "s/extendedops_go_proto/extended_operations_go_proto/g" internal/gengapic/BUILD.bazel
+	sed -i '' "s/@com_github_golang_protobuf\/\/protoc-gen-go\/plugin/@io_bazel_rules_go\/\/proto\/wkt:compiler_plugin_go_proto/g" cmd/protoc-gen-go_gapic/BUILD
+	sed -i '' "s/@com_github_golang_protobuf\/\/protoc-gen-go\/plugin/@io_bazel_rules_go\/\/proto\/wkt:compiler_plugin_go_proto/g" cmd/protoc-gen-go_cli/BUILD.bazel
+
```
noahdietz, line 38: So now we don't have to manually revert these changes, and we can more easily add new files/deps to our targets.

internal/gengapic/well_known_types.go
```diff
@@ -0,0 +1,56 @@
...
+var wellKnownTypeFiles = []*descriptor.FileDescriptorProto{
+	protodesc.ToFileDescriptorProto(emptypb.File_google_protobuf_empty_proto),
+	protodesc.ToFileDescriptorProto(anypb.File_google_protobuf_any_proto),
+	protodesc.ToFileDescriptorProto(structpb.File_google_protobuf_struct_proto),
+	protodesc.ToFileDescriptorProto(durationpb.File_google_protobuf_duration_proto),
+	protodesc.ToFileDescriptorProto(timestamppb.File_google_protobuf_timestamp_proto),
+	protodesc.ToFileDescriptorProto(fieldmaskpb.File_google_protobuf_field_mask_proto),
+	protodesc.ToFileDescriptorProto(wrapperspb.File_google_protobuf_wrappers_proto),
+	protodesc.ToFileDescriptorProto(typepb.File_google_protobuf_type_proto),
+}
```
noahdietz, line 56: brute force add all of the `google.protobuf` descriptors to our type registry so that they are resolvable even if not directly imported by a target proto.

quartzmo (approved):

CONTRIBUTING.md
```diff
@@ -32,12 +32,18 @@ However, this project supports Bazel build environments as well. As such, the Ba
 to be kept up-to-date.
 
 When new files are added, or existing ones are deleted or moved, the appropriate BUILD files must
-be updated. This can, and should, be done automatically via Gazelle e.g. `bazel run //:gazelle`.
+be updated. This can, and should, be done automatically via Gazelle e.g. `make gazelle`.
+
+_Note: If running on linux, you must comment remove the `''` value of the `-i` flag in the this
```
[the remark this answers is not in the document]
noahdietz: Done.

internal/gengapic/BUILD.bazel
```diff
@@ -62,11 +71,13 @@ go_test(
         "gengapic_test.go",
         "genrest_test.go",
         "helpers_test.go",
+        "imports_test.go",
         "markdown_test.go",
         "metadata_test.go",
         "mixins_test.go",
         "options_test.go",
         "paging_test.go",
+        "snippets_test.go",
```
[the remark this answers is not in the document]
noahdietz: All good! It wasn't "straight forward" so no problem at all. With `make gazelle`, now it will be!

PR #1292, opened 2023-04-07
fix: add time import for compute operations

codyoss:
I believe without the deadline code time was no longer being included here or something. Tested locally and the target now compiles again with this change

Also removed the compute regen test as compute can no longer be regenerated with the old tooling.


codyoss:
I am not sure if the compute test still provides us value or not. I removed for now because gapicgen can no longer generate it, but I am sure we could make a googleapis style test if needed. I did swap these changes into googleapis though and the gapic target now builds again.

quartzmo (approved):

PR #1310, opened 2023-04-18
fix(internal/gengapic): fix mixin file filter

noahdietz:
Fixes filtering out mixin proto files unless generating a mixin gapic explicitly.

The check is now based on if the Go package path for the GAPIC to be generated includes the `iam` `location` or `longrunning` package in it e.g. `cloud.google.com/go/iam/apiv1`. This ensures we can still generate a GAPIC for those mixin services, but prevents them from being accidentally included in a dependent GAPIC.

Refactors some of this code to make it easier to unit test.


codyoss (approved):
LGTM

PR #1349, opened 2023-05-30
chore: make renovate config closer to other packages

codyoss:
- don't assign reviewer, leave it to CO
- extend the base weekly strat
- group deps bumps together.

Let's try bundling dep updagrades so a single pr can do it all for this repo. It may be we want to explicitly package some seperatly, if so we can carve those out in a future pr.


noahdietz:
Thank yoU!

PR #1358, opened 2023-06-05
fix(gengapic): remove unknown enum error helper

### LINKED ISSUE #1302: gapic: remove maybeUnknownEnum error checker and usage
The `maybeUnknownEnum` helper was meant as a stop gap until `rest-numeric-enum` support was rolled out, which it has been. So this helper should be removed along with the generated usages of it in REGAPIC clients.

noahdietz:
Most if not all REGAPIC clients use numeric enum value serialization now, all new APIs will, and we haven't had an unknown enum value issue reported to date. It is likely safe to remove the error parsing at this point. 

Fixes #1302


codyoss (approved):
Nice!

PR #1359, opened 2023-06-13
fix(gengapic): use gax GoVersion

### LINKED ISSUE #1299: gapic: move versionGo helper into gax-go
The [`versionGo`](https://github.com/googleapis/google-cloud-go/blob/f60bde45662a74b55fd58305891a7fa4e44e4bdd/alloydb/apiv1/doc.go#L158-L191) helper doesn't need to be generated into every single GAPIC's `doc.go` file.

Let's consolidate it into `gax-go`.

noahdietz:
Replace per-package generated `versionGo` helper with use of `gax.GoVersion` in gax/v2 `v2.11.0`. Update deps accordingly.

Fixes #1299


codyoss (approved):
Nice negative commit! We need to make sure we fully roll-out the gax update before we put this into googleapis

noahdietz:
> We need to make sure we fully roll-out the gax update before we put this into googleapis

Yep, totally.

PR #1361, opened 2023-06-18
fix(gengapic): remove quotes from query params for well known types

Charlesleonius:
Update rest calls have quotes around the updateMask query parameter that causes validation issues on the server side. 
Ex:
`updateMask=%22spec.ingressPolicies%2Cspec.egressPolicies%2Ctitle%2CuseExplicitDryRunSpec%22`

### DIFF
--- a/internal/gengapic/genrest.go
+++ b/internal/gengapic/genrest.go
@@ -419,7 +419,11 @@ func (g *generator) generateQueryString(m *descriptor.MethodDescriptorProto) {
 				b.WriteString("  return nil, err\n")
 			}
 			b.WriteString("}\n")
-			b.WriteString(fmt.Sprintf("params.Add(%q, string(%s))", key, field.GetJsonName()))
+			// Convert to string and remove surrounding quotes
+			b.WriteString(fmt.Sprintf("%sString := string(%s)\n", field.GetJsonName(), field.GetJsonName()))
+			b.WriteString(fmt.Sprintf("%sString = %sString[1:len(%sString)-1]\n", field.GetJsonName(), field.GetJsonName(), field.GetJsonName()))
+			// Add field to query parameters
+			b.WriteString(fmt.Sprintf("params.Add(%q, %sString)", key, field.GetJsonName()))
 			paramAdd = b.String()
 		} else {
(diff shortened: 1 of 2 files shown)


noahdietz:
I'm not sure that we should change this because this is the output of pkg.go.dev/google.golang.org/protobuf/encoding/protojson. Example here: https://go.dev/play/p/a1-8k9W27Di

I'm not comfortable with manipulating the output of `protojson` as it is supposed to be "correct". This should probably be raised as an issue in google.golang.org/protobuf, because all clients shouldn't have to implement this behavior. 

Technically, I think it aligns with the [protobuf JSON spec](https://protobuf.dev/programming-guides/proto3/#json), but perhaps that spec is not specific enough. We can keep this open until we have more answers.

If you have more information on how the setup that is having issues with this, that would be useful. If it is Google internal infra, let's take it offline.

noahdietz:
I should add, **thank you** for being willing to contribute in the first place, it really is appreciated.

Charlesleonius:
Thanks for the quick reply @noahdietz! 

I agree that this isn't the ideal solution and even if we were to go this route there should be more specific handling for different types to ensure each one is properly formatted. I created this PR as is in order to see if it would break any integration tests as I was having trouble getting them set up myself. 

The behavior does align with the spec but if the Google APIs do not accept the format as specified then somehow we have to get around it. We can discuss the internals of the service in question offline as you suggested. It may not be the case for all services but the issue has been observed by a user in at least one instance.

noahdietz:
> I created this PR as is in order to see if it would break any integration tests as I was having trouble getting them set up myself.

I've "approved" the tests for execution. Lmk what failed with `make test` - might be a dependency issue (you need a few dependencies installed).

noahdietz:
This is just a play link for the record - testing of quoted strings encoded by `url.Values`: https://go.dev/play/p/I4k_faon8-2

Charlesleonius:
Awesome thanks for approving the tests. As for the url.Values example, since the params are quoted like that even by Go's net library then it seems the API in question should be able to handle it. I am looking into why its the case that it doesn't. We aren't doing any special parsing, its all done by the API platform that provides a rest surface for gRPC services. We can discuss more offline.

noahdietz:
Let's use #1364 instead. Thanks for opening this and finding the issue!

PR #1364, opened 2023-06-20
fix(gengapic): remove encoded quotes in query params

### LINKED ISSUE #1363: regapic: well known types encoded as strings have extra quotations
#1361 points out that some well known types, like a `FieldMask`, are encoded as a string with quotations. These need to have the wrapping quotes removed. The extra quotations encoded break decoding on the server side.

noahdietz:
Removes all quotation marks from `protojson` encoded well known types that are encoded as strings. Lazily use `strings.ReplaceAll` to make it easier when the WKT doesn't necessarily have quotations included.

Showcase integration still works (some other languages do not encode quotations), and I tested locally with another API's samples.

Fixes #1363


noahdietz:
cc: @Charlesleonius

Charlesleonius:
What if there are valid quotes within the marshaled data?

noahdietz:
Some quick testing of different WKT encodings: https://go.dev/play/p/YRah-sNojnC

Number specific wrapper types e.g. `DoubleValue` will be encoded as the raw number, not a string, whereas `BytesValue`, for example, will be a quoted string. A `ListValue` or `Value` of any of the string-encoded wrapper types will also involve quotes but such a value isn't typically used in the query params...updated the solution to handle just the known string encoded types

noahdietz:
> What if there are valid quotes within the marshaled data?

Yeah I've switched back to that more targeted solution to avoid replacing quotes in `StringValue` fields.

codyoss (approved):

PR #1366, opened 2023-06-28
test(showcase): add script to test against local showcase repo

vchudnov-g:
This is useful during Showcase development.


quartzmo:
I get the following error, am I doing something wrong?

```sh
$ ./showcase-local.sh
./showcase-local.sh: line 74: `test-with-local-showcase': not a valid identifier
```

vchudnov-g:
@quartzmo Oh, it's probably that I am running under `zsh`. I'll add a shebang line to the script.

Issue #1301, opened 2023-04-17
gapic: move buildHeaders helper into gax-go


noahdietz:
The [`buildHeaders`](https://github.com/googleapis/google-cloud-go/blob/f60bde45662a74b55fd58305891a7fa4e44e4bdd/alloydb/apiv1/doc.go#L204-L210) helper doesn't need to be generated into every single GAPIC's `doc.go` file.

There could be a transport-agnostic solution in conjunction with #1300

Issue #1300, opened 2023-04-17
gapic: move insertMetadata helper into gax-go


noahdietz:
The [`insertMetadata`](https://github.com/googleapis/google-cloud-go/blob/f60bde45662a74b55fd58305891a7fa4e44e4bdd/alloydb/apiv1/doc.go#L127-L136) helper doesn't need to be generated into every single GAPIC's `doc.go` file.

noahdietz:
Here are the docs for gRPC Metadata (aka headers): https://github.com/grpc/grpc-go/blob/master/Documentation/grpc-metadata.md

PR #1368, opened 2023-07-07
feat(gengapic): use gax.BuildHeaders and gax.InsertMetadataIntoOutgoingContext

quartzmo:
closes: #1300
closes: #1301


noahdietz (commented):
I think you trying to do grpc in one PR and rest in another - i'd just do them both in the same PR

quartzmo:
> I think you trying to do grpc in one PR and rest in another - i'd just do them both in the same PR

I tried, but REST is much more involved. So I opened this PR for gRPC to get feedback. If it looks good, I can add the REST changes, and update the title, etc.

How does this look to you so far?

noahdietz:
Ok let me take a deeper look, I just skimmed it before.

quartzmo:
@noahdietz ping

go.sum
noahdietz, line 1: I would revert this and catch up with `main` it's creating conflicts atm

showcase/go.sum
noahdietz, line 1: I would revert this and catch up with `main` it's creating conflicts atm

noahdietz (commented):
looking OK so far, yeah

showcase/go.sum
quartzmo: Done, thanks!

go.sum
quartzmo: Done, thanks!

internal/gengapic/gengapic.go
```diff
@@ -380,9 +380,12 @@ func (g *generator) insertRequestHeaders(m *descriptor.MethodDescriptorProto, t
 		p("")
 		switch t {
 		case grpc:
-			p("ctx = insertMetadata(ctx, c.xGoogMetadata, md)")
+			p("hds = append(c.xGoogHeaders, hds...)")
+			p("ctx = gax.InsertMetadataIntoOutgoingContext(ctx, hds...)")
 		case rest:
-			p(`headers := buildHeaders(ctx, c.xGoogMetadata, md, metadata.Pairs("Content-Type", "application/json"))`)
+			p(`hds = append(c.xGoogHeaders, hds...)`)
+			p(`hds = append(hds, "Content-Type", "application/json")`)
```
noahdietz, line 387: I kind of wish `"Content-Type", "application/json"` was just part of `c.xGoogHeaders` ya know? It's set on every regapic request...but `xGoogHeaders` has always been `x-goog-api-client` headers only...idk wdyt? @codyoss hbu?

noahdietz (commented):
LGTM just one question that isn't really blocking

internal/gengapic/gengapic.go
quartzmo: I'll wait to hear back from Cody on this before merging.

codyoss: I would be fine with it just being agnostic and just general headers.

tritone: I would imagine `XGoogHeaders` is specifically for headers with the x-goog prefix (of which there are many, not just `x-goog-api-client`)

quartzmo: @noahdietz What's your call on this?

noahdietz: Let's just keep it as is.

noahdietz (approved):

PR #1379, opened 2023-07-20
chore(ci): split out bazel go dep updates

noahdietz:


codyoss (approved):
I think this is right, I do wonder if we need a different go modules grouping for if that will still come in a seperate deps PR, but lets try this!

PR #1397, opened 2023-09-21
fix(internal/gengapic): add workaround for operation collision

codyoss:
We have a service that is not yet released that uses the same RPC name across two services in the same proto package. Our Operation wrapper identifiers are not scoped to a service so this causes issues when generated a client that does this. For now this is for one client, which may change before GA, so a hacky solution is fine. But we should keep this in the back of our minds should we ever v2 our libraries this is something we would want to avoid.

### DIFF
--- a/internal/gengapic/lro.go
+++ b/internal/gengapic/lro.go
@@ -21,6 +21,7 @@ import (
 	longrunning "cloud.google.com/go/longrunning/autogen/longrunningpb"
 	"github.com/golang/protobuf/protoc-gen-go/descriptor"
 	"github.com/googleapis/gapic-generator-go/internal/pbinfo"
+	"google.golang.org/genproto/googleapis/api/annotations"
 	"google.golang.org/protobuf/proto"
 )
 
@@ -38,7 +39,7 @@ func (g *generator) lroCall(servName string, m *descriptor.MethodDescriptorProto
 		return err
 	}
 
-	lroType := lroTypeName(m.GetName())
+	lroType := lroTypeName(m)
 	p := g.printf
 
 	lowcaseServName := lowerFirst(servName + "GRPCClient")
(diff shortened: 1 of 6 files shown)


noahdietz (approved):

internal/gengapic/lro.go
```diff
@@ -280,6 +281,17 @@ func (g *generator) lroType(servName string, serv *descriptor.ServiceDescriptorP
 	return nil
 }
 
-func lroTypeName(methodName string) string {
-	return methodName + "Operation"
+func lroTypeName(m *descriptor.MethodDescriptorProto) string {
+	// This whole if block is a hack to workaround a operation handler namespace
+	// collision. We should remove this in the future if the design is fixed for
+	// the v1 api.
```
quartzmo, line 287: Can you add a link or the full name/path of the problematic API?

internal/gengapic/lro.go
```diff
@@ -280,6 +281,17 @@ func (g *generator) lroType(servName string, serv *descriptor.ServiceDescriptorP
...
-func lroTypeName(methodName string) string {
-	return methodName + "Operation"
+func lroTypeName(m *descriptor.MethodDescriptorProto) string {
+	// This whole if block is a hack to workaround a operation handler namespace
+	// collision. We should remove this in the future if the design is fixed for
+	// the v1 api. This is for aiplatform.featureregistryservice.createfeature.
+	if eHTTP, ok := proto.GetExtension(m.GetOptions(), annotations.E_Http).(*annotations.HttpRule); ok && eHTTP != nil && eHTTP.Pattern != nil {
+		switch t := eHTTP.Pattern.(type) {
+		case *annotations.HttpRule_Post:
+			if t.Post == "/v1beta1/{parent=projects/*/locations/*/featureGroups/*}/features" {
```
quartzmo, line 291: It seems at least a tiny bit possible that someday another RPC will match this pattern...

quartzmo (approved):

internal/gengapic/lro.go
codyoss: It may, but I believe it is unlikely. Ideally this code gets deleted in the next couple of months. If we end up having to do this for the v1 service we can make the logic tighter at that time. I was trying to avoid passing a lot more info throughout the stack if possible so this is easy to unwind.

PR #1406, opened 2023-10-05
chore( internal/txtdiff): replace lclDiff with cmp.Diff

julieqiu:
Instead of writing a custom diff function (`lclDiff`), use `cmp.Diff` instead. github.com/google/go-cmp is already used in other parts of this module.


noahdietz (approved):

noahdietz:
Nice! Love some good code cleanup

PR #1419, opened 2023-10-06
chore(internal/errors): replace errors.E with fmt.Errorf

julieqiu:
Replace uses of `errors.E` with `fmt.Errorf` and delete the internal/errors package. 

At the moment, the majority of use cases for `errors.E` passes in `nil` as the first arugment, so the output is just a normal error string. In the 3 cases where an actual error is passed in, it is now formatted accordingly using `fmt.Errorf`. This removes an extra level of indirection.


noahdietz (approved):
Love it

PR #1423, opened 2023-10-09
chore(.github/workflows): add staticcheck as github action

julieqiu:
[staticcheck](https://staticcheck.io) is a common Go linter for finding bugs and performance issues, providing simplifications, and enforcing Go style rules.

 Now that all the staticcheck errors have been fixed, add the GitHub action to check for future errors.


noahdietz (commented):
Is this ready for review @julieqiu ? Looks unfinished...I think staticcheck needs to be invoked at some point? I imagine it would replace all of our golint and vet call sites?

julieqiu:
@noahdietz - I think it should be running as part of `Generator tests / lint (pull_request)`, step 6:
https://github.com/googleapis/gapic-generator-go/actions/runs/6461386147/job/17540942152?pr=1423

I also tested this on https://github.com/googleapis/gapic-generator-go/pull/1424 by mailing a staticcheck error and it fails in an expected manner:
https://github.com/googleapis/gapic-generator-go/actions/runs/6473583416/job/17576606869?pr=1424

With regard to `staticcheck` and `go vet`, it looks like they run separate checks:
https://github.com/dominikh/go-tools/issues/149

I'm not as familiar with `golint`, but I can ask around and send a PR in the future if it isn't useful anymore.

noahdietz:
> I also tested this on #1424 by mailing a staticcheck error and it fails in an expected manner

Looks like it fails because it runs and found an issue with `string.Title` being used. So it works!

Screenshot 2023-10-10 at 3 05 10 PM

> With regard to staticcheck and go vet, it looks like they run separate checks:

Ok, that's fine! I wasn't sure, just assumed. The docs indicate the same.

noahdietz (approved):

.github/workflows/ci.yaml
```diff
@@ -20,6 +20,9 @@ jobs:
     - name: Install golint
       run: go install golang.org/x/lint/golint@latest
       if: ${{ always() }}
+    - uses: dominikh/staticcheck-action@v1.3.0
+      with:
+        version: "2022.1.3"
```
noahdietz, line 25: Should we just exclude this and use latest?

julieqiu: that works, updated

Issue #1399, opened 2023-09-22
future: have generic handler or service namespaces operation wrappers


codyoss:
Although this is unlikely as we have seen in some recent PRs collisions can happen. We should design around this in the future if we are going to take a v2. ExampleL: #1398

noahdietz:
As way of helping improve the debugging/error reporting and reduce issues where there might not be one, we can actually share operation wrapper types when the following things are true between two RPCs with the same name:
* the RPC simple names are the same e.g. `CreateFoo`
* the `response_type` is the same
* the `metadata_type` is the same

This will make the generation logic work for situations like #1398 while also enabling the generator to report specific collisions in the even that some, but not all of those conditions are true such that they result in a collision.

PR #1428, opened 2023-10-12
feat(internal/gengapic): move operations & iterators to aux.go

### LINKED ISSUE #156: gapic: generate auxiliary types in a separate file
It is possible to have naming collisions in the generated auxiliary types. While generated the Ads v2 API this occurred with an Iterator type. I've fixed it with #155 .

However, this is still possible for LROs given that the name of the generated type is based only on the Method's name. If two services in the same package share a method name, there will be a collision.

I think that the fix in #155 makes this tool more complicated, in order to prevent disturbing the generated output. Eventually, I think it would be better to generate all of the auxiliary types for a **package** in a single file, thus avoiding collisions entirely and simplifying the deduping code. The new file could possibly be named `aux.go`.

noahdietz:
Generate all `Iterator` and `Operation` wrapper types into a single file - `aux.go`. The auxiliary types can be shared across the entire client package when certain criteria are met.

For `Iterator` wrappers, it is simply a matching repeated type, common when multiple RPCs paginate over the same message.

For `Operation` wrappers, the RPC name must be the same, as well as the response and metadata types declared inside of the operation returned. This is less common, but can happen if two proto services in the same package share an RPC name and types. This introduces a new generator error mode when the aforementioned conditions are not met. This is good because previously the generator would succeed, but the generated code would not compile (colliding type names) and not provide a clear reason to debug.

The operation "builder" methods that are generated onto a service client type surface remain in the same place - in the service client file. These are specific to each client, even if they are mostly the same in surface between clients, so we keep them generated in the client file.

An example of the changes can be found in #8703. Building all Go gapic targets under google/cloud worked.

Note: Thus undoes #1398, which will result in a breaking change in the beta client it address a collision in. This is OK because the client is marked as beta/unstable, and it is more true to the surface.

Fixes #156.
Updates #1399.


internal/gengapic/auxiliary.go
```diff
@@ -0,0 +1,365 @@
...
+	"github.com/googleapis/gapic-generator-go/internal/pbinfo"
+	"google.golang.org/protobuf/proto"
+	"google.golang.org/protobuf/reflect/protoreflect"
+	"google.golang.org/protobuf/types/descriptorpb"
+)
+
+// auxTypes gathers details of types we need to generate along with the client
+type auxTypes struct {
+	// Map of RPC descriptor to assigned operationWrapper.
+	methodToWrapper map[*descriptorpb.MethodDescriptorProto]operationWrapper
```
codyoss, line 33: Someday we should create a lightweight generic sorted map impl. We always be sorting in generator because of key randomness today.

internal/gengapic/auxiliary.go
```diff
@@ -0,0 +1,365 @@
...
+
+	// RPC-specific operation wrapper types by wrapper name e.g. CreateFooOperation.
+	opWrappers map[string]operationWrapper
+
+	// "List" of iterator types. We use these to generate FooIterator returned by paging methods.
+	// Since multiple methods can page over the same type, we dedupe by the name of the iterator,
+	// which is in turn determined by the element type name.
+	iters map[string]*iterType
+
+	customOp *customOp
```
codyoss, line 43: nit: rest of the fields have docs, could you add a note here as well.

internal/gengapic/auxiliary_test.go
```diff
@@ -0,0 +1,447 @@
...
+		},
+	}
+	for _, tC := range testCases {
+		t.Run(tC.desc, func(t *testing.T) {
+			a := auxTypes{
+				opWrappers: tC.existing,
+			}
+			got, gotErr := a.wrapperExists(tC.target)
+			if tC.want != got {
+				t.Errorf("wrapperExists(%+v): expected %v, got %v, with existing: %+v", tC.target, tC.want, got, tC.existing)
```
codyoss, line 109: nit: unless this is already a pattern in this project can we switch this to be in the style of got want instead of want got.

codyoss (approved):
LGTM, just a couple nits

internal/gengapic/auxiliary.go
noahdietz: Oo that would be great, and fun :)

If you want to, we could start one in this repo's internal package, then potentially give it its own repo? Would be good practice with generics...

internal/gengapic/auxiliary.go
noahdietz: Sure thing, done.

internal/gengapic/auxiliary_test.go
noahdietz: You bet! Done.

internal/gengapic/auxiliary_test.go
```diff
@@ -0,0 +1,447 @@
...
+				t.Errorf("maybeAddOperationWrapper(%+v): got %q, want %q, with existing %+v", m, err.Error(), tC.wantErrStr, tC.existing)
+			} else if diff := cmp.Diff(g.aux.opWrappers, tC.want, cmp.Comparer(proto.Equal), cmp.AllowUnexported(operationWrapper{})); diff != "" {
+				t.Errorf("maybeAddOperationWrapper(%+v): got(-),want(+)\n%s", m, diff)
+			}
+		})
+	}
+}
+
+func TestGenOperations(t *testing.T) {
+	wrappers := map[string]operationWrapper{
```
quartzmo, line 347: Is it possible to test the sorting in `genOperations`? Would it work to reverse the order of `wrappers` here?

quartzmo (approved):
The comments in this PR are great. They made it really easy to understand.

Also: Bravo on closing an issue from 2019! 🎉 

I had one question about testing, super minor.

internal/gengapic/auxiliary_test.go
noahdietz: Did you one better, I moved the sorting code into their own functions and added tests.

quartzmo (approved):

noahdietz:
Ready to release this now. merging.

PR #1436, opened 2023-11-02
fix(gengapic): fix generator reset ordering

noahdietz:


quartzmo (approved):
Thank you for the quick fix! 🎉

PR #1452, opened 2024-01-19
feat(gengapic): add universe domain support

quartzmo:
* upgrade showcase google.golang.org/api to v0.157.0


quartzmo:
Similar to https://github.com/googleapis/google-api-go-client/pull/2335

internal/gengapic/gengrpc.go
```diff
@@ -177,7 +177,9 @@ func (g *generator) grpcClientOptions(serv *descriptor.ServiceDescriptorProto, s
 	p("func default%sGRPCClientOptions() []option.ClientOption {", servName)
 	p("  return []option.ClientOption{")
 	p("    internaloption.WithDefaultEndpoint(%q),", host)
+	p("    internaloption.WithDefaultEndpointTemplate(%q),", generateDefaultEndpointTemplate(host))
 	p("    internaloption.WithDefaultMTLSEndpoint(%q),", generateDefaultMTLSEndpoint(host))
+	p("    internaloption.WithDefaultUniverseDomain(%q),", googleDefaultUniverse)
```
noahdietz, line 182: Without knowing much how this will ultimately work, do the `WithDefaultEndpoint` and `WithDefaultMTLSEndpoint` options go away assuming endpoint resolution is done by the middleware?

quartzmo: `WithDefaultEndpoint` is already deprecated, and will eventually go away post-transition. `WithDefaultMTLSEndpoint` is unaffected by these changes as that auth flow does not support universe domain.

noahdietz (approved):
Last question: Are all of the go.mod files updated with the required version of google.golang.org/api to leverage these options?

That must happen first before releasing this code.

quartzmo:
Thanks, yes the `go.mod` files in GAPIC clients are currently at the latest version of  `google.golang.org/api` (`v0.157.0`).

For example: https://github.com/googleapis/google-cloud-go/blob/main/accessapproval/go.mod#L7

PR #1460, opened 2024-02-14
feat(gengapic): add support for AutoPopulatedFields UUID4

quartzmo:


internal/gengapic/gengapic.go
```diff
@@ -348,6 +348,30 @@ func (g *generator) insertRequestHeaders(m *descriptor.MethodDescriptorProto, t
 	}
 }
 
+// insertAutoPopulatedFields generates the conditional initialization of any
+// default-value request fields (for the given method) that are specified for
+// auto-population by autoPopulatedFields.
+//
+// If the field value is not equal to default value at the time of sending the
+// request, implying it was set by the user, or if the field has explicit
+// presence and is set by the user, the field must not be auto-populated by
```
quartzmo, line 357: Is there currently any way with Go protos to determine explicit presence?

noahdietz: The `FieldDescriptorProto` should have an api for that `GetProto3Optional`

https://pkg.go.dev/google.golang.org/protobuf@v1.32.0/types/descriptorpb#FieldDescriptorProto.GetProto3Optional

noahdietz: If `GetProto3Optional()` is `true`, then the field will be a pointer e.g. `*string` in the Go code, so you can generate some code like `if req.RequestId == nil { /*autopopulate*/ }`

yes I know that code uses a direct field access instead of a getter, but this is the only way to check presence with protobuf-go open structs :)

quartzmo: Thank you, will do!

noahdietz: Looks like I added a helper for this: https://github.com/googleapis/gapic-generator-go/blob/e446700543fd12d2c3c479263c532f7e4bbeee0a/internal/gengapic/helpers.go#L84-L94

quartzmo: Updated to use `req.RequestId == nil`, can you please verify that this looks correct in this func and in the `want` files below?

quartzmo: OK, I won't rely on the field simply not being REQUIRED, I'll add the `isOptional` check as well, thanks.

noahdietz: > OK, I won't rely on the field simply not being REQUIRED, I'll add the `isOptional` check as well, thanks.

`google.api.field_behavior` doesn't actually indicate anything about field presence - `proto3` using the label `optional` for field presence was kind of awkward. So IDK if the `REQUIRED` check is even necessary unless the design states as much.

noahdietz: Oh actually the label `required` (which you are checking) is only for `proto2`. Definitely don't need that

internal/gengapic/generator.go
```diff
@@ -302,3 +304,44 @@ func (g *generator) nestedName(nested pbinfo.ProtoType) string {
...
+			if f.GetName() == apf {
+				field = f
+				break
+			}
+		}
+		// Do nothing and continue iterating unless all conditions above are met.
+		switch {
+		case field == nil:
+		case field.GetType() != fieldTypeString:
+		case field.GetLabel() == descriptor.FieldDescriptorProto_LABEL_REQUIRED:
```
noahdietz, line 340: Taking the other thread here, if you need to check field behavior, you need to check the option, not the label - i know it's confusing.

So use the helper `isRequired` https://github.com/googleapis/gapic-generator-go/blob/2f1a8d3c867ec551e5a8bd1ce51e181ed4546319/internal/gengapic/helpers.go#L185-L200

`case isRequired(field):`

noahdietz (commented):
Since we are adding a net new dependency to generated code, can you also include the dep for bazel here: https://github.com/googleapis/gapic-generator-go/blob/main/rules_go_gapic/go_gapic.bzl#L156-L157

`"@com_github_google_uuid//:go_default_library"` I think should do it.

internal/gengapic/generator.go
```diff
@@ -302,3 +304,44 @@ func (g *generator) nestedName(nested pbinfo.ProtoType) string {
...
+	var validated []string
+	for _, apf := range apfs {
+		var field *descriptorpb.FieldDescriptorProto
+		// Find the input's FieldDescriptorProto by AutoPopulatedField name.
+		for _, f := range inType.GetField() {
+			if f.GetName() == apf {
+				field = f
+				break
+			}
+		}
```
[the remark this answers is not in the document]
quartzmo: Great, will do.

internal/gengapic/generator.go
quartzmo: Wow, yes, confusing. Will do, thanks!

internal/gengapic/generator.go
```diff
@@ -302,3 +304,44 @@ func (g *generator) nestedName(nested pbinfo.ProtoType) string {
...
+		// Do nothing and continue iterating unless all conditions above are met.
+		switch {
+		case field == nil:
+		case field.GetType() != fieldTypeString:
+		case field.GetLabel() == descriptor.FieldDescriptorProto_LABEL_REQUIRED:
+		case field.GetOptions() == nil:
```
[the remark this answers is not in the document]
quartzmo: Thanks, will remove.

internal/gengapic/gengapic.go
```diff
@@ -348,6 +348,30 @@ func (g *generator) insertRequestHeaders(m *descriptor.MethodDescriptorProto, t
...
+func (g *generator) initializeAutoPopulatedFields(servName string, m *descriptor.MethodDescriptorProto) {
+	p := g.printf
+	apfs := g.autoPopulatedFields(servName, m)
+	if len(apfs) == 0 {
+		return
+	}
+	g.imports[pbinfo.ImportSpec{Path: "github.com/google/uuid"}] = true
+	for _, apf := range apfs {
+		apf = snakeToCamel(apf)
+		p("if req != nil && req.%s == nil {", apf)
```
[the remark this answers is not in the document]
noahdietz: You can use these helpers to simplify some of the code here (i.e. remove the `snakeToCamel` call): https://github.com/googleapis/gapic-generator-go/blob/2f1a8d3c867ec551e5a8bd1ce51e181ed4546319/internal/gengapic/gengapic.go#L372-L392

These should really be in `helpers.go` so that they are easier to find.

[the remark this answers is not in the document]
quartzmo: Will refactor to the former, thanks!

internal/gengapic/gengapic_test.go
```diff
@@ -183,6 +187,12 @@ func TestGenGRPCMethods(t *testing.T) {
 				Type:     typep(descriptor.FieldDescriptorProto_TYPE_ENUM),
 				TypeName: proto.String(".my.pkg.InputType.NestedEnum"),
 			},
+			{
+				Name:    proto.String("request_id"),
+				Type:    typep(descriptor.FieldDescriptorProto_TYPE_STRING),
+				Label:   labelp(descriptor.FieldDescriptorProto_LABEL_OPTIONAL),
```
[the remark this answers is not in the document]
quartzmo: `Label:   labelp(descriptor.FieldDescriptorProto_LABEL_OPTIONAL),` appears in 17 locations across 5 test files, should I change them all?

internal/gengapic/gengapic.go
```diff
@@ -348,6 +348,30 @@ func (g *generator) insertRequestHeaders(m *descriptor.MethodDescriptorProto, t
...
+		p("  req.%s = uuid.New().String()", apf)
```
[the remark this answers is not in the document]
quartzmo: Fixed, thanks!

internal/gengapic/gengapic.go
quartzmo: Fixed, thanks!

internal/gengapic/generator.go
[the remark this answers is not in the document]
quartzmo: Done, thanks!

internal/gengapic/generator.go
quartzmo: Fixed, thanks!

internal/gengapic/generator.go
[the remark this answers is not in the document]
quartzmo: Removed, thanks!

internal/gengapic/gengapic_test.go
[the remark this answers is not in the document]
quartzmo: Updated only this usage, by adding `Proto3Optional: proto.Bool(true),`.

quartzmo:
@noahdietz Updated with the UUID dependency, thank you.

I locally ran the bazel targets for the "usual suspects" set of gapic libraries:

```
bazelisk build //google/cloud/apigeeconnect/v1:apigeeconnect_go_gapic
bazelisk build //google/cloud/asset/v1:asset_go_gapic
bazelisk build //google/cloud/language/v1:language_go_gapic
bazelisk build //google/spanner/v1:spanner_go_gapic
bazelisk build //google/cloud/aiplatform/v1beta1:aiplatform_go_gapic
bazelisk build //google/cloud/compute/v1:compute_go_gapic
```

internal/gengapic/generator.go
```diff
@@ -302,3 +303,36 @@ func (g *generator) nestedName(nested pbinfo.ProtoType) string {
...
+//
+// * The field is a top-level string field of a unary method's request message.
+// * The field is not annotated with google.api.field_behavior = REQUIRED.
+// * The field name is listed in google.api.publishing.method_settings.auto_populated_fields.
+// * The field is annotated with google.api.field_info.format = UUID4.
+func (g *generator) autoPopulatedFields(servName string, m *descriptor.MethodDescriptorProto) []*descriptor.FieldDescriptorProto {
+	var apfs []string
+	// Find the service config's AutoPopulatedFields entry by method name.
+	for _, s := range g.serviceConfig.GetPublishing().GetMethodSettings() {
+		if s.GetSelector() == g.fqn(m) {
```
noahdietz, line 319:
```suggestion
	mfqn := g.fqn(m)
	for _, s := range g.serviceConfig.GetPublishing().GetMethodSettings() {
		if s.GetSelector() == mfqn {
```

internal/gengapic/generator.go
```diff
@@ -302,3 +303,36 @@ func (g *generator) nestedName(nested pbinfo.ProtoType) string {
...
+			apfs = s.AutoPopulatedFields
+			break
+		}
+	}
+	inType := g.descInfo.Type[*m.InputType].(*descriptor.DescriptorProto)
```
noahdietz, line 324:
```suggestion
	inType := g.descInfo.Type[m.GetInputType()].(*descriptor.DescriptorProto)
```
Does this work? we should avoid using direct access/dereferencing b.c they cause `panic`

internal/gengapic/generator_test.go
```diff
@@ -0,0 +1,141 @@
...
+	got := g.autoPopulatedFields(serv.GetName(), m)
+	if want := 1; len(got) != want {
+		t.Errorf("len(got) = %d, want: %d, got: %v", len(got), want, got)
+	}
+	if want := "request_id"; got[0].GetName() != want {
+		t.Errorf("got[0].GetName() = %s, want: %s", got[0].GetName(), want)
+	}
+	if !got[0].GetProto3Optional() {
+		t.Error("got[0].GetProto3Optional() = false, want: true")
+	}
```
noahdietz, line 140: What's the goal of this assertion?

internal/gengapic/generator_test.go
```diff
@@ -0,0 +1,141 @@
...
+	g.descInfo.ParentElement[m] = serv
+	serv.Method = []*descriptor.MethodDescriptorProto{m}
+
+	got := g.autoPopulatedFields(serv.GetName(), m)
+	if want := 1; len(got) != want {
+		t.Errorf("len(got) = %d, want: %d, got: %v", len(got), want, got)
+	}
+	if want := "request_id"; got[0].GetName() != want {
+		t.Errorf("got[0].GetName() = %s, want: %s", got[0].GetName(), want)
+	}
```
noahdietz, line 137: This is not a required change. But another way to do this while being a little more future proof/exact is as follows:

```go
if diff := cmp.Diff(got, want, protocmp.Transform()); diff != "" {
  t.Errorf(...)
}
```

Where [`protocmp.Transform`](https://pkg.go.dev/google.golang.org/protobuf/testing/protocmp#Transform) is a protobuf provided `cmp.Option` comparing protobuf messages

internal/gengapic/gengapic.go
```diff
@@ -348,6 +348,36 @@ func (g *generator) insertRequestHeaders(m *descriptor.MethodDescriptorProto, t
...
+	g.imports[pbinfo.ImportSpec{Path: "github.com/google/uuid"}] = true
+	p := g.printf
+	for _, apf := range apfs {
+		f := snakeToCamel(apf.GetName())
+		if apf.GetProto3Optional() {
+			// Type will be *string if field has explicit presence.
+			p("if req != nil && req.%s == nil {", f)
+		} else {
+			// Type will be string if field does not have explicit presence.
+			p("if req != nil && req.%s == \"\" {", f)
```
noahdietz, line 374:
```suggestion
			p("if req != nil && req.Get%s() == \"\" {", f)
```
But we may want to use `buildAccessor` to get this string instead of just doing snakeToCmael...

noahdietz, line 374:
```suggestion
			p(`if req != nil && req.%s == "" {`, f)
```

internal/gengapic/gengapic.go
```diff
@@ -348,6 +348,36 @@ func (g *generator) insertRequestHeaders(m *descriptor.MethodDescriptorProto, t
...
+		}
+		p("  req.%s = uuid.NewString()", f)
```
noahdietz, line 376: If it is `Proto3Optional` you will need to make this a pointer value e.g. 
`req.%s = proto.String(uuid.NewString())` via the [`proto.String` helper](https://pkg.go.dev/google.golang.org/protobuf@v1.32.0/proto#String)

internal/gengapic/generator_test.go
quartzmo: I tried using `cmp` first, but it turned out ugly due to unexported fields, so I switched to this.

internal/gengapic/generator_test.go
```diff
@@ -0,0 +1,141 @@
...
+				Name:           proto.String("request_id"),
+				Type:           typep(descriptor.FieldDescriptorProto_TYPE_STRING),
+				Label:          labelp(descriptor.FieldDescriptorProto_LABEL_OPTIONAL),
+				Proto3Optional: proto.Bool(true),
+				Options:        optsUUID4,
+			},
+			{
+				Name:  proto.String("invalid_auto_populated_not_in_serviceconfig"),
+				Type:  typep(descriptor.FieldDescriptorProto_TYPE_STRING),
+				Label: labelp(descriptor.FieldDescriptorProto_LABEL_OPTIONAL),
```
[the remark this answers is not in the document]
quartzmo: No idea, but existing fixtures have it, so I'm just conforming to the existing usage.

internal/gengapic/gengapic.go
quartzmo: I get:

```
gen/github.com/googleapis/gapic-showcase/client/echo_client.go:682:19: cannot use proto.String(uuid.NewString()) (value of type *string) as string value in assignment
```
Any suggestions?

internal/gengapic/generator_test.go
noahdietz: Did you use the option I mentioned above? It fixes issues with unexported comparisons.

internal/gengapic/gengapic.go
noahdietz: > If it is `Proto3Optional` you will need to make this a pointer value

Only if it's a `Proto3Optional` field :)

internal/gengapic/generator.go
quartzmo: Fixed, thanks!

internal/gengapic/generator_test.go
quartzmo: Just a dev-time sanity check, deleted, thanks!

internal/gengapic/gengapic.go
quartzmo: Fixed, thanks!

internal/gengapic/gengapic.go
quartzmo: Fixed, thanks!

internal/gengapic/gengapic.go
quartzmo: Please check the update `.want` files to verify the output now looks good, thanks!

internal/gengapic/generator.go
```diff
@@ -316,12 +316,13 @@ func (g *generator) autoPopulatedFields(servName string, m *descriptor.MethodDes
 	var apfs []string
 	// Find the service config's AutoPopulatedFields entry by method name.
 	for _, s := range g.serviceConfig.GetPublishing().GetMethodSettings() {
-		if s.GetSelector() == g.fqn(m) {
+		mfqn := g.fqn(m)
```
noahdietz, line 319: I should've been more clear in my comment. Moving the `g.fqn(m)` call outside of the loop b.c it doesn't change inside of the loop, so calling it there repeatedly is a waste of cycles.

internal/gengapic/gengapic.go
```diff
@@ -365,15 +365,17 @@ func (g *generator) initializeAutoPopulatedFields(servName string, m *descriptor
 	g.imports[pbinfo.ImportSpec{Path: "github.com/google/uuid"}] = true
 	p := g.printf
 	for _, apf := range apfs {
-		f := snakeToCamel(apf.GetName())
+		getF := buildAccessor(apf.GetName(), false)
+		f := buildAccessor(apf.GetName(), true)
 		if apf.GetProto3Optional() {
 			// Type will be *string if field has explicit presence.
-			p("if req != nil && req.%s == nil {", f)
+			p("if req != nil && req%s == nil {", getF)
```
noahdietz, line 372: Again, I should've been more clear, I'm sorry. With `Proto3Optional` fields, the only way to check for presence is to use direct access e.g. `req.Foo == nil`. Using the accessor will return the default empty value e.g. `""` but won't convey if it was explicitly set to that.

So in the proto3optional branch, we need to check presence with direct access. But with non-proto3optional, we want to use the accessor.

noahdietz (commented):
I'm sorry @quartzmo I should've been more explicit in my comments

internal/gengapic/gengapic.go
quartzmo: Thanks that's clear now, updated!

internal/gengapic/generator.go
quartzmo: Actually I understood perfectly the first time, but I simply put the new variable on the wrong line! Ugh!

internal/gengapic/generator_test.go
quartzmo: Fixed, thanks for the tip!

noahdietz (approved):
LGTM great work and thanks for your patience with my review. Add that test we spoke about offline and you are good to go.

PR #1468, opened 2024-03-07
fix(internal/gengapic): add iter response access example

### LINKED ISSUE #1467: Make getting raw responses from iterators easier
Recently we have had a couple of issues opened around extracting such values. Our current iterator design has us generating an iterator type for each repeated field that automatic pagination for an RPC operates over: https://google.aip.dev/client-libraries/4233. This means that two different RPC could return the same iterator type. This is nice in the fact that we are generating less pagination helper code, but can makes it harder for users that want to extract raw responses from the RPC due to things like extra metadata that are returned from a List RPC. This means the the Response field of every iterator is an `interface{}/any`. Unless a user digs into the code a bit it is not always clear what the type could be. We should find a way to make this experience better in the future and/or generate some extra docs/samples somewhere around how and what type to extract the Response two.

Context:
- https://github.com/googleapis/google-cloud-go/issues/9487
- https://github.com/googleapis/google-cloud-go/issues/9092

noahdietz:
Adds some code to the paging call example that demonstrates accessing the Iterator's underlying RPC specific response type for a page. Since Iterators are generated per-resource, rather than per RPC, the `Response` field has to be generic, but this makes it harder for users to get at the underlying response message if there are other interesting fields.

Fixes #1467

### DIFF
--- a/internal/gengapic/example.go
+++ b/internal/gengapic/example.go
@@ -147,7 +147,9 @@ func (g *generator) exampleMethodBody(pkgName, servName string, m *descriptor.Me
 		return err
 	}
 	if pf != nil {
-		g.examplePagingCall(m)
+		if err := g.examplePagingCall(m); err != nil {
+			return err
+		}
 	} else if g.isLRO(m) || g.isCustomOp(m, httpInfo) {
 		g.exampleLROCall(m)
 	} else if *m.OutputType == emptyType {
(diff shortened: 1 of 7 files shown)


internal/gengapic/example.go
```diff
@@ -225,9 +237,16 @@ func (g *generator) examplePagingCall(m *descriptor.MethodDescriptorProto) {
 	p("  }")
 	p("  // TODO: Use resp.")
 	p("  _ = resp")
+	p("")
+	p("  // TODO: Use the underlying response message.")
```
codyoss, line 241: I assume most people will never have to actually do this, I wonder if we should use more cautious language to imply they usually don't need this. Something like "If you need to access the underlying request you can do so by casting the Response as follows ". Any maybe even keep that line completely commented out. Thoughts?

noahdietz: Sounds good to me. I agree, I think we should warn folks off a bit.

noahdietz: I don't want to comment it out though b.c I want the import resolution for the reference.

noahdietz: Updated. WDYT?

codyoss (approved):

quartzmo (commented):
Great idea!

PR #1485, opened 2024-04-09
chore(ci): use bazelisk

noahdietz:
Bazelisk is the preferred "runner" of bazel and more closely aligns with tools we use in googleapis.


codyoss (approved):
Nice!

PR #1486, opened 2024-04-09
fix(bazel): upgrade rules_go & gazelle to drop go_googleapis shading

noahdietz:
`rules_go` and `gazelle` dropped the `go_googleapis` dependency [that we had to shade](https://github.com/googleapis/gapic-generator-go/commit/1dcdb84e79e8dac82a8c7da95e6b0e4c0ec5bc38) with our own copy of googleapis  here and also in googleapis/WORKSPACE itself.

Using gazelle resolve directives, we can instead resolve the various proto imports we have in this project to the googleapis sources, and all of the common proto Go bindings to the genproto modules.

In the future I'd like to be able to move away from the genproto module as a dependency of the generator in Bazel, instead resolving the imports to the googleapis `go_proto_library` targets, but I'm not sure if that is feasible atm. Using genproto modules *may* end up being a source of import path linker collisions in the future, but for now it works...

...I know it works because I tested this locally in googleapis, building all `go_gapic_library` targets using my local branch of the generator! While testing this I had to add some `build_directives` to a few `go_repository` entries. I don't know why this was necessary when building within googleapis itself, but documentation (and recent experience with doing the same thing the otel libraries) indicates it's just necessary.

I was running into bazel go issues in #1479 so I'm hoping that doing this now helps make that easier as well.

Note: In order to integrate this into googleapis, we will need to update `rules_go` and `gazelle` in the same CL there as well. This is what was necessary to get the `go_gapic_library` targets built locally.

A complete list of changes:
* upgrade rules_go to 0.46.0
* upgrade gazelle to v0.36.0
* rename `go_googleapis` shader to normal `com_google_googleapis` (still need it for some proto imports)
* remove `sed` commands in favor of `gazelle:resolve` directives
* run `make gazelle` to update all BUILD.bazel files
* add `build_directives` to `go_repositories` that are necessary to build gapics in googleapis

Supercedes #1457.


noahdietz:
Updates:
* I rebased my protobuf-go migration PR onto this one and it *just worked*! https://github.com/googleapis/gapic-generator-go/pull/1487
* I ran googleapis presubmits with this branch and it worked (pending a weird network flake in the stable dir) cl/623329761

codyoss (approved):

PR #1489, opened 2024-04-10
fix: refactor codebase to protobuf-go v2

noahdietz:
A number of things are done to the project, all in the name of moving away from the deprecated protobuf-go v1 module.

* We've replaced every usage of the protobuf-go v1 with the v2 drop-in equivalent
  * This includes both code generators, gapic and cli, as well as supporting libraries
* Regenerated the grpc service config Go proto with the protobuf-go v2 code generator
* Update all BUILD.bazel files with appropriate Go dependency changes
* Remove protobuf-go v1 from `go.mod`/`go.sum`
* Updated `repositories.bzl` based on `go.mod`/`go.sum` changes
* Switch `showcase.bash` to use the individual protobuf-go v2 and grpc code gen plugins
* Switch/add installs for protobuf-go v2 and grpc code gen plugins in CI workflow


codyoss (approved):
Thank you for this!

PR #1503, opened 2024-05-03
fix(gencli): don't generate grpc.WithInsecure

egonelbre:
grpc.WithInsecure is deprecated as per https://pkg.go.dev/google.golang.org/grpc#WithInsecure


noahdietz (dismissed):
Thanks! This is in `gencli` not `gengapic`, but I've changed the title, so automerge will use the proper commit

quartzmo (approved):

PR #1508, opened 2024-05-11
chore(internal/testing/sample): add package

julieqiu:
Package sample is added, which provides sample values for use in tests.

The intent is to make it easier to reason about the various test cases,
by using one example and explaining where various values are defined,
using bigquery as an example.

As a starting point, TestGenSnippetFile has been refactor to use the
constants in internal/testing/sample, instead of hardcoding values in
the tests. Other tests will be refactored in follow up PRs.


noahdietz (approved):
This is really cool. Love the idea, thanks @julieqiu

PR #1520, opened 2024-05-15
chore(internal/gengapic): refactor g.init to newGenerator

julieqiu:
At the moment, the `g.init` modifies a type `generator`. In many cases, `g` is empty and populated by `g.init`. It is generally more idiomatic to use a `new<thing>` function to construct a new instance of a type, so g.init is replaced with a `newGenerator` function to follow that pattern.

Additionally, errors were not captured in tests for some calls to `g.init`, which are now handled.

### DIFF
--- a/internal/gengapic/client_init_test.go
+++ b/internal/gengapic/client_init_test.go
@@ -552,7 +552,6 @@ func TestClientInit(t *testing.T) {
 				t.Fatal(err)
 			}
 			g.apiName = "Awesome Foo API"
-			g.imports = map[pbinfo.ImportSpec]bool{}
 			g.comments = map[protoiface.MessageV1]string{
 				tst.serv:                "Foo service does stuff.",
 				tst.serv.GetMethod()[0]: "Does some stuff.",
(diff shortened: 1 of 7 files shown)


internal/gengapic/client_init_test.go
```diff
@@ -549,7 +546,12 @@ func TestClientInit(t *testing.T) {
 				Parameter: tst.parameter,
 				ProtoFile: fds,
 			}
-			g.init(&request)
+			g, err := newGenerator(&request)
+			if err != nil {
+				t.Fatal(err)
+			}
+			g.apiName = "Awesome Foo API"
+			g.imports = map[pbinfo.ImportSpec]bool{}
```
quartzmo, line 554: Is this redundant given the init of this field in `newGenerator` (line 97)?

quartzmo (approved):

noahdietz (approved):
Nice! Agreed on Chris' comment though

internal/gengapic/client_init_test.go
julieqiu: removed

PR #1523, opened 2024-05-21
feat(gengapic): conditionally enable cloud.google.com/go/auth

quartzmo:
* Initial allowlist: `cloudkms.googleapis.com`, `secretmanager.googleapis.com`

### DIFF
--- a/internal/gengapic/generator.go
+++ b/internal/gengapic/generator.go
@@ -36,10 +36,10 @@ import (
 	"google.golang.org/protobuf/types/pluginpb"
 )
 
-var allowListEnableNewAuthLibrary = []string{
-	"cloudkms.googleapis.com",
-	"secretmanager.googleapis.com",
-	"showcase.googleapis.com",
+var enableNewAuthLibrary = map[string]bool{
+	"cloudkms.googleapis.com":      true,
+	"secretmanager.googleapis.com": true,
+	"showcase.googleapis.com":      true,
 }
 
 type generator struct {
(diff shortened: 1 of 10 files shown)


internal/gengapic/generator.go
```diff
@@ -36,6 +36,12 @@ import (
 	"google.golang.org/protobuf/types/pluginpb"
 )
 
+var allowListEnableNewAuthLibrary = []string{
+	"cloudkms.googleapis.com",
+	"secretmanager.googleapis.com",
+	"showcase.googleapis.com",
+}
+
```
noahdietz, line 44: I'd just use a `map[string]bool` - no need for custom lookup code.

internal/gengapic/gengrpc.go
```diff
@@ -185,6 +185,9 @@ func (g *generator) grpcClientOptions(serv *descriptorpb.ServiceDescriptorProto,
 	p("    internaloption.WithDefaultAudience(%q),", generateDefaultAudience(host))
 	p("    internaloption.WithDefaultScopes(DefaultAuthScopes()...),")
 	p("    internaloption.EnableJwtWithScope(),")
+	if g.enableNewAuthLibrary() {
```
noahdietz, line 188: This would become `_, ok := newAuthLibraryAllowlist[g.serviceConfig.GetName()]; ok {`

internal/gengapic/generator.go
quartzmo: Updated, thanks.

noahdietz (approved):

PR #1533, opened 2024-06-14
fix(gencli): dont return iterator.Done

liveFreeOrCode:
When generating a `Paged` method: https://google.aip.dev/132, the `err` at the end will always be an `iterator.Done` (as long as the iterator was built successfully).

List CLI methods should **not** return `iterator.Done` method.


noahdietz:
I keep finding that folks actually use this! haha Thanks for the fix, you might need to update the baselines as well.

noahdietz:
> you might need to update the baselines as well.

or not! that's a gap...would you mind adding a test case to the `cmd_file_test.go`? Then run 
`go test github.com/googleapis/gapic-generator-go/internal/gencli -update_golden` to write the baseline to disk assuming it is correct

quartzmo (approved):

quartzmo:
@liveFreeOrCode Thank you for this contribution! 🎉

PR #1542, opened 2024-06-28
feat(gengapic): generate helpers for Go 1.23 iterators

codyoss:
This changes starts generating build tag protected adapters for using our iterator types in Go range expressions.

Related: https://github.com/googleapis/gax-go/pull/358


codyoss:
requires gax changes to be fully released and rolled out to clients

codyoss:
The new gax libraries have been rolled out to all clients. This is now safe for review.

noahdietz (approved):
Looks great. Let's just make sure we don't merge this until we are ready to have it in our clients e.g. in case we need to fix something and cut a release and this would be in that release.

codyoss:
Agreed. Let's wait until 1.23 is out, which should be in about 2-3 weeks.

internal/gengapic/paging.go
```diff
@@ -389,3 +389,20 @@ func (g *generator) pagingIter(pt *iterType) {
 		g.imports[spec] = true
 	}
 }
+
+func (g *generator) pagingIterGo123(pt *iterType) {
+	p := g.printf
+
+	p("// All returns an iterator. If an error is returned by the iterator, the")
+	p("// iterator will stop after that iteration.")
```
quartzmo, line 397: Maybe mention here how to access the error?

codyoss: I decided to just generate an example of the usage. LMK if that is good enough or you still think we should say more.

quartzmo: I think the example is great, it makes everything clear.

quartzmo (approved):

PR #1552, opened 2024-08-08
docs: trivial tweak to verify testing

shollyman:


quartzmo:
@noahdietz Any idea what might be causing this failure?

noahdietz:
> @noahdietz Any idea what might be causing this failure?

The issue seems to be that the gRPC streaming clients changed concrete types and `apidiff` has flagged it.

Went from an `interface` e.g. `genproto.Echo_ChatClient` to a generic `google.golang.org/grpc.BidiStreamingClient[github.com/googleapis/gapic-showcase/server/genproto.EchoRequest, github.com/googleapis/gapic-showcase/server/genproto.EchoResponse]`

Looks like this was a recent change in the gRPC code generator and our CI job for this uses `latest`: https://github.com/googleapis/gapic-generator-go/blob/main/.github/workflows/ci.yaml#L70

Perhaps we need to add the flag they mention in the notes to disable this feature until we want to take it

quartzmo:
@noahdietz Thanks for figuring that out!

shollyman:
I would have never found this without the pointer!

Turns out that's not the flag that we need, but it is a new flag in the plugin.  What we need is `--go-grpc_opt=use_generic_streams_experimental=false` which was defaulted to true roughly a month ago.  I've verified that it resolves the issue in https://github.com/googleapis/gapic-generator-go/pull/1553, I'll work on following up with grpc team and sorting out what we should be doing here.

I'm going to close this PR for now.

PR #1559, opened 2024-08-19
fix: use correct function name

codyoss:
Ref: https://pkg.go.dev/google.golang.org/protobuf/proto#Uint32


quartzmo (approved):
https://pkg.go.dev/github.com/golang/protobuf/proto#Uint32

PR #1561, opened 2024-08-23
fix: make iter error handling clearer

codyoss:


quartzmo (approved):
Good addition.

Another possibility might be:

```
// TODO: Handle error. Iteration will stop after any error.
break
```

PR #1563, opened 2024-08-27
chore: upgrade go.mod go versions to 1.23

quartzmo:
* showcase/gen/github.com/googleapis/gapic-showcase/client/auxiliary.go:18:1: package requires newer Go version go1.23

refs: #1562


quartzmo:
Replaced by #1568.

Issue #1554, opened 2024-08-12
Testing Issue: breaking changes related to grpc generator


shollyman:
As part of showcase testing in this generator, CI tests effectively compare a recent release of the showcase generated code to one that's output by the generator using apidiff. 

The grpc plugin (https://pkg.go.dev/google.golang.org/grpc/cmd/protoc-gen-go-grpc) has a new option (`use_generic_streams_experimental`) that recently defaulted to true.  Because our showcase testing grabs the plugin latest version, we were picking up that new change.

To resolve our immediate issue, we need to add the experimental flag to our build process to disable the newly enabled behavior, and verify what further actions are needed.

quartzmo:
Great diagnosis, thank you!

hongalex:
It seems like that [flag was removed](https://github.com/grpc/grpc-go/pull/8936) and is causing our latest dependency update PRs to fail.

PR #1569, opened 2024-08-29
chore: upgrade rules_go to 0.49.0

quartzmo:
* upgrade bazel_gazelle to 0.38.0
* upgrade com_google_protobuf to 25.4
* add rules_proto 6.0.2
* remove -use_generic_streams_experimental
  * fix error: protoc-gen-go-grpc: no such flag -use_generic_streams_experimental

refs: #1554

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -19,9 +19,9 @@ bazel_skylib_workspace()
 
 http_archive(
     name = "com_google_protobuf",
-    sha256 = "af9236a5b5b0f641b20f5511c1e4527efa981ac91d7342c6b1033f4f03bc5d21",
-    strip_prefix = "protobuf-25.4",
-    urls = ["https://github.com/protocolbuffers/protobuf/archive/v25.4.tar.gz"],
+    sha256 = "d19643d265b978383352b3143f04c0641eea75a75235c111cc01a1350173180e",
+    strip_prefix = "protobuf-25.3",
+    urls = ["https://github.com/protocolbuffers/protobuf/archive/v25.3.tar.gz"],
 )
 
 load("@com_google_protobuf//:protobuf_deps.bzl", "protobuf_deps")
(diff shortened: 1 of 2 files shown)


noahdietz:
> remove -use_generic_streams_experimental
> * fix error: protoc-gen-go-grpc: no such flag -use_generic_streams_experimental

We will need to take the "breaking change" introduced by this if we want to remove it. The change applies to our showcase testing stuff, so not that big of a deal tbh. 

grpc-go went from a stream client `interface`, to a Go "generic" type, which apidiff flags as an incompat change in the grpc code we generate on the fly for showcase.

quartzmo:
@noahdietz You mean we just need to merge this PR in spite of the failing apidiff, correct? Or do you think the merge commit should flag the breaking change with a bang, and that the release notes should mention it?

quartzmo:
I ran `bazelisk clean --expunge && bazelisk build //google/cloud/asset/v1:asset_go_gapic` locally against latest googleapis with googleapis/WORKSPACE pointed at local branch for this PR, is that what you mean?

quartzmo:
@noahdietz My local `protoc` was outdated, which is why I got the error `no such flag -use_generic_streams_experimental` and mistakenly assumed the flag had been removed. Restored.

codyoss:
@quartzmo I would try to build the whole package. The reason I ask is because our workspace in googleapis explicitly sets the version of some of these deps. But I think this change would cause them to raise:
- https://github.com/googleapis/googleapis/blob/6a474b31c53cc1797710206824a17b364a835d2d/WORKSPACE#L152
- https://github.com/googleapis/googleapis/blob/6a474b31c53cc1797710206824a17b364a835d2d/WORKSPACE#L111

quartzmo:
@codyoss Thanks for pointing that out. I reverted the updates to `com_google_protobuf` and `bazel_gazelle`, leaving only the `io_bazel_rules_go` update.

PR #1572, opened 2024-09-18
feat: add logging to gapics

codyoss:
As a follow up to this PR we will need to add a little logging support in the gax REST streaming helper. For now we should add this with minimal changes but in the future we may want to feed that helper more information to have richer logs. This is a very small amount of RPCs that use this code path today.


codyoss:
Requires https://github.com/googleapis/gax-go/pull/372 and rollout to all clients and showcase.

PR #1573, opened 2024-10-08
fix(internal/gencli): revert broken migration to protobuf-go v2 in gencli

noahdietz:
Reverts googleapis/gapic-generator-go#1555, which broke gencli generation. The dependency replacement for proto-JSON encoding actually requires an upgrade (it is not a compat api) and the `proto` lib change also requires changes to generated cli code.

Since the gencli isn't a production system and should probably be cleaned up entirely, I'm just reverting this change to unblock development on showcase. What is generated is currently broken.


codyoss (approved):
Thanks, maybe file a bug?

PR #1577, opened 2024-11-01
feat: add logging support to generated clients

codyoss:
This change instruments all generated clients with logging. In order to reduce the amount of codegen this would take I refactored some common logic about executing the request and parsing the response into a common helper. This change should reduce LoC for generated clients as a nice side-effect.


codyoss:
Requires https://github.com/googleapis/gax-go/pull/380 and https://github.com/googleapis/google-api-go-client/pull/2854 be rolled out to all clients first

internal/gengapic/gengapic.go
```diff
@@ -257,6 +268,52 @@ func (g *generator) genAndCommitHelpers(scopes []string) error {
 	}
 	p("  }")
 	p("}")
+	p("")
+
```
quartzmo, line 272: Can these HTTP and gRPC-specific funcs be moved from `genAndCommitHelpers` to somewhere where they will only be generated for clients with HTTP or gRPC methods? (`restClientInit` or similar?)

internal/gengapic/gengrpc.go
```diff
@@ -295,6 +295,8 @@ func (g *generator) grpcClientInit(serv *descriptorpb.ServiceDescriptorProto, se
 
 	p("// The x-goog-* metadata to be sent with each request.")
 	p("xGoogHeaders []string")
+	p("")
+	p("logger *slog.Logger")
```
quartzmo, line 299: Could this and the import below be moved to `clientInit`?

internal/gengapic/genrest.go
```diff
@@ -91,13 +91,16 @@ func (g *generator) restClientInit(serv *descriptorpb.ServiceDescriptorProto, se
 	p("")
 	p("  // Points back to the CallOptions field of the containing %sClient", servName)
 	p("  CallOptions **%sCallOptions", servName)
+	p("")
+	p("  logger *slog.Logger")
```
quartzmo, line 95: Could this and the import below be moved to `clientInit`?

internal/gengapic/gengrpc.go
codyoss: no, this is needed in by the concrete impl clients. It could also be added to that top level client, but I don't think there is a benefit.

internal/gengapic/genrest.go
codyoss: no, this is needed in by the concrete impl clients. It could also be added to that top level client, but I don't think there is a benefit.

internal/gengapic/gengapic.go
codyoss: Sure. Followup question. Are you interesting in gating these so they only generate if the transport is enabled or are you saying generate this code in the client file instead of the helper file?

quartzmo: I was just thinking about gating these so they only generate if the transport is enabled .

quartzmo (approved):

PR #1580, opened 2024-11-05
feat: add service renaming

### LINKED ISSUE #1581: feat: support service overriding
As part of internal issue 360155816, we will be supporting the ability to override the default Service name provided in the proto and generate Clients and other structs with a custom service name. For example, we wish to override Pub/Sub's `PublisherService` to generate `TopicAdminClient` instead of `PublisherClient`

hongalex:
This PR adds support for service name overriding (renaming).

~~This PR cannot be merged until the new client.proto changes land to support the `renamed_services` field and the actual code in `generator.getServiceNameOverride` is uncommented.~~

Pieces of the generator that are currently not changed by service renaming:
- Custom operations (in genrest.go)
- gencli
-  [snippets](https://github.com/googleapis/gapic-generator-go/tree/624962423ddf8b57b9b0cd4138bde6bbdd491427/internal/snippets)

Diff for Pub/Sub is viewable here: https://github.com/googleapis/google-cloud-go/compare/main...hongalex:google-cloud-go:new-gapic-diff?expand=1

Fixes #1581


codyoss (commented):
Could you extend one of the existing tests to show this is working as expected?

quartzmo (commented):
I noticed the following generated doc in the provided diff:

```
// NewTopicAdminClient creates a new publisher client based on gRPC.
```

Note: "new publisher client".
This seems acceptable to me for most cases, but I suppose it could result in some confusion depending on the original name.

hongalex:
> Note: "new publisher client". This seems acceptable to me for most cases, but I suppose it could result in some confusion depending on the original name.

Yes, that was a one off where the code calls `serv.GetName` that I didn't catch before. I took a look again and it seems like the other references are fine. Please check the [diff branch again](https://github.com/googleapis/google-cloud-go/compare/main...hongalex:google-cloud-go:new-gapic-diff?expand=1).

quartzmo:
I still found one usage in the diff, if you want to look into it:

```
 NewTopicAdminRESTClient creates a new publisher rest client.
```

hongalex:
@quartzmo that should be fixed now

@codyoss I added a test that generates a stub of the gRPC client with service renaming, which I think is sufficient.

quartzmo (approved):

PR #1585, opened 2024-11-14
fix: revert new auth lib for storage

tritone:
This partly reverts commit 74b6c39e442857ee568756c0fff8bc3c68200cc7.


tritone:
Preempted by googleapis/google-cloud-go#11135 which rolls the auth dep forward instead

PR #1594, opened 2024-12-09
fix: have a helper that returns a http Response as well

codyoss:
When a rpc returns an httpBody, see
https://pkg.go.dev/google.golang.org/genproto/googleapis/api/httpbody#HttpBody, we need to return a response from the helper so we can examine the headers.


internal/gengapic/genrest.go
```diff
@@ -1135,7 +1135,11 @@ func (g *generator) unaryRESTCall(servName string, m *descriptorpb.MethodDescrip
 	p("  httpReq = httpReq.WithContext(ctx)")
 	p("  httpReq.Header = headers")
 	p("")
-	p("  buf, err := executeHTTPRequest(ctx, c.httpClient, httpReq, c.logger, %s, %q)", logBody, m.GetName())
+	if isHTTPBodyMessage {
+		p("  buf, httpRsp, err := executeHTTPRequestWithResponse(ctx, c.httpClient, httpReq, c.logger, %s, %q)", logBody, m.GetName())
```
quartzmo, line 1139: How does `httpRsp` get used in the client?

codyoss: I would direct you to the updated generated file: internal/gengapic/testdata/rest_HttpBodyRPC.want

quartzmo: Thanks, when I look at the whole file I see the usage:

``
if headers := httpRsp.Header; len(headers["Content-Type"]) > 0 {
			resp.ContentType = headers["Content-Type"][0]
		}
```

But it doesn't show up in the diff. Odd that this usage existed before.

quartzmo (approved):

PR #1595, opened 2024-12-09
chore: bump deps

codyoss:
Also had to regen some internal protos to make protoc-gen-go happy with staticcheck.


internal/grpc_service_config/grpc_service_config.pb.go
```diff
@@ -29,8 +29,8 @@
 
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
-// 	protoc-gen-go v1.33.0
-// 	protoc        v3.21.12
+// 	protoc-gen-go v1.35.2
+// 	protoc        v5.28.3
```
quartzmo, line 33: Is it OK to use protoc `v5.28.3` to generate these files? In our build pipeline the generator will be invoked with version `25` IIRC.

quartzmo (changes requested):

internal/grpc_service_config/grpc_service_config.pb.go
codyoss: It should make no difference, especially because it is internal to the generator. But I pushed the commit anyways which just updated this line.

quartzmo (approved):

PR #1600, opened 2024-12-19
fix: update golang.org/x/net to v0.33.0

quartzmo:


codyoss:
Looks like this needs a go mod tidy

PR #1602, opened 2025-01-16
feat: prevent body double-read in googleapi.CheckResponse

bjohnso5:
Call into non-reading variant of CheckResponse added [here](https://github.com/googleapis/google-api-go-client/pull/2964).

Fixes https://github.com/googleapis/google-cloud-go/issues/11458 (part 2 of 2).


quartzmo:
Setting `do not merge` label pending publication of change in https://github.com/googleapis/google-api-go-client/pull/2964.

bjohnso5:
@quartzmo do I need to update the go.mod in the showcase to point at the 0.218.0 release of `google.golang.org/api`?

codyoss (approved):

codyoss:
Looks like we need to add a new leakchecker skip. maybe `transport/controlbuf.go`. The file that holds the items to ignore is `leakcheck.go`. Other than that this looks good to me. This is likely do to the new version on grpc

showcase/leakcheck.go
```diff
@@ -44,6 +44,7 @@ var goroutinesToIgnore = []string{
 	"runtime_mcall",
 	"(*loggingT).flushDaemon",
 	"goroutine in C code",
+	"internal/transport/http2_client.go",
```
bjohnso5, line 47: @codyoss this led to a successful local run of `showcase/showcase.bash`, but I'm not sure I have enough background context to understand if this is the right thing to ignore. If this is the accepted change, I'll squash down these commits to get this changeset landed.

codyoss: lgtm. I can squash on merge :)

codyoss:
@bjohnso5 Thank you for all your work on these contributions! We will get our generator updated. I expect the new code won't land until early next week though just FYI.

PR #1634, opened 2025-06-04
feat(gengapic): conditionally enable directpath for storagecontrol

rockspore:
Given storagecontrol being the only use case we know, hard-coding it in the generator seems the most convenient approach.

To uniquely identity the service, it uses the API and service name joined by "/".


rockspore:
@quartzmo @tritone PTAL. This is for b/413419701.

quartzmo (approved):

quartzmo:
If it ever turns out that the test enablement for `Foo` clashes with some other testing goals, we might need to move the testing here to a bespoke RPC path (`FooDirectPath`). For now I guess this is OK.

tritone (approved):

PR #1698, opened 2026-02-06
fix(gengapic): use recursive resource name heuristic

quartzmo:
Refactor the resource identifying field heuristic to be recursive and address specific gaps in the current implementation.

Satisfies Requirements:

1.  **Nested Field Support:** Recursively searches for fields annotated with `google.api.resource_reference` up to 2 levels deep. This directly addresses AIP-134 (Update) scenarios where the resource is often nested (e.g., `req.Secret.Name`).
2.  **Tie-breaking:** Prioritizes annotated fields that also appear in the method's `google.api.http` path template, ensuring the correct "primary" resource is selected when multiple references exist.
3.  **Code Generation:** Returns the full field path (e.g., `secret.name`) to generate idiomatic Go accessors for nested fields (e.g., `req.GetSecret().GetName()`), addressing AIP-134 scenarios.

TODO (Remaining Requirements from Design):

- [ ] **Vocabulary Learning:** Implement generation-time scanning to build a
    vocabulary of valid resource types from HTTP path literals. This is
    required for legacy services (Compute, SQL, BigQuery) that lack
    annotations.
- [ ] **Path Reconstruction:** Implement logic to parse "exploded" HTTP
    paths and generate code to reassemble multiple request fields into a
    single canonical resource name.
- [ ] **Path Pruning:** Implement logic to prune non-resource segments
    (custom verbs) from reconstructed paths.
- [ ] **Heuristic Allow-listing:** Add configuration to restrict these
    advanced heuristics to the specific legacy services identified in the
    design.

Tests added in `internal/gengapic/resource_name_test.go` verify the recursive search and tie-breaking logic.

### DIFF
--- a/internal/gengapic/gengapic.go
+++ b/internal/gengapic/gengapic.go
@@ -886,12 +886,12 @@ func (g *generator) resourceNameField(m *descriptorpb.MethodDescriptorProto) str
 		return ""
 	}
 
-	var candidates []string
-	for _, f := range msg.GetField() {
-		if proto.HasExtension(f.GetOptions(), annotations.E_ResourceReference) {
-			candidates = append(candidates, f.GetName())
-		}
-	}
+	// Recursively find fields with the annotation, up to 2 levels deep.
+	// This covers top-level "name" or "parent" fields, and also nested
+	// fields like "secret.name" (AIP-134).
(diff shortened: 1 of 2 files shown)


quartzmo:
Replaced by #1722

PR #1725, opened 2026-04-13
chore(showcase): add tracing tests

quartzmo:
This PR is based on @westarle 's work on [westarle:gapic-generator-go:test-plan-go-tf1-logs](https://github.com/googleapis/gapic-generator-go/compare/main...westarle:gapic-generator-go:test-plan-go-tf1-logs). It adds the following:

1.  Upgraded dependencies in `showcase/go.mod` (specifically `google.golang.org/api`, `cloud.google.com/go/auth`, and `otelgrpc`). This was crucial because the older versions weren't correctly propagating telemetry attributes through the transitive chain.
2.  Updated `trace_test.go` to be less strict about unexpected attributes. The upgraded dependencies now automatically add `gcp.grpc.resend_count` and `rpc.response.status_code` in some cases where the tests previously disallowed them. I removed them from the `unexpectedAttrs` lists.
3.  Added a `telemetry` build tag to exclude these tests from `make test`.
4.  Added a new section to `showcase/README.md` explaining how the telemetry tests work, including the `telemetry` build tag, environment variables, and how to find the traces in Cloud Trace.

### Verification

I successfully ran the new `make test-telemetry` target on my cloudtop using my dev project. The integration test passed in ~36 seconds.

I also visited Trace Explorer to verify that the spans were arriving as expected (I had to adjust the time range to "Last 1 hour" to see them). I filtered for the client spans wiht `AttemptSequence` and they looked as expected with the T4 telemetry attributes we were looking for. Here's an example of attributes from an error span:

cloud.account.id	cloudtop-prod-us-west
cloud.availability_zone	us-west4-a
cloud.platform	gcp_compute_engine
cloud.provider	gcp
cloud.region	us-west4
error.type	CLIENT_TIMEOUT
exception.type	*status.Error
gcp.client.artifact	github.com/googleapis/gapic-showcase/client
gcp.client.language	go
gcp.client.repo	googleapis/google-cloud-go
gcp.client.service	showcase
gcp.client.version	UNKNOWN
gcp.gce.instance.hostname	myproject.us-west4-a.c.cloudtop-prod-us-west.internal
gcp.gce.instance.name	myproject
gcp.grpc.resend_count	0
gcp.project_id	myproject-testing
gcp.resource.destination.id	//localhost:7469/sequences/2
host.id	xxxx
host.name	myproject
host.type	projects/xxxxxxxx/machineTypes/e2-custom-24-98304
rpc.method	google.showcase.v1beta1.SequenceService/AttemptSequence
rpc.response.status_code	DEADLINE_EXCEEDED
rpc.system.name	grpc
server.address	127.0.0.1
server.port	7469
service.name	test-app
status.message	context deadline exceeded
telemetry.sdk.language	go
telemetry.sdk.name	opentelemetry
telemetry.sdk.version	1.43.0
url.domain	showcase.googleapis.com

This proves that the generated client is correctly delegating to the modern transport wrappers under the hood and emitting the necessary telemetry, without us needing to change the generator structure itself


showcase/trace_test.go
```diff
@@ -0,0 +1,550 @@
...
+		clientOpts = []option.ClientOption{
+			option.WithEndpoint("http://127.0.0.1:7469"),
+			option.WithTokenSource(oauth2.StaticTokenSource(&oauth2.Token{AccessToken: "dummy-token"})),
+		}
+	}
+
+	return fix, clientOpts
+}
+
+func verifyInMemorySpan(t *testing.T, fix *observabilityFixture, expectedName string, traceID trace.TraceID, wantAttrs map[string]any, unexpectedAttrs []string) {
```
shollyman: Do you feel like we're going to have a lot of additional span validation tests in the future?  I'm mostly trying to reason about whether we want this method to have this baked in knowledge of what span attrs are dynamic or if we should have an easy way to express something like "expected these attribute keys with non-nil values" vs "expected these attribute-value pairs".

quartzmo: @westarle What's your take on the brittleness of these assertions?

showcase/cloud_trace_test.go
```diff
@@ -0,0 +1,207 @@
...
+	projectID := setupCloudTrace(t)
+	ctx := context.Background()
+
+	grpcClientOpts := []option.ClientOption{
+		option.WithEndpoint("127.0.0.1:7469"),
+		option.WithTokenSource(oauth2.StaticTokenSource(&oauth2.Token{AccessToken: "dummy-token"})),
+		option.WithGRPCDialOption(grpc.WithTransportCredentials(insecure.NewCredentials())),
+	}
+
+	seqClient, err := showcase.NewSequenceClient(ctx, grpcClientOpts...)
```
quartzmo: @westarle Should we perform the same tests and assertions for the matching showcase REST clients?

showcase/trace_test.go
```diff
@@ -0,0 +1,550 @@
...
+					"server.address":              "127.0.0.1",
+					"server.port":                 int64(7469),
+					"status.message":              "context deadline exceeded",
+					"url.domain":                  "showcase.googleapis.com",
+				}
+				unexpectedAttrs = []string{}
+			} else {
+				expectedName = "POST /v1beta1/{name=sequences/*}"
+				wantAttrs = map[string]any{
+					"error.type":                  "context.deadlineExceededError",
```
quartzmo: @westarle Is`error.type: context.deadlineExceededError` an implementation bug? Should the raw client exception by replaced with `CLIENT_TIMEOUT` as it is in gRPC?

showcase/cloud_trace_test.go
```diff
@@ -0,0 +1,207 @@
...
+	"golang.org/x/oauth2"
+	"golang.org/x/oauth2/google"
+	"google.golang.org/api/option"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials"
+	"google.golang.org/grpc/credentials/insecure"
+	"google.golang.org/grpc/credentials/oauth"
+)
+
+func setupCloudTrace(t *testing.T) string {
```
westarle, line 51: nit: it'd be helpful to write down what this does as a doc.  I think it configures and installs a new tracer provider with a GCP resource detector that points to telemetry.googleapis.com.

showcase/cloud_trace_test.go
```diff
@@ -0,0 +1,207 @@
...
+	t.Cleanup(func() {
+		ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+		defer cancel()
+		tp.Shutdown(ctx)
+	})
+
+	return projectID
+}
+
+func verifyTrace(t *testing.T, ctx context.Context, traceClient *trace.Client, projectID string, traceID [16]byte) {
```
westarle:
```suggestion
func verifyTraceExists(t *testing.T, ctx context.Context, traceClient *trace.Client, projectID string, traceID [16]byte) {
```

showcase/cloud_trace_test.go
westarle: yes please!

showcase/observability_fixture_test.go
```diff
@@ -0,0 +1,385 @@
...
+	semconv "go.opentelemetry.io/otel/semconv/v1.26.0"
+	pblog "go.opentelemetry.io/proto/otlp/collector/logs/v1"
+	pbmetric "go.opentelemetry.io/proto/otlp/collector/metrics/v1"
+	pb "go.opentelemetry.io/proto/otlp/collector/trace/v1"
+	v1common "go.opentelemetry.io/proto/otlp/common/v1"
+	olog "go.opentelemetry.io/proto/otlp/logs/v1"
+	"google.golang.org/grpc"
+)
+
+type mockTraceServer struct {
```
westarle, line 36: nit: it might help to have some comments on structs to serve as breadcrumbs like "This struct represents an OTLP service that receives traces" and for CapturedSpan etc.

nit: this is probably a "Fake" since it doesn't do any verifications.

showcase/observability_fixture_test.go
```diff
@@ -0,0 +1,385 @@
...
+						Attributes: attrs,
+					})
+				}
+			}
+		}
+	}
+	return spans
+}
+
+type mockMetricServer struct {
```
westarle: Nit: remove from this PR if we aren't using it yet?

showcase/trace_test.go
```diff
@@ -0,0 +1,550 @@
...
+	time.Sleep(100 * time.Millisecond)
+
+	spans := fix.traceServer.GetCapturedSpans()
+	if len(spans) == 0 {
+		t.Fatalf("expected to receive trace exports, got none")
+	}
+
+	var gotSpan *CapturedSpan
+	for _, s := range spans {
+		if string(s.TraceID) == string(traceID[:]) && s.Name == expectedName {
```
westarle, line 97: maybe this function can just be responsible for returning the capturedspan

Then tests can do whatever assertions they like.

My personal preference is each test does a strict attribute map comparison, and we deal with version updates etc. as they come as reminders that the test is working.

showcase/trace_test.go
westarle: this seems like a bug, could you file one please?

showcase/trace_test.go
westarle: I think it might help to just have this return the span, then tests can assert on what they know about instead of trying to communicate it to a "verify" function that might have bugs.

I'd like to see the DYNAMIC elements replaced with actual values, unless we really can't fix them.  Then the tests could remove them from the map and do a realistic assertion (not nil, semver, etc.)

showcase/trace_test.go
quartzmo: I verified via a local test that the correct mapping logic to CLIENT_TIMEOUT in auth/httptransport is executing. However, the outer otelhttp.NewTransport wrapper overwrites the error.type attribute with the raw Go error type after our custom transport returns.

quartzmo: Opened bug https://github.com/googleapis/google-cloud-go/issues/14441

showcase/cloud_trace_test.go
quartzmo: Done!

showcase/observability_fixture_test.go
quartzmo: Done, along with mockLogServer and related.

showcase/observability_fixture_test.go
quartzmo: Done! But I kept "mock" for easy pattern matching although technically incorrect.

showcase/trace_test.go
quartzmo: I'll refactor verifyInMemorySpan to simply return the captured span. Re: brittle tests, we can validate the presence and format of dynamic attributes like versions or URLs rather than hardcoding values. I'd rather avoid the maintenance burden of strict comparisons, hopefully a pattern-matching approach is acceptable.

showcase/trace_test.go
quartzmo: Refactored to findInMemorySpan and assertInMemorySpan. Will address assertion brittleness in a separate commit. See comment below.

showcase/trace_test.go
quartzmo: Done, PTAL.

showcase/cloud_trace_test.go
```diff
@@ -141,67 +146,107 @@ func TestObservability_Tracing_CloudTrace_Integration(t *testing.T) {
...
+				} else {
+					_ = runTracingServerFailureScenarioREST(ctxSpan, t, seqClient)
+				}
+				span.End()
+				traceID := span.SpanContext().TraceID()
+				otel.GetTracerProvider().(*sdktrace.TracerProvider).ForceFlush(ctx)
+				verifyTraceExists(t, ctx, traceClient, projectID, traceID)
+			})
+
+			// 3. Client Failure Scenario
```
westarle, line 223: nit: we could remove this comment and rely on the test name?

westarle (approved):

westarle:
> Updated trace_test.go to be less strict about unexpected attributes. The upgraded dependencies now automatically add gcp.grpc.resend_count and rpc.response.status_code in some cases where the tests previously disallowed them. I removed them from the unexpectedAttrs lists.

In a followup, (1) I think we should assert on the "wrong" values.  Also (2) I think with your new "constraints" concept, we should be able to assert on every attribute in the map and report any "new" attribute as a failure (so we remember to add an appropriate test.)

showcase/cloud_trace_test.go
quartzmo: Done, thanks!

quartzmo:
> (1) I think we should assert on the "wrong" values. Also (2) I think with your new "constraints" concept, we should be able to assert on every attribute in the map and report any "new" attribute as a failure (so we remember to add an appropriate test.)

Pushed fix for this in  d727d5b

showcase/trace_test.go
```diff
@@ -132,6 +132,26 @@ func assertCapturedSpan(t *testing.T, gotSpan *CapturedSpan, wantAttrs map[strin
...
 	}
+
+	// Report any "new" attribute not covered by wantAttrs or constraints as a failure
+	for k := range gotSpan.Attributes {
+		_, inWant := wantAttrs[k]
+		_, inConstraints := constraints[k]
+
+		if !inWant && !inConstraints {
+			// Check if it was explicitly expected to be absent
+			isUnexpected := false
```
westarle: can we express this as a constraint that always fails to make things simpler?

westarle (approved):

showcase/trace_test.go
quartzmo: Great suggestion, done.

westarle (approved):

PR #1726, opened 2026-04-15
feat: make exporting goog client info a generator capability

shollyman:
This PR moves the functionality to export setting google client info into the generator layer, enabled by a new feature flag.  The internal issue b/489495186 describes this in more detail.

This PR:
* adds a new featureID for controlling this behavior, and also sorts the entries in feature.go
* Adds new logic to add an exported SetGoogleClientInfo alongside the non-exported version in the public Client contract.  The implementation itself calls the non-exported version of the func.
* Adds a golden testcase for client code


internal/gengapic/client_init.go
```diff
@@ -235,6 +235,15 @@ func (g *generator) clientInit(serv *descriptorpb.ServiceDescriptorProto, servNa
 	p("func (c *%sClient) setGoogleClientInfo(keyval ...string) {", servName)
 	p("  c.internalClient.setGoogleClientInfo(keyval...)")
 	p("}")
+	if g.featureEnabled(ExportSetGoogleClientInfoFeature) {
+		p("")
+		p("// SetGoogleClientInfo sets the name and version of the application in")
+		p("// the `x-goog-api-client` header passed on each request. Intended for")
+		p("// use by Google-written clients.")
```
quartzmo, line 242: What do you think about adding a concurrent-access warning here? 

The generated client docs state:

```
// Methods, except Close, may be called concurrently. However, fields must not be modified concurrently with method calls.
```

However, SetGoogleClientInfo (a method, not a field) modifies an internal field (xGoogHeaders) in place without synchronization.

Maybe:

```
// SetGoogleClientInfo is not concurrency-safe and should only be invoked
// sequentially before concurrent operations begin.
```

shollyman: Yeah, this is a relatively static usage so I think the caveat in the method documentation is sufficient.  If we find more dynamic usages down the road we can always revisit.

quartzmo (approved):

PR #1728, opened 2026-04-20
feat(gengapic): update generator to inject client_span_name for client spans

westarle:
This PR updates the Go GAPIC generator to inject a client_span_name into the telemetry context for every RPC method. This is a prerequisite for implementing T3 spans (Client Request) in gax-go, allowing the runtime to know the logical name of the call (e.g., GoPackage.Struct.Method).

The attribute is injected inside the existing telemetry feature flag guard (gax.IsFeatureEnabled("METRICS") || gax.IsFeatureEnabled("TRACING") || ...), respecting the user opt-in model.


internal/gengapic/gengapic.go
```diff
@@ -706,8 +706,13 @@ func (g *generator) injectTelemetryContext(m *descriptorpb.MethodDescriptorProto
 		serv := g.descInfo.ParentElement[m].(*descriptorpb.ServiceDescriptorProto)
 		fqn := fmt.Sprintf("%s.%s/%s", g.descInfo.ParentFile[serv].GetPackage(), serv.GetName(), m.GetName())
 
+		override := g.getServiceNameOverride(serv)
+		servName := pbinfo.ReduceServNameWithOverride(serv.GetName(), g.cfg.pkgName, override)
+		clientSpanName := fmt.Sprintf("%s.%s.%s", g.cfg.pkgPath, servName, m.GetName())
```
shollyman, line 719: Do you think we need any guarding logic that ensures all the values are populated?  I don't know across all our generated services if there are exceptional cases here that need more scrutiny.

bhshkh: Great catch! I looked into this, and here is what we have:

1. **`pkgPath` and `m.GetName()`** are guaranteed to be populated by generator-level validation and protobuf specifications.

2. **`servName` can indeed be empty.** If the service name matches the Go package name (e.g., service `SecretManagerService` in package `secretmanager`), `pbinfo.ReduceServName` returns `""` so the generator can produce a clean `secretmanager.Client` instead of `secretmanager.SecretManagerClient`. 

If `servName` is empty, the generated code would have ended up with an invalid double dot in the middle of the span name (e.g., `cloud.google.com/go/secretmanager/apiv1..AccessSecretVersion`).

To address this, I have refactored this to introduce a new helper inside the `pbinfo` package. PTAL.

PR #1729, opened 2026-04-21
feat(internal/gapic): Selective GAPIC generation

bhshkh:
Selective GAPIC Generation (SGG) is a feature that allows service owners to tightly control the public surface area of their generated Go client libraries. 

Instead of unconditionally generating client methods for every RPC defined in a protobuf service, SGG allows the generation process to be restricted via a configuration allowlist. Methods not in the allowlist are either **completely omitted** (not generated at all) or **internalized** (generated as unexported Go methods on the canonical client struct), depending on the configuration.

Design: [go/selective-gapic-generation-go](http://goto.google.com/selective-gapic-generation-go)


bhshkh:
### 1. Summary of Changes
This PR implements **Selective GAPIC Generation (SGG)** in the Go GAPIC generator. SGG allows API producers to control the public API surface area of generated Go client libraries via an allowlist in `service.yaml`. 

This implementation supports two modes for handling RPC methods omitted from the allowlist:
1.  **Pure Omission (Default):** Omitted methods are completely excluded from the generated code.
2.  **Internalization Mode (`generate_omitted_as_internal: true`):** Omitted methods are generated as **unexported** Go methods (lowercased), while the public client struct and constructors retain their canonical names. This allows the generated client to be extended directly with handwritten methods (e.g., synchronous LRO polling helpers) in a separate, preserved file within the same package (package extension/augmentation).

The implementation uses a **Lazy Policy Filtering** approach, intercepting method collection at the point of use in each generator component rather than destructive pre-pruning of the Protobuf descriptors.

---

### 2. Detailed Technical Changes by File

#### 2.1. Generator Context (`internal/gengapic/generator.go`)
*   Added `clientProtoPkg` string field to the `generator` struct.
*   **Purpose:** Tracks the host service's protobuf package during generation. This is critical for **Mixins** (e.g., `google.longrunning.Operations`), which reside in foreign packages. When mixin methods are evaluated for SGG, this field ensures they inherit the SGG allowlist configuration of the host service currently being generated, rather than attempting to look up configuration for the mixin's own package.

#### 2.2. The SGG Policy Engine (`internal/gengapic/method_selective.go`)
Created a centralized engine to handle all SGG filtering, visibility, and naming decisions:
*   `getSelectiveGapicGeneration(protoPkg string)`: Resolves and retrieves the SGG configuration for a given proto package from the parsed `service.yaml` `LibrarySettings`.
*   `getMethods(s *descriptorpb.ServiceDescriptorProto)`: The new centralized entry point for retrieving a service's methods. It combines native methods and active mixin methods, then filters them through `shouldGenerateMethod`.
*   `shouldGenerateMethod(s, m)`: Determines if a method should be generated at all. Returns `true` if the method is allowlisted, or if internalization mode is enabled. Returns `false` (omission) otherwise.
*   `isMethodInternal(m)`: Returns `true` if SGG is active, `generate_omitted_as_internal` is `true`, and the method is *not* in the allowlist.
*   `methodName(m)`: Visibility mutator. Returns `lowerFirst(m.GetName())` if `isMethodInternal(m)` is true, unexporting it in Go. Otherwise returns the standard name.
*   `isInternalService(s)`: Scans all methods before filtering. If any method is internalized, the entire service is flagged as internal, triggering the suppression of standalone executable snippets and package documentation examples (since they would attempt to call unexported private methods).

#### 2.3. Client & Interface Composition (`internal/gengapic/client_init.go`)
*   Switched all method loops from direct `serv.GetMethod()` to `g.getMethods(serv)`.
*   Updated interface and public wrapper method generation to use `g.methodName(m)` for naming, ensuring internal methods are unexported.
*   Suppressed public wrapper method generation for internalized methods (they are only declared on the internal interface and implemented on low-level transport clients, to be wrapped by handwritten exported methods).

#### 2.4. Transport Composers (`gengrpc.go`, `genrest.go`)
*   Updated gRPC and REST method generation loops to use the filtered `g.getMethods(serv)`.
*   Updated low-level network stub and invoke methods to use `g.methodName(m)`, ensuring the unexported transport methods match the internal interface definitions.

#### 2.5. Documentation, Snippets, and Metadata (`doc_file.go`, `snippets.go`, `example.go`, `metadata.go`)
*   **`doc_file.go` & `example.go`:** Updated to use `g.getMethods(serv)` to ensure package-level documentation and inline examples only reference public methods. Example file generation is entirely suppressed for services with internalized methods.
*   **`snippets.go`:** Updated to use `g.getMethods(serv)`. Prevents the creation, writing, or registration of standalone executable snippets for completely omitted or internalized methods.
*   **`metadata.go`:** Updated `gapic_metadata.json` generation to record method names via `g.methodName(m)` instead of `m.GetName()`. This ensures that the metadata correctly maps RPCs to their actual unexported Go names (e.g., `deleteBook` instead of `DeleteBook`), preserving metadata accuracy for downstream tooling.

---

### 3. Technical Architecture & Lifecycle Flow

#### 3.1. Interception Flowchart
The flowchart below illustrates where SGG introduces logic branches into the core generation baseline (retaining canonical client names):

```mermaid
graph TD
    classDef default fill:#f9f9f9,stroke:#333,stroke-width:2px,color:#333;
    classDef oldFlow fill:#ffcccc,stroke:#cc0000,stroke-width:2px,color:#900;
    classDef newFlow fill:#ccffcc,stroke:#009900,stroke-width:2px,color:#030;
    classDef hook fill:#cce5ff,stroke:#0066cc,stroke-width:2px,color:#003366;

    A[Start: gen service loop] --> B[Set host context: g.clientProtoPkg = pkg]
    B --> F[Invoke Component: Client/GRPC/REST/Snippets]
    
    F --> G[Call g.getMethods serv]
    G --> H[Fetch allowlist from service.yaml]
    H --> I[Loop through native + mixin methods]
    
    I --> J{Method in allowlist?}
    J -->|Yes: Public| K[Include method in filtered list]
    J -->|No: Omitted| L{generate_omitted_as_internal == true?}
    
    L -->|Yes: Internalized| M[Include method in filtered list]
    L -->|No: Omitted| N[Completely skip method / Omission]
    
    K --> O[Next Method]
    M --> O
    N --> O
    
    O --> P[Iterate through Filtered List to generate Go code]
    P --> Q{isMethodInternal method?}
    Q -->|Yes| R[Call g.methodName: Lowercase first letter<br>e.g. createBook]
    Q -->|No| S[Use standard name<br>e.g. CreateBook]
    
    R --> T[Print Go Method Definition]
    S --> T
    T --> U[End Service Loop]
```

#### 3.2. Sequence Diagram
This diagram maps SGG policy enforcement across the generator lifecycle phases:

```mermaid
sequenceDiagram
    participant P as protoc (CodeGeneratorRequest)
    participant G as generator (gengapic.go)
    participant S as service config (YAML)
    participant MS as method_selective.go
    participant C as client_init.go
    participant Trans as gengrpc / genrest
    participant Snip as snippets.go

    P->>G: Gen(req)
    G->>G: collectMixins()
    
    loop For each Service in Proto Files
        G->>G: Set g.clientProtoPkg = service.Package
        
        Note over G,MS: 1. Client Initialization Phase
        G->>C: makeClients(serv, servName)
        C->>MS: g.getMethods(serv)
        MS->>S: Read LibrarySettings for g.clientProtoPkg
        S-->>MS: Return SelectiveGapicGeneration Config
        MS->>MS: Combine native + mixin methods
        
        loop For each method
            MS->>MS: shouldGenerateMethod()
            alt Omission = true
                MS-->>MS: Skip method
            else Omission = false
                MS-->>MS: Keep method
            end
        end
        MS-->>C: Return filtered methods
        C->>C: Generate Interfaces & Client Structs
        
        Note over G,Trans: 2. Transport Generation Phase
        G->>Trans: genGRPCMethods() / genRESTMethods()
        Trans->>MS: g.getMethods(serv)
        MS-->>Trans: Return filtered methods
        
        loop For each filtered method
            Trans->>MS: isMethodInternal()
            alt Internal = true
                Trans->>Trans: Generate unexported func (e.g. `foo()`)
            else Internal = false
                Trans->>Trans: Generate exported func (e.g. `Foo()`)
            end
        end
        
        Note over G,Snip: 3. Snippet Generation Phase
        G->>Snip: genAndCommitSnippets(serv)
        Snip->>MS: g.getMethods(serv)
        MS-->>Snip: Return filtered methods
        Snip->>Snip: Generate snippets ONLY for included methods
    end
```

---

### 4. Verification & Testing Done
*   **Unit Tests:** Added comprehensive unit tests in `internal/gengapic/method_selective_test.go` covering `shouldGenerateMethod`, `isMethodInternal`, and `isInternalService` across scenarios with empty allowlists, partial allowlists, mixins, and both flag states.
*   **Golden Unit Tests:** Added new golden test cases in `client_init_test.go` and `gengapic_test.go` loading `selective_gapic_generation` configuration:
    *   Verified shape for **Pure Omission** (`selective_gapic_client_init_omitted.want` check).
    *   Verified shape for **Internalization Mode** with lowercase methods and unexported stubs, while retaining canonical client names (`selective_gapic_client_init.want` check).
*   **Formatting:** Ran `gapic-generator-go` tests and verified output formatting adheres to `gofmt` guidelines.

internal/gengapic/snippets.go
```diff
@@ -63,12 +79,24 @@ func (g *generator) addSnippetsMetadataResult(m *descriptorpb.MethodDescriptorPr
...
+	if parent, ok := g.descInfo.ParentElement[m].(*descriptorpb.ServiceDescriptorProto); ok && parent != nil {
+		if g.isInternalService(parent) {
+			return
+		}
+	}
+	g.snippetMetadata.UpdateMethodResult(servName, g.methodName(m), resultType)
 }
 
 // genAndCommitSnippets generates and commits a snippet file for each method in a client.
 // Does nothing and returns nil if opts.omitSnippets is true.
```
hongalex:
```suggestion
// Does nothing and returns nil if opts.omitSnippets is true or if internal generation mode.
```

internal/gengapic/method_selective.go
```diff
@@ -0,0 +1,123 @@
...
+		}
+	}
+
+	sgg := g.getSelectiveGapicGeneration(protoPkg)
+	if sgg == nil || !sgg.GetGenerateOmittedAsInternal() {
+		return false
+	}
+
+	mfqn := g.fqn(m)
+	for _, inc := range sgg.GetMethods() {
```
hongalex: This (and the for loop above in `shouldGenerateMethod`) feels like a linear scan of all the selective gapic methods. I think most of our API surfaces are small enough that it doesn't matter, but if it ever gets to a point where we want to optimize, it may be better to store the allowed methods in a map or something.

bhshkh: Right. Updated the code to use map for lookups.

hongalex (approved):

PR #1792, opened 2026-07-29
fix: refactor off of protoiface dep

noahdietz:
`protoiface.MessageV1` is replaced by `proto.Message`. We don't need to depend on this pseudo-internal interface for our purposes.

Note: Using `fix` commit to elicit a release PR, though we can use `chore` if preferred.


shollyman (approved):
The machine said there's a problem with my face, and it needs to go away.  I bow to the will of the machine.

PR #1795, opened 2026-08-20
feat: consolidate iterator gencode

shollyman:
This PR consolidates the newer All() style iteration methods into auxiliary.go rather than having them in a go:build protected file (iterator_go123.go).

We are now well-past the point where new style iterators are supported.

related: https://github.com/googleapis/google-cloud-go/issues/20229

### DIFF
--- a/internal/gengapic/paging.go
+++ b/internal/gengapic/paging.go
@@ -412,6 +412,18 @@ func (g *generator) pagingCall(servName string, m *descriptorpb.MethodDescriptor
 
 func (g *generator) pagingIter(pt *iterType) {
 	p := g.printf
+
+	// include the new style iterator first.
+	p("// All returns an iterator. If an error is returned by the iterator, the")
+	p("// iterator will stop after that iteration.")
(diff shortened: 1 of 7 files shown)


internal/gengapic/paging.go
```diff
@@ -412,6 +412,18 @@ func (g *generator) pagingCall(servName string, m *descriptorpb.MethodDescriptor
 
 func (g *generator) pagingIter(pt *iterType) {
 	p := g.printf
+
+	// include the new style iterator first.
+	p("// All returns an iterator. If an error is returned by the iterator, the")
+	p("// iterator will stop after that iteration.")
+	p("func (it *%s) All() iter.Seq2[%s, error] {", pt.iterTypeName, pt.elemTypeName)
+	p("  return gaxIterator.RangeAdapter(it.Next)")
```
noahdietz: I think import renaming should align with package naming which is typically all lowercase

maybe giter ? or gxiter?

shollyman: went with gaxiter, as giter tends to make me think version-control things.

noahdietz (approved):

PR #1812, opened 2026-09-22
impl(o11y): generate `ClientTracing` options

haphungw:
Generate `gax.NewClientTracing` and bind `gax.WithClientTracing` to `CallOptions` in generated gRPC and REST client constructors.

Towards googleapis/google-cloud-go#20535


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/googleapis/gapic-generator-go
 
-go 1.26.0
```
shollyman, line 3: What's causing this?

haphungw: My local Go toolchain (1.26.6) bumped `go.mod` while I was updating `showcase` dependency to get CI pass.
1.26.0 is what's needed, so I reverted `go.mod` back to 1.26.0.

shollyman (approved):

